Issue #166, opened 2016-07-10
Disallow duplicate designators in ttp:{content,processor}Profiles


skynavga:
If multiple designators are specified in a designator list, then their absolutized forms must be distinct; i.e., no duplicates allowed.

nigelmegitt:
I'm not clear why you want to disallow this - in either the case `"all(X Y Z)"` or `"any(X Y Z)"` there would be no semantic difference by including a duplicate, e.g. `"all(X Y Z)"` would clearly mean exactly the same as `"all(X Y Y Z)"`. Perhaps this should be a _should_ rather than a _must_?

skynavga:
I consider it an error of redundancy to have multiple identical descriptors. Further, it may not be apparent from reading the content (at first glance) that they are identical. Only after their absolutized forms are determined can the check be performed. Closing this as it has been merged. A follow-on issue can be created if desired.

PR #167, opened 2016-07-10
Disallow duplicate absolutized designators in ttp:{content,processor}…

skynavga:
…Profiles.


skynavga:
See #166.

nigelmegitt:
Looks good to me - it might be improved further by including a reference to the place where the mechanism for absolutizing profile designators is defined, i.e. §5.2.3 Profile Designators - or by making that a separate subsection in its own right, or even a defined term "absolute profile designator".

Issue #175, opened 2016-09-08
Add background related styles


skynavga:
Need to add the following styles based on CSS Backgrounds and Borders Level 3:

* tts:backgroundClip
* tts:backgroundOrigin
* tts:backgroundSize

These styles are desirable to adequately support background images on regions and content elements.

skynavga:
Use tts:backgroundExtent rather than tts:backgroundSize for improved consistency.

PR #177, opened 2016-09-10
Add tts:background{Clip,Extent,Origin}.

skynavga:
Add tts:background{Clip,Extent,Origin}.
Also add [background {painting,positioning} area] defines (as TBDs).

See issue #175.


spec/ttml2.xml
```diff
@@ -7548,6 +7646,130 @@ span.</p>
...
+<loc href="#content-vocabulary-span"><el>span</el></loc>
+</td>
+</tr>
+<tr>
+<td><emph>Inherited:</emph></td>
+<td>no</td>
+</tr>
+<tr>
+<td><emph>Percentages:</emph></td>
+<td>relative to width and height of the applicable <loc href="#terms-background-positioning-rectangle">background positioning rectangle</loc></td>
```
nigelmegitt: This seems a bit confusing - isn't `backgroundExtent` _specifying_ the background positioning rectangle?

skynavga: No. It is specifying the extent (size) of the image with respect to the background positioning rectangle. The default 'auto' uses the intrinsic size, but the author may override that, e.g., by specifying cover or contain or two measures. See CSS3 B&B L3 background-size property for more info.

nigelmegitt: Oh, then the name `backgroundExtent` seems to confusing. Why not call it `imageExtent`?

skynavga: Let me give this some thought, since we might need the same attribute be used with tt:image.

nigelmegitt:
Notes from f2f meeting 2016-09-19:
1. Need to consider the style attribute names - they all are prefixed "background" but some refer to the background content and some to the background area. (there is the background painting area and the background positioning area, both defined in the terminology section)
2. Consider whether backgroundExtent can apply to foreground images.
3. Add some example images.

Proposal is to merge this and raise issues for each of the above.

Issue #176, opened 2016-09-09
Add support for rounded borders


skynavga:
The use of rounded borders (or backgrounds) on caption/subtitle regions is becoming prominent. With the addition of ``tts:border`` in TTML2, we should also add support for rounded borders basing this support on features defined by CSS Backgrounds and Border Level 3 [1]. Note that this feature may also be used to indicate the use of rounding with background coloration applied to regions or content elements, even when no border is drawn (since the default clip region for the _background painting area_ is the border region).

[1] https://drafts.csswg.org/css-backgrounds-3/#the-border-radius

The proposal is to add an optional ``<border-radii>`` component to the value syntax for ``tts:border`` as follows:

<pre>
tts:border
  : [ &lt;border-thickness&gt; || &lt;border-style&gt; || &lt;border-color&gt; || &lt;border-radii&gt; ]

&lt;border-radii&gt;
  : 'radii(' &lt;length&gt; ( ',' &lt;length&gt; )? ')'
</pre>

where, if two lengths are specified, then the first length corresponds to the quarter ellipse radius for the inline progression dimension (for content) or horizontal direction (for regions), while the second length, if present, corresponds to the quarter ellipse radius for the block progression dimension (for content) or vertical direction (for regions), or, if only one length is present, then it is interpreted as if two lengths were specified with the same value.

If a percentage value is used for a ``<length>``, then it is resolved with respect to the appropriate dimension of the border box's extent.

nigelmegitt:
None of the examples of rounded borders I have seen use different radii for the two directions, so I'd advocate simplifying to a single value for now, until a need for two radii is presented.

However one thing that may be underspecified at the moment is the flow of rounded rectangles between adjacent areas. I've seen this kind of thing:
![image](https://cloud.githubusercontent.com/assets/4988915/18413555/0d647110-77a3-11e6-8f7d-49f1f920b0b0.png)


where I would rather have seen something like this:
![image](https://cloud.githubusercontent.com/assets/4988915/18413553/05a8c354-77a3-11e6-9815-af3746928234.png)

or even perhaps:

![image](https://cloud.githubusercontent.com/assets/4988915/18413552/f194230e-77a2-11e6-9a97-1c660f0d0a3b.png)
 
In other words when creating a composite background shape from congruent area backgrounds the radial centre should be offset in the direction of the hypotenuse of the right angle triangle formed between the vertex being drawn and the two adjacent vertices, not away from it. There are probably more elegant ways to describe this!

[...]

PR #179, opened 2016-09-10
Add support for rounded borders by introducing <border-radii> compone…

skynavga:
Add support for rounded borders by introducing <border-radii> component to tts:border value expression.

See issue #176.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -7846,7 +7846,8 @@ as a style property only to those element types indicated in the following table
 [
 <loc href="#style-value-border-thickness">&lt;border-thickness&gt;</loc> ||
 <loc href="#style-value-border-style">&lt;border-style&gt;</loc> ||
-<loc href="#style-value-border-color">&lt;border-color&gt;</loc>
+<loc href="#style-value-border-color">&lt;border-color&gt;</loc> ||
+<loc href="#style-value-border-radii">&lt;border-radii&gt;</loc>
 ]
 </td>
 </tr>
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -12585,6 +12593,32 @@ or more borders.</p>
...
+</tbody>
+</table>
+<p>If two <loc href="#style-value-length">&lt;length&gt;</loc> expressions are specified, then the first length
+corresponds to the quarter ellipse radius for the inline progression dimension (for content) or
+horizontal direction (for regions), while the second length, if present, corresponds to the
+quarter ellipse radius for the block progression dimension (for content) or vertical direction
+(for regions), or, if only one length is present, then it is interpreted as if two lengths were
+specified with the same value.</p>
+<p>If a <loc href="#terms-presentation-processor">presentation processor</loc> supports only a single radii length and
+two lengths are specified, then the larger of the two lengths should be used as a single radii length.</p>
```
nigelmegitt: This 'larger radius applies' rule is unlikely to be implemented correctly and does not match what I would expect as the converse of the 'if one radius is specified then it is used for both dimensions' rule. See also https://www.w3.org/TR/css3-background/#the-border-radius which does not admit this scenario.

It seems more intuitive to say that a processor that supports only a single radii length should behave identically to a processor that does support two radii lengths but is provided only a single value. In that case the rule would be that the first length should be used as a single radius length. This is also simpler to implement since it does not require an additional comparison, which itself actually requires support for two radii length expression parsing even if not two radii length rendering.

nigelmegitt:
F2F meeting conclusion re single vs double value support: the easiest option right now is to require double value support.

Additionally, we can merge this and raise a new PR for handling merged background areas as per the issue comments.

Issue #174, opened 2016-08-16
Augment <position> syntax


skynavga:
In [1], the following two-component expression is missing:

<pre>
  | offset-position-v offset-position-h
</pre>

This expression is needed to maintain compatibility with CSS3 syntax.

[1] https://www.w3.org/TR/ttml2/#style-value-position

skynavga:
Committed.

PR #180, opened 2016-09-10
Add missing two component expression to <position> value syntax.

skynavga:
See issue #174.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -13610,12 +13610,17 @@ equivalence tables:</p>
 <td><code>left <loc href="#style-value-length">&lt;length&gt;</loc> top 100%</code></td>
 </tr>
 <tr>
-<td><code><loc href="#style-value-length">&lt;length&gt;</loc> <loc href="#style-value-length">&lt;length&gt;</loc></code></td>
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -13444,7 +13445,7 @@ edge-keyword-v
 <gitem>
 <label>two components</label>
 <def>
-<p>a horizontal offset (<code>offset-position-h</code>) followed by a vertical offset (<code>offset-position-v</code>)</p>
+<p>a horizontal offset (<code>offset-position-h</code>) and a vertical offset (<code>offset-position-v</code>), in any order</p>
```
nigelmegitt: The keywords `offset-position-h` and `offset-position-v` allow for a `position-keyword-h` or `position-keyword-v` OR a `<length>`. If two lengths are used then the order is ambiguous. Suggest qualifying this sentence with "where at least one offset value is a keyword, or if that keyword is `center` then both of them are keywords" or something equivalent.

nigelmegitt:
Review from f2f 2016-09-19: final value in 2 component mapping table does not distinguish the two `<length>` components - they need to be labelled e.g. `<length-h>` and `<length-v>`; also consider adding helpful text to ensure readers area aware of the mapping table implications when two values are ambiguously h or v.

PR #197, opened 2016-10-11
Issue 0173 safe crop area editor

skynavga:
Add ttp:safeCropArea (#173).


spec/ttml2.xml
```diff
@@ -4890,6 +4892,69 @@ If specified, then both numerator and demoninator must be non-zero.</p>
...
+container region</loc> within which the rendered content should not to be cropped during presentation processing.</p>
+<p>If specified, the value of this attribute must adhere to the following syntax:</p>
+<table id="parameter-attribute-safeCropArea-syntax" role="syntax">
+<caption>Syntax Representation &ndash; ttp:safeCropArea</caption>
+<tbody>
+<tr>
+<td>
+<eg xml:space="preserve">
+ttp:safeCropArea
+  : origin? extent
```
nigelmegitt: Making origin optional is unnecessarily complex - I would simplify this by requiring it. (incidentally in an early draft of the submission I did exactly the same thing but got strong feedback that it would be better not to do this.)

spec/ttml2.xml
```diff
@@ -4890,6 +4892,69 @@ If specified, then both numerator and demoninator must be non-zero.</p>
 <p>A <att>ttp:pixelAspectRatio</att> attribute is considered to be significant only
 when specified on the <el>tt</el> element.</p>
 </div3>
+
+<div3 id="parameter-attribute-safeCropArea">
+<head>ttp:safeCropArea</head>
+<p>The <att>ttp:safeCropArea</att> attribute may be used to designate a subset of the <loc href="#terms-root-container-region">root
+container region</loc> within which the rendered content should not to be cropped during presentation processing.</p>
```
nigelmegitt: I cannot trace any words in this pull request through to show how a presentation processor can be specified as being required to show the content within the safe crop area - it's SHOULDs all the way. I was expecting at least some mandatory words, even if only in the feature designator definition as applied to presentation processors.

spec/ttml2.xml
```diff
@@ -4890,6 +4892,69 @@ If specified, then both numerator and demoninator must be non-zero.</p>
...
+</eg>
+</td>
+</tr>
+</tbody>
+</table>
+<p>The interim computed values of the <emph>width</emph> and <emph>height</emph> of the extent of the safe crop area are determined as follows:</p>
+<ulist>
+  <item><p>the interim computed value of <emph>width</emph>, <emph>I<sub>W</sub></emph>,
+  is the minimum of the resolved value of the specified <emph>width</emph> and
+  the computed <emph>width</emph> of the <loc href="#terms-root-container-region">root container region</loc>;</p></item>
```
nigelmegitt: This seems like a very complicated way to say that the width must not be more than 100%.

spec/ttml2.xml
```diff
@@ -4890,6 +4892,69 @@ If specified, then both numerator and demoninator must be non-zero.</p>
...
+  <item><p>the interim computed value of <emph>height</emph>, <emph>I<sub>H</sub></emph>,
+  is the minimum of the resolved value of the specified <emph>height</emph> and
+  the computed <emph>height</emph> of the <loc href="#terms-root-container-region">root container region</loc>.</p></item>
```
nigelmegitt: Likewise, why not simply restrict the height to a maximum of 100%?

spec/ttml2.xml
```diff
@@ -4890,6 +4892,69 @@ If specified, then both numerator and demoninator must be non-zero.</p>
...
+</ulist>
+<p>The computed values of the <emph>x</emph> and <emph>y</emph> of the origin of the safe crop area are determined as follows:</p>
+<ulist>
+  <item><p>if no <code>origin</code> component is specified, then the computed value of <emph>x</emph> is
```
nigelmegitt: We can just remove this by requiring the origin to be specified.

spec/ttml2.xml
```diff
@@ -4890,6 +4892,69 @@ If specified, then both numerator and demoninator must be non-zero.</p>
...
+  <item><p>if no <code>origin</code> component is specified, then the computed value of <emph>y</emph> is
+  one half of the difference between the computed <emph>height</emph> of the <loc href="#terms-root-container-region">root container region</loc> and <emph>I<sub>H</sub></emph>;
+  otherwise, it is the minimum of the resolved value of the specified <emph>y</emph> and the computed <emph>height</emph> of the
+  <loc href="#terms-root-container-region">root container region</loc>.</p></item>
+</ulist>
+<p>The final computed values of the <emph>width</emph> and <emph>height</emph> of the extent of the safe crop area are determined as follows:</p>
+<ulist>
+  <item><p>the final computed value of <emph>width</emph> is the minimum of (1) the difference between the computed <emph>width</emph>
+  of the <loc href="#terms-root-container-region">root container region</loc> and the computed value of <emph>x</emph> and (2) the interim computed value of <emph>width</emph>,
+  <emph>I<sub>W</sub></emph>;</p></item>
```
nigelmegitt: It's unclear what this is trying to achieve, or why it is a good idea. Same goes for height.

For example, say the width is 60% and x is 30%; then the difference between those is 30%, which is smaller than the interim computed value of width, being 60%. So safe crop area is specified as having extent 60% and this clause makes it 30% instead. That seems like an unexpected result.

spec/ttml2.xml
```diff
@@ -4950,8 +4959,8 @@ x, y, width, height
   <emph>I<sub>H</sub></emph>.</p></item>
 </ulist>
 <p>Percentage values specified by this attribute are resolved with respect to the <loc href="#terms-root-container-region">root container region</loc>.</p>
-<p>When presented on a visual medium, the rendered content of the safe crop area should remain visible, i.e., not be cropped, unless overridden
-by a higher level protocol.</p>
+<p>When presented on a visual medium, the rendered content of the safe crop area must not be cropped unless overridden
+by a <loc href="#terms-higher-level-protocol">higher level protocol</loc>.</p>
```
nigelmegitt: Thanks, this is better. I'm not sure exactly why we need to distinguish the processing context from the higher level protocol - from the perspective of the TTML 2 semantics they seem like they are both "instructions from above" and the precise origin doesn't make any difference.

Issue #181, opened 2016-09-15
Viewport unit {vw,vh} naming: {rw,rh} preferred.


nigelmegitt:
TTML2 introduces the excellent new units `vw`and `vh`, which are relative to the width and height of the root container region. However this definition could be perceived as clashing with the CSS definition of the identically named units, which are relative to the viewport. This may cause (at least) two implementation difficulties:

1. In the case of a root container region that is not coincident with a related media object (or the media player) an implementation may assume that the root container region is in fact coincident and calculate sizes incorrectly.

2. A naïve translation to HTML and CSS for rendering on a media player may not translate the units at all and assume that for TTML rendering purposes the viewport is identical to a host HTML page's  viewport, resulting in incorrect sizes except when viewing in full screen mode.

This is a case of something that is not exactly _wrong_ as is, but may encourage or lead to poor implementations. I would propose nudging implementors towards checking the specification more carefully by renaming these units `rw` and `rh` to make it clear that they are related to the root container region and not the viewport.

I would also propose the addition of an editorial note describing how to convert `rw` to `vw` and and `rh` to `vh` respectively given a root container region whose size is defined relative to a media player window and whose media player window size is known relative to the viewport.

skynavga:
Given there are already multiple implementations using the new units, and given there is already a substantial amount of content using them, changing names would be unnecessarily disruptive and serve no technical purpose. My response to confused users is RTFM.

nigelmegitt:
On further thought the current use of `vw` and `vh` is in fact a bug since, on RTFMing CSS they mean something different there, and we now are referencing CSS. Reopening.

nigelmegitt:
By the way, commenting as Chair, please note that prematurely closing issues without offering a chance for discussion and consideration is an abuse of process; also, though early implementations are highly valued, I'm sure that the implementers are very well aware that they are using a working draft only that is subject to change. The arguments used to dismiss this issue initially are applicable to CR stage but not WD stage. We do have freedom to make changes where deemed appropriate at WD stage.

skynavga:
If there were a technical reason for considering a change, I would do so. But there is none, and the usage is already active. I am removing the bug label. In no way is this a bug.

skynavga:
Also, regarding CSS, we are entirely consistent with the CSS definition since the TTML root container region corresponds with notion of the CSS initial containing block.

nigelmegitt:
[...]

PR #199, opened 2016-10-11
Change {vw,vh} unit names to {rw,rh} (#181).

skynavga:
See #181.


skynavga:
Regenerated ED on gh-pages.

nigelmegitt:
Thanks, that's fixed it.

PR #218, opened 2016-11-03
Add tts:textShadow example and modify semantics

nigelmegitt:
This pull request makes some reversions back to CSS for `tts:textShadow` and adds an example image. Specifically the `inset` keyword and the 4th, "spread" value are prohibited in CSS so I removed them. This also dealt with an editorial note to define the semantics of `inset`. If there is a strong need to bring back and define those two extensions relative to CSS then I can revert those parts of the pull request.


nigelmegitt:
[Meeting 2016-11-03] @skynavga to review this pull request. Working plan is to remove `inset` and spread value from TTML2 spec to align with CSS and also for @nigelmegitt to ask @bert-github and @fantasai what is the reasoning for omitting them from [`text-shadow`](https://www.w3.org/TR/css-text-decor-3/#text-shadow-property) in CSS when they are present in [`box-shadow`](https://www.w3.org/TR/css3-background/#box-shadow).

nigelmegitt:
Some helpful pointers on `inset` and spread from CSSWG people:
> http://lists.w3.org/Archives/Public/www-style/2010Nov/0412.html
> http://lists.w3.org/Archives/Public/www-style/2009Sep/0108.html
> http://lists.w3.org/Archives/Public/www-style/2013Jul/0589.html

This selection seems to indicate that at least some CSS people would like `inset` added and possibly spread too, though there's less clarity on the semantics for spread on complex glyph outlines.

skynavga:
Based on the input cited above from others, I recommend we keep <shadow> defined as is. Further, it might be we should change the name from tts:textShadow to tts:shadow, and allow it to apply to region as well (to serve in the same capacity as box-shadow).

nigelmegitt:
I'd rather take the simpler approach of leaving it as it is in this pull request. That way we can easily create examples, remain compatible with CSS, and if at a later point CSS adopts a richer semantic we can consider adding that in a non-breaking addition. CSS hasn't exactly rushed to enrich text-shadow, albeit that I can see the modifications might be great for some applications.

We have no requirements identified for box shadow or for inset text shadow, so going to effort at this stage to define something is not a worthwhile activity in my opinion, in the context of our current priorities. 

If someone wants to specify and implement something I won't block it but neither will I support extra effort when our focus should be on getting a new WD out. Also, I think it will be difficult to specify spread semantics - one commenter pointed out that the distance-and-radius model is too simplistic for complex vector outlines, and I think I can see the issue. It would make more sense in the case of text to specify a modified glyph size to use as the basis for the shadow and also define a zero-point centering algorithm on a per glyph basis.

It will not be straightforward to create examples of inset shadows using existing implementations.

fantasai:
I believe the CSSWG has no objection to extending text-shadow syntax to handle spread and inset; there is only a concern about implementation.

bert-github:
> On 03 Nov 2016, at 16:35, Nigel Megitt <notifications@github.com> wrote:
> 
> [Meeting 2016-11-03] @skynavga to review this pull request. Working plan is to remove inset and spread value from TTML2 spec to align with CSS and also for @nigelmegitt to ask @bert-github and @fantasai what is the reasoning for omitting them from text-shadow in CSS when they are present in box-shadow.

W.r.t. to the reasons ’text-shadow’ and ‘box-shadow’ in CSS are different, I think there are two:

1) ’text-shadow’ predates ‘box-shadow’; and the spread and inset features are relatively late additions to ‘box-shadow’.

2) The research into how hard it would be to add inset and spread to ’text-shadow’ has not been done. Some font APIs on some platforms might make it hard or slow to implement.

In other words, it’s not impossible that spread and inset get added to ’text-shadow’ in the future.



Bert
--
  Bert Bos                                ( W 3 C ) http://www.w3.org/
  http://www.w3.org/people/bos                               W3C/ERCIM
  bert@w3.org                             2004 Rt des Lucioles / BP 93
  +33 (0)4 92 38 76 92            06902 Sophia Antipolis Cedex, France

nigelmegitt:
Thank you for that input @fantasai and @bert-github - from a pragmatic perspective I take that as supporting my view that we follow CSS for now and consider extending later. However I also understand @skynavga 's point that we could specify now: my concern is purely practical given our timing need to move towards CR.

nigelmegitt:
@skynavga if you change `tts:textShadow` to `tts:shadow` how do you distinguish between shadows applied to boxes and to shadows applied to text within boxes? I can imagine an example like:

```xml
<p tts:boxShadow="..." tts:textShadow="...">Some text with shadow in a box also with a different shadow</p>
```

skynavga:
> how do you distinguish between shadows applied to boxes and to shadows applied to text within boxes?

yes, I thought of that as well (after making the earlier suggestion to merge them); so, I'm fine with removing inset and spread on tts:textShadow for now, and, in the future, considering an enhanced definition of <shadow> syntax to support a tts:shadow property that would apply to areas (aka boxes); at such a time, we could also consider adding inset/spread to textShadow if desired;

nigelmegitt:
@skynavga ok we're agreed. All the changes needed to apply that decision are already present in this pull request, if I have not made a mistake.

nigelmegitt:
Self-reviewing this, I would like to change the example slightly to avoid using `xml:space="preserve"` and spacing at the beginning of the second line. Will commit an update to this branch shortly.

nigelmegitt:
Fix now in place - this example is preferable now to my eyes than the previous one I submitted, and less suggestive of better-avoided-if-possible practices like using spaces for horizontal positioning.

Issue #232, opened 2017-01-09
Apply missing style properties to tt:image.


skynavga:
In CSS, border and padding can be applied to images in both block and inline contexts.  For symmetry, and since we now allow both to apply to content elements, we should also allow them to apply to foreground images expressed by tt:image.

skynavga:
Also need to apply backgroundColor, backgroundImage, etc to foreground images.

skynavga:
See http://codepen.io/skynavga/pen/EZZqxW for example using CSS rendering.

skynavga:
Also need to apply bpd, display, ipd, visibility.

skynavga:
Resolved in 49cc2cd3687e9e1d39277ce65eef4730b705fc83.

PR #234, opened 2017-01-19
Apply additional style properties to tt:image, allow @style on tt:image (#232).

skynavga:
See #232.


skynavga:
Good points. I'll add.

PR #242, opened 2017-02-10
Issue 0216 fontsize em on region

skynavga:


spec/ttml2.xml
```diff
@@ -9613,8 +9612,8 @@ as a style property only to those element types indicated in the following table
 </tr>
 <tr>
 <td><emph>Percentages:</emph></td>
-<td>if not <loc href="#content-vocabulary-span"><el>region</el></loc> element,
-then relative to parent element's font size; otherwise, relative to the
+<td>if not <loc href="#document-structure-vocabulary-tt"><el>tt</el></loc> element,
+then relative to nearest styled ancestor element's computed font size; otherwise, relative to the
```
palemieux: If fontSize="100%" is specified on region alone, then

* in TTML1, fontSize on the region is computed to be "1c"
* in TTML2, fontSize on the region would be computed to 100% of _nearest styled ancestor element's computed font size_

What is a _styled element_?

skynavga:
On Thu, Feb 9, 2017 at 11:30 PM, Pierre-Anthony Lemieux <

> *@palemieux* requested changes on this pull request.
>
> In spec/ttml2.xml
> <https://github.com/w3c/ttml2/pull/242#pullrequestreview-21184885>:
>
> > @@ -9613,8 +9612,8 @@ as a style property only to those element types indicated in the following table
>  </tr>
>  <tr>
>  <td><emph>Percentages:</emph></td>
> -<td>if not <loc href="#content-vocabulary-span"><el>region</el></loc> element,
> -then relative to parent element's font size; otherwise, relative to the
> +<td>if not <loc href="#document-structure-vocabulary-tt"><el>tt</el></loc> element,
> +then relative to nearest styled ancestor element's computed font size; otherwise, relative to the
>
> If fontSize="100%" is specified on region alone, then
>
>    - in TTML1, fontSize on the region is computed to be "1c"
>    - in TTML2, fontSize on the region would be computed to 100% of *nearest
>    styled ancestor element's computed font size*
>
> What is a *styled element*?
>
I needed a term that would apply to tt but not to layout or head, since
styles cannot be specified on them. In other words, region inherits from tt
in TTML2, and the initial value of fontSize is 1c on tt (absent being
overridden by initial).

I will add a glossary entry for *styled element* to this end.

palemieux:
> I needed a term that would apply to tt but not to layout or head, since styles cannot be specified on them.

Since styles are computed after ISD construction, where regions can be made children of tt, thereby removing the need to create new terminology.

skynavga:
On Fri, Feb 10, 2017 at 2:26 PM, Pierre-Anthony Lemieux <

> I needed a term that would apply to tt but not to layout or head, since
> styles cannot be specified on them.
>
> Since styles are computed after ISD construction, where regions can be
> made children of tt, thereby removing the need to create new terminology.
>
Incorrect.

skynavga:
To elaborate on [1]:

>>Since styles are computed after ISD construction, where regions can be made children of tt, thereby removing the need to create new terminology.
>
>Incorrect.

ISD construction flattens styles as far as possible, including resolution of computed styles values. Only in some special circumstances, namely, those where the resolution of a computed style requires that layout occur or require access to presentation time state, is it the case that styles are computed after ISD construction. 

[1] https://github.com/w3c/ttml2/pull/242#issuecomment-279102556

Issue #203, opened 2016-10-18
Add examples of new features commonly used with Japanese content (2 of 2)


dae-kim:
To address the following features:
Editorial note: Ruby Offset Example
Editorial note: Ruby Overflow Example
Editorial note: Ruby Overhang Example
Editorial note: Ruby Overhang Class Example
Editorial note: Text Combine Example

dae-kim:
See: https://github.com/w3c/ttml2/pull/273

nigelmegitt:
I've just checked this and realise that there are unresolved comments following the merge of #273, which can be found at https://github.com/w3c/ttml2/pull/273#issuecomment-299896141 so there's a little more work to do on this please @dae-kim .

skynavga:
Fixed (outstanding comments) and merged from #312.

PR #273, opened 2017-03-03
Issue 0203 add examples of ja features part 2

dae-kim:
See: https://github.com/w3c/ttml2/issues/203


nigelmegitt:
@r12a You can view this version of the doc (for this pull request) at https://rawgit.com/w3c/ttml2/b4ad18b81c329bb317a2158e15699820cb5e95b3/spec/ttml2.html#style-attribute-ruby (fragment id gets you near to the main changes)

r12a:
I began looking. I continue when back in the office next week.

One observation for now. In Example Fragment – Ruby Overflow some of the angle brackets are the wrong way round.

nigelmegitt:
Meeting 2017-04-27: @skynavga to review for correctness, @r12a to be "pinged" since it looks as though he was part way through reviewing this.

r12a:
I took another look. I think it's mostly ok, and thanks for adding that. A couple of comments:

Looking at http://w3c.github.io/ttml2/spec/ttml2.html#ruby-position-example-1
- it would be much easier to read if meaningful style names were used, rather than s4, which you have to look up to read the code.  How about just naming them 'ruby', 'rt', etc?
- i think i would have preferred to see a single string with annotations on both sides, rather than two separate strings, with one annotation each, which looks quite wierd to me.

If you take a decision on that, then i think we could probably close this issue. If anything else crops up, i can raise it as a separate issue, but the main request for addition of examples appears to be done.

PR #283, opened 2017-03-23
Add tts:hdrAbsoluteLuminanceGain

### LINKED ISSUE #233: Add support for compositing a TTML2 document onto HDR video
Colors in TTML2 documents are specified using sRGB. This yields ambiguities when compositing such documents onto HDR video, whose allows peak luminance much higher than that associated with sRGB, e.g. ITU Rec 2100 allows luminance values up to 10000 cd/m^2 while sRGB specifies a reference white of 80 cd/m^2.

As a practical example, white sRGB caption text should rarely, if ever, be composited onto HDR video assuming that sRGB (255,255,255) maps to peak HDR white: the resulting text would be unacceptably brighter than the rest of scene, whose average luminance is typically far below peak HDR white. At what luminance level should sRGB captions be composited onto HDR images -- without having to define new colorspaces and encodings for TTML2?

To allow authorial control of the compositing of TTML2 documents onto HDR images, the [attached](https://github.com/w3c/ttml2/files/778373/ttml-hdr-20170208.pdf) proposes an `tts:hdrAbsoluteLuminanceGain` attribute, which allows the normalized rgb components of each pixel of a region to be mapped to absolute optical output components (in cd/m^2) 

(R, G, B)= 80 cd/m^2 ∙ hdrAbsoluteLuminanceGain ∙ (r, g, b)

It would also be useful to note that, in the case of image-based subtitles, `tts:tts:hdrAbsoluteLuminanceGain` is not necessary if the image is an HDR image.

palemieux:
Closes #233


nigelmegitt:
@skynavga Why the closure of this pull request, without any commentary, please? What is your plan here for addressing #233?

skynavga:
Working this issue now, so stand by.

mrd489:
@palemieux In the example for "Absolute linear optical output calculation using tts:hdrAbsoluteLuminanceGain," the resulting triplet does only conditionally has units of cd/m<sup>2</sup>. The calculation of luminance yields only one number, and is still a weighted average of the resulting triplet. A triplet with the same 3 linear output values, for example, would refer to cd/m<sup>2</sup>. Since these numbers are different, the luminance units does not apply this way.

palemieux:
@mrd489 To confirm, you recommend that the units of cd m^-2 be removed from the constants 80 cd m^-2 and 10000 cd m^-2 when used in the equations?

mrd489:
@palemieux Yes, since the RGB triplet is not always a neutral. You could have an informative note that mentions this.

skynavga:
@nigelmegitt regarding your inquiry

>@skynavga Why the closure of this pull request, without any commentary, please? What is your plan here for addressing #233?

I closed this PR in order to create a new branch with the desired naming convention. Unfortunately, there is no way to rename a branch in a remote repository that is already referenced by existing PRs, so it was easier to just create a new branch and new PR (#285).

I responded to an inquiry from @palemieux here [1] regarding what changed.

[1] https://github.com/w3c/ttml2/issues/233#issuecomment-290346018

Issue #96, opened 2015-11-04
Specify date as well as hours, minutes and seconds in time expressions


plehegar:
Proposal from EBU to permit date to be specified in time expressions, optionally. Purpose of this is to allow 24 hour clock times to be used without ambiguity at midnight crossing. Also to permit second-accurate duration calculations by allowing externally managed reference of dates that include leap seconds.

An implementation may usefully allow dates to be inferred or inherited from a reference element rather than requiring the same date on every begin and end attribute where date semantics are needed.

(raised by Nigel Megitt on 2013-10-31)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/293

skynavga:
Lacking a hard requirement and draft specification text, I am marking this for ttml.next and closing. If there is a hard requirement here, please draft specification text and reopen.

nigelmegitt:
There is a clear and hard requirement here. Consider a midnight crossing example (or is it?):

```xml
<tt ttp:timeBase="clock" ttp:clockMode="utc">
...
<p begin="23:59:58" end="00:00:01">[content here]</p>
...
</tt>
```

Comparison of the two time expressions reveals that the end of the `p` is earlier than its begin, therefore it never becomes active. There is no conformant way to discern this is happening using clock-time expressions - only offset-time expressions have some hope of doing so, and even then the epoch (time zero) is poorly defined.

The request in this issue is to allow for a date component to be added to clock-time expressions.

nigelmegitt:
[Meeting 2016-03-16] Group just about agrees to adopt the date format specified in SMIL when `ttp:timeBase="clock"` - see https://www.w3.org/TR/smil/smil-timing.html#Timing-WallclockSyncValueSyntax - it is based in part on ISO8601 and allows a year-month-day expression to precede the time value.

nigelmegitt:
Meeting 2017-05-18: Discussion about timezone addition to time expressions; majority view against doing so since UTC can be used instead. I take action to propose wording to clarify the intended meaning of `clockMode="local"` as being undefined with respect to UTC and intended to be "whatever time this is where you are".

skynavga:
Add clarifying notes in c36dffb71744a489372fe6a86ab303dc2f258a5d, which closes the outstanding comments.

PR #288, opened 2017-04-13
Issue 0096 date time

skynavga:
Add wallclock-time form for time-expression syntax, allowing explicit dates; however, do not support explicit time zone offsets since ``ttp:clockMode`` already implicitly selects time zone that applies to time expressions.

Add #time-wall-clock feature designator.

See #96.


spec/ttml2.xml
```diff
@@ -16823,18 +16817,43 @@ explicit specification of frame rate, sub-frame rate, tick rate, and time base.<
...
   : time-count fraction? metric?
 
+wallclock-time
+  : "wallclock(" <loc href="#style-value-lwsp">&lt;lwsp&gt;</loc>? ( date-time | wall-time | date ) <loc href="#style-value-lwsp">&lt;lwsp&gt;</loc>? ")"
+
+date-time
+  : date "T" wall-time
+
+wall-time
+  : ( hhmm-time | hhmmss-time )
```
nigelmegitt: This is missing a `(TZD)?` component.

spec/ttml2.xml
```diff
@@ -16886,6 +16908,8 @@ when the <code>clock</code> time base applies.</p>
 <p>If a &lt;time-expression&gt; is expressed in terms of an
 <emph>offset-time</emph> and no <emph>metric</emph> is specified, then it is to be treated as
 if a metric of <code>s</code> (seconds) were specified.</p>
+<p>It is considered an error if the <emph>wallclock-time</emph> form of a &lt;time-expression&gt;
+is used in a <loc href="#terms-document-instance">document instance</loc> and the government time base is not <code>clock</code>.</p>
```
nigelmegitt: s/government/governing

?

spec/ttml2.xml
```diff
@@ -22464,29 +22512,30 @@ as follows:
 </sitem>
 <sitem/>
 <sitem>
-(1) if <code>local</code>, then the difference between the local real time at the most immediately prior local midnight and the local real time
+(1) if <code>local</code>, then the difference between the local real time at the most immediate prior local midnight and the local real time
```
nigelmegitt: NB it is not possible to infer anything from this which definition of 'local' applies - this does not need changing in my view but does point to why the `(TZD)?` component needs to be available in wallclock time expressions.

nigelmegitt: The definitions of epoch when `local` applies and `C` seem to conflict. It seems to me that the date part of a wallclock time defines the epoch and the applicable midnight to use and the `C` is just an interpretation of the time part of the expression relative to that epoch.

skynavga:
There is no need to unmerge to affect changes. Further, there is already a
discussed and agreed status. Please create a new issue listing any new
changes requested, and I will process ASAP.

aphillips:
I have some heartburn here. What happens with `ttp:clockMode` when its value is `local`? A local time offset (LTO) is not a good definition of a time zone and the syntax here doesn't permit the expression of the LTO in any case. Are local time zone rules expected to apply to the date and time values? If not, how do "local" times work?

For reference, see https://www.w3.org/TR/timezone/

skynavga:
I originally included a timezone component in the wallclock-time expression
in order to align with SMIL 3.1. However, I soon realized that this would
be at odds with the fact that a fixed ttp:clockMode already implies a
specific time zone. In the case of 'local' clock mode, it effectively means
whatever time zone applies to the document processing context (not the
authoring context). Adding a timezone component in this case would only be
confusing and cumbersome. The same applies with 'utc' clock mode.

The only question that remains in my mind is whether we should allow the
optional use of a 'Z' timezone component when clock mode is 'utc'.

nigelmegitt:
> In the case of 'local' clock mode, it effectively means
whatever time zone applies to the document processing context (not the
authoring context)

As stated on the pull request review comments I would like to reintroduce the optional `(TZD)?` component so that it can be defined in the authoring context, which seems like a reasonable thing to want to do. Also note that one use case for this is when the generated document is an archive of what was actually created, rather than a document intended for later presentation. In the case of live subtitles this is an important use case, and it is reasonable to record the time zone in this use case.

> Adding a timezone component in this case would only be
confusing and cumbersome.

I don't agree.

> The only question that remains in my mind is whether we should allow the
optional use of a 'Z' timezone component when clock mode is 'utc'.

That would be of no benefit that I can determine.

skynavga:
On Thu, Apr 13, 2017 at 10:27 AM, Nigel Megitt <notifications@github.com>
wrote:

> In the case of 'local' clock mode, it effectively means
> whatever time zone applies to the document processing context (not the
> authoring context)
>
> As stated on the pull request review comments I would like to reintroduce
> the optional (TZD)? component so that it can be defined in the authoring
> context, which seems like a reasonable thing to want to do. Also note that
> one use case for this is when the generated document is an archive of what
> was actually created, rather than a document intended for later
> presentation. In the case of live subtitles this is an important use case,
> and it is reasonable to record the time zone in this use case.
>
While I do not take any position on this possible use case, it has not been
articulated previously (AFAICT). Further, it is a significant departure
from the use of time expressions in TTML to specify intended presentation
time, and not recorded time. As such, I question whether we should be
considering such use case in the present context, namely the use of time
expressions in TTML as presently the case.

One can imagine introducing metadata items to be used in recording time
stamps, and one can imagine that timezone data may be useful there, but to
conflate this usage with an interpretation of time expressions that are
based on ttp:timeBase and ttp:clockMode is over-reaching in the context of
this issue.

nigelmegitt:
> it has not been articulated previously (AFAICT)

In fact it was articulated back in October 2014: https://www.w3.org/2014/10/27-tt-minutes.html and in general the use case was mentioned in https://www.w3.org/wiki/TTML/changeProposal025 

Certainly there has never been any text in the spec that constrains the times to be in the future, and I have always read it that the times may either be historical or future-based, especially when `ttp:timeBase="clock"`. So it is not at all a significant departure but a natural consequence of the language already present (and absent) from the specification.

Further, there is implementation work that creates TTML documents based on historical events already, using exactly this approach.

skynavga:
Thanks for the pointer. I will review. At first glance, supporting this
mode would require introducing a new clockMode value, say 'archive', which,
if used, would permit the use of timezone information.

nigelmegitt:
> At first glance, supporting this
mode would require introducing a new clockMode value, say 'archive', which,
if used, would permit the use of timezone information.

I would say no such change is needed - if anything the fact that the times are archive times is a matter of metadata rather than the times themselves. I have not found anything other than time zone that cannot be done with the spec as it is when creating archived recordings of live subtitles, so I would do as little as possible here.

skynavga:
The problem is that using a wall clock time with explicit time zone in
either clock mode local or utc doesn't make any sense, unless you ignore
the meaning of local and utc. That is, IMO, a time expression that attempts
to express a time outside of the local time system while saying it uses the
local time system is not well formed, and similarly for utc.

aphillips:
@skynavga All time values need to become incremental times measured on the system's epoch-based timekeeping system (monotonic time measured using integers) for processing purposes. Otherwise you can't really do anything. Time zones (or LTO if that's all you have) can be used as part of the process of converting the time fields into the incremental representation.

A wall clock time `2017-04-13T10:24` with a time zone of `America/Los_Angeles` has a fully specified relationship to incremental time. So does `2017-04-13T17:24Z`. Or `2017-04-13T10:24-0700`. A wall clock time that relies on `local` can work if and only if `local` is defined to include time zone rules and noting that some time values (during daylight savings/summer time changes) are ambiguous unless they include at least LTO.

Does that make sense?

PR #294, opened 2017-04-22
Revised tts:luminanceGain

palemieux:
Addressed https://github.com/w3c/ttml2/pull/283#issuecomment-290472127
Changed tts:luminance to tts:luminanceGain
Corrected tts:luminanceGain example
Removed conformance requirements on the values of tts:luminanceGain
Minor prose tweaks

PR #299, opened 2017-04-25
Issue #298 sRGB on HLG

### LINKED ISSUE #233: Add support for compositing a TTML2 document onto HDR video

### LINKED ISSUE #298: Add support for sRGB compositing for HLG
Opening a separate issue to track the editing work to support HLG, as introduced in https://github.com/w3c/ttml2/issues/233#issuecomment-296168903 .

nigelmegitt:
Resolve #298 by moving the existing Appendix P PQ section to a sub-section and adding a new sub-section for HLG.


nigelmegitt:
Please note that the branch name for this pull request was derived from the originating #233 before I split the new issue out, hence the slightly confusing naming.

nigelmegitt:
I plan to add an informative CIE XYZ reference when I can confirm the correct citation.

nigelmegitt:
Informative reference to CIE XYZ now added, I think this is good to review please.

nigelmegitt:
Notes from meeting 2017-04-27: Following merge of #294 need to merge those changes in and resolve conflicts (@nigelmegitt to do). Request was made by @skynavga to modify the CIE XYZ reference to align it better with other ISO references and change the key to just `[XYZ]` for conciseness.

nigelmegitt:
https://github.com/w3c/ttml2/pull/299#issuecomment-297954369 addressed in d7ca266 - this can now be reviewed again please @skynavga and @palemieux (and anyone else).

skynavga:
Merging after 3 days have passed from last commit without further comment.

PR #301, opened 2017-04-27
Adds xml:lang to metadata examples (issue #271)

### LINKED ISSUE #271: Metadata examples do not include xml:lang (editorial)
https://w3c.github.io/ttml2/spec/ttml2.html#metadata-vocabulary-metadata

There are several examples of `ttm:title` or `ttm:desc` containing natural language text, but no `lang` attributes are supplied.

palemieux:
Closes #271


nigelmegitt:
I'm happy with the idea of a note about indicating language of metadata items. I would also be happy to introduce a new example specifically to make this clear, with for example two `ttm:item` elements representing the same thing in different languages, each with a different `xml:lang` value. If those two changes were made then the changes made to existing examples in this pull request (in its current state at 4f499e7) would not be needed.

nigelmegitt:
I've opened #307 as an alternate pull request to this one - if we adopt #307 then I suggest we close this without merging.

skynavga:
Again, I reject this approach in dealing with the original issue. The correct way to resolve the original issue is to add a note as suggested above, and not change the example.

skynavga:
Closing as this has been superseded by #307.

PR #303, opened 2017-05-02
Add note re: bopomofo

### LINKED ISSUE #253: Support for bopomofo ruby needed
10.2.39 tts:rubyPosition
https://www.w3.org/TR/2016/WD-ttml2-20161117/#style-attribute-rubyPosition

There is no mention of support for zhuyin fuhao/bopomofo ruby, needed for Traditional Chinese.  Please add something.

For background information see http://w3c.github.io/clreq/#overview_of_positioning_of_interlinear_annotations ff

palemieux:
Closes #253


nigelmegitt:
I'm confused about the change made in a96395f because when we have discussed this before @skynavga you have always been careful to use the term zhuyin fuhao rather than bopomofo, yet you have removed that term from the note here. Why not use both terms as in the original proposal?

skynavga:
Although zhuyin fuhao is the more accurate (native Chinese) technical term, both Unicode and CSS use bopomofo to label these characters. [1] Furthermore, it is the ``inter-character`` [2] positioning that is not supported. Bopomofo can still be used, but with before or after positioning.

[1] http://www.unicode.org/charts/PDF/U3100.pdf
[2] https://www.w3.org/TR/css-ruby-1/#inter-character-layout

PR #305, opened 2017-05-04
Updated tts:writingMode example (issue #269)

### LINKED ISSUE #269: tts:writingMode example is not helpful
https://w3c.github.io/ttml2/spec/ttml2.html#style-attribute-writingMode

The illustration of the `rltb` writing mode uses `unicodeBidi` override and `direction=rtl` to show some English text backwards. This isn't particularly helpful, since it isn't really an effect of writing mode and non-observant readers may not realize that the example has these features turned on in the example. They might therefore attribute the RTL results to the writing mode. I would instead suggest showing the text strongly right aligned, possibly in conjunction with an Arabic or Hebrew translation.

palemieux:
Closes #269


palemieux:
@r12a I have addressed your comment.

r12a:
lgtm tx

PR #307, opened 2017-05-10
Issue 0271 xml lang metadata examples take 2

nigelmegitt:
2nd take at resolving #271, this time as discussed by adding an example with two `ttm:item` elements, each with a different `xml:lang`.


nigelmegitt:
@skynavga I thought from our meeting yesterday that I had this action. In any case I object to removing the example and would like to reinstate it as discussed in order better to illustrate the note you have added.

skynavga:
I disagree. It is not required in the example, adds noise to the example,
and produces an unwanted and unnecessary precedent. No illustration of the
note is required.

skynavga:
Besides, wouldn't it be better to ask the original commenter if the note
suffices?

nigelmegitt:
The commenter has already given a hint here, when he added a +1 to the proposal to add a note and an example as per https://github.com/w3c/ttml2/pull/301#issuecomment-298730418 - nevertheless I'm of course happy to receive further feedback.

I disagree that the new example adds noise; it illustrates the note and therefore makes that note easier to understand. That is the purpose of examples. This is not setting any new precedent. Unless you think that it does not correctly illustrate the note then this seems like a pointless debate; I'm afraid I haven't yet been able to understand why you are so against adding the example @skynavga. By the way just to be clear, I am talking about reinstating the specific short example that only illustrates two `ttm:item` elements each with differing `xml:lang` values and content as introduced by 4679535, not the addition of `xml:lang` to an existing example, which I think we have agreed not to do.

skynavga:
On Fri, May 12, 2017 at 1:55 AM, Nigel Megitt <notifications@github.com>
wrote:

> The commenter has already given a hint here, when he added a +1 to the
> proposal to add a note and an example as per #301 (comment)
> <https://github.com/w3c/ttml2/pull/301#issuecomment-298730418> -
> nevertheless I'm of course happy to receive further feedback.
>
> I disagree that the new example adds noise; it illustrates the note and
> therefore makes that note easier to understand. That is the purpose of
> examples. This is not setting any new precedent. Unless you think that it
> does not correctly illustrate the note then this seems like a pointless
> debate; I'm afraid I haven't yet been able to understand why you are so
> against adding the example @skynavga <https://github.com/skynavga>. By
> the way just to be clear, I am talking about reinstating the specific short
> example that only illustrates two ttm:item elements each with differing
> xml:lang values and content as introduced by 4679535
> <https://github.com/w3c/ttml2/commit/46795357edb8f1844ec03cf2625cc4b2568b1171>,
> not the addition of xml:lang to an existing example, which I think we
> have agreed not to do.
>

Let me repeat for about the 5th time. This section is not about language
identification, it is about named metadata item elements. There is nothing
special about named metadata item elements with respect to language
identification that doesn't apply to every other element type that permits
xml:lang. It is quite absurd, incongruous, and inconsistent to insist that
anything at all be said about language identification in this context,
since it is entirely orthogonal to the element definition.

If we add this silly example here, the next thing you or someone is going
to do is ask for one everywhere. And I'm just not going to go down that
path, so please stop trying to force a bad and wrong resolution of the
original comment. I would say the original comment needs no response in the
spec whatsoever, but I have added a note to assuage the original commenter,
and that should suffice.

skynavga:
I'm willing to accept an alternative which perhaps would suit you: remove
the note I added under 14.1.6 (leaving that section unchanged) and then add
a new example under 8.2.8, with the caption "Example Fragment - Language
Identification on Content and Metadata", and with code that shows xml:lang
on root, on p, and on ttm:item.

nigelmegitt:
@skynavga That would work for me also.

skynavga:
Added example in 8.2.8. Decided to leave note in place in 14.1.6.

skynavga:
No outstanding substantive comments and 3 days have passed.

Issue #219, opened 2016-11-05
Step 10.4.4.2(6)(a) does not apply to textDecoration


palemieux:
Section 10.2.44 implies that styling inheritance should apply to each of the underline, lineThrough and overline decorations individually.

_if P is present in the specified style set of E, SSS(E), then continue to the next style property_ in step 10.4.4.2(6)(a) need to be rephrased.

skynavga:
No it doesn't. There is only one textDecoration property, and it applies equally to underline, overline, and line through.

palemieux:
If the parent specifies `textDecoration="underline lineThrough"` and the child specifies `textDecoration="noUnderline"`, the result is `textDecoration="lineThrough"`, right?

skynavga:
ok, I see what you are saying; this is a special case where inheritance applies even when there is an explicit inline style; hmm, going to have to think about this a bit;

nigelmegitt:
[Meeting 2017-01-12] Agreed that @skynavga will craft some words to say that each individual property is inherited separately, resulting in the outcome described in https://github.com/w3c/ttml2/issues/219#issuecomment-258591321.

skynavga:
Resolved in #308 merge.

PR #308, opened 2017-05-13
Issue 0219 text decoration inheritance

skynavga:
See #219.


spec/ttml2.xml
```diff
@@ -12776,6 +12776,21 @@ value in any order, such as <code>"noUnderline overline lineThrough"</code>.</p>
...
+<div4 id="style-attribute-textDecoration-special-semantics">
+<head>Special Semantics</head>
+<p>The computed value of the property associated with this attribute consists of a 3-tuple
+of binary sub-values, each of which denotes, respectively, whether an underline, line through, or overline
+decoration is to be applied to the affected text.
+Furthermore, when applying inheritance semantics, each of these sub-values is considered to
+be inherited separately from the others.</p>
+<note role="example">
+<p>For example, if the computed value of <code>tts:textDecoration</code> that applies to a <code>div</code> (division) element
+is the tuple <code>(false, true, true)</code>, denoting <code>(noUnderline, lineThrough, overline)</code>, and a
```
nigelmegitt: I don't think the use of booleans helps this example for comprehensibility. Why not just use `(noUnderline, lineThrough, overline)` directly and remove the `(false, true, true)` equivalent? That boolean equivalent is not defined anywhere.

skynavga: Accepted, will change.

Issue #105, opened 2015-11-04
Prefer end to dur on generated out-of-line regions


plehegar:
The new process [generate out-of-line regions] in chapter 9.3.2 always adds a dur attribute to the created region element R' even when dur attributes are not permitted. This situation is explicitly described in chapter 10.2.3 for the case when a document instance specifies the use of the smpte time base and discontinuous marker mode.

I think (needs checking) that this could be resolved by specifying a computed end attribute in the root temporal extent in place of dur. If not, I guess we need to add more logic to carry across the appropriate combination of begin, dur and end for the content depending on the other factors in the document.

(raised by Nigel Megitt on 2014-07-09)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/324

skynavga:
The intent is to specify simple duration, and not active duration. Furthermore, the following applies:

> This section defines the semantics of region layout and presentation in terms of a standard processing model as follows:

> Any implementation is permitted provided that the externally observable results are consistent with the results produced by this model.

The algorithm isn't generating dur for use in serialized documents, but for internal processing, so profiles (especially external profiles) don't matter here.

nigelmegitt:
Just to clarify how this can work, I need to understand in smpte discontinuous what is the computed simple duration of B that is derived for the following bullet:

> add a `dur` attribute to the `[attributes]` information item property of R'  with a value equivalent to the computed simple duration of B;

Are you suggesting that an implementation may have an internal representation of a computed simple duration that is equivalent to "indefinite" in SMIL terminology? I cannot see how else it would work since the active end in smpte discontinuous mode is only defined by observation of the end marker value; the alternative assumption that could be the default in SMIL if not specified otherwise is a simple duration of 0 which would result in the element never being active.

Perhaps it would be worth clarifying this case, since simple duration has no effective meaning or use in smpte discontinuous mode and implementations are forced to wait for an end event in this case.

Reopening since I do not think that the impact of this part of the spec is quite clear enough yet - I cannot determine if it is correct and does not need to be changed or if the presented case is not handled, and would like to leave this open until we have had a chance to discuss it.

nigelmegitt:
[Meeting 2016-10-27] @skynavga indicated that further specification is indeed needed for the smpte discontinuous case.

skynavga:
Address reviewer comments and merge from #309.

PR #309, opened 2017-05-13
Issue 0105 out of line region timing

skynavga:
See #105.


spec/ttml2.xml
```diff
@@ -16300,8 +16300,10 @@ otherwise, let <emph>ID</emph> be the value of the <att>xml:id</att> attribute o
 a value equivalent to the computed begin time of <emph>B</emph> within the root temporal extent;</p>
 </item>
 <item>
-<p>add a <att>dur</att> attribute to the <code>[attributes]</code> information item property of <emph>R'&nbsp;</emph> with
-a value equivalent to the computed simple duration of <emph>B</emph>;</p>
+<p>if not operating in smpte time base and discontinuous marker mode, add a <att>dur</att> attribute to the <code>[attributes]</code>
+information item property of <emph>R'&nbsp;</emph> with value equivalent to the computed simple duration of <emph>B</emph>; otherwise,
+add an <att>end</att> attribute to the <code>[attributes]</code> information item property of <emph>R'&nbsp;</emph> with
+a value equivalent to the computed end time of <emph>B</emph> within the root temporal extent;</p>
```
nigelmegitt: How is it possible to have a value equivalent to the computed end time in smpte discontinuous, given that time computations are not meaningful? Maybe we need to add an explanatory note that this can only be done if there is knowledge of all the timecode values and their expected times or if a local assumption can be made that the timecode values will increase monotonically.

skynavga: Rewrite to better handle smpte discontinuous mode.

nigelmegitt: Thank you!

PR #313, opened 2017-05-16
fix errors in testsuite, ESH Additions plus one other file

btsimonh:
running a parse on all files in the testsuite revealed:
 a number with missing xml headers, 
 some with region definition not in layout

this pull request contains the fixes.

### DIFF
--- a/testsuite/ESH Additions/RealPCPride.wmv.en.xml
+++ b/testsuite/ESH Additions/RealPCPride.wmv.en.xml
@@ -1,4 +1,5 @@
-﻿<tt
+﻿<?xml version="1.0" encoding="utf-8"?>
+<tt
     ttp:profile="http://www.w3.org/ns/ttml#profile-dfxp"
     xmlns="http://www.w3.org/ns/ttml"
     xmlns:ttm="http://www.w3.org/ns/ttml#metadata"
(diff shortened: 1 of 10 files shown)


btsimonh:
I don't make any claim to any of the IPR here :)

skynavga:
Wrong repo obviously.

btsimonh:
really Glenn?  Which repo holds the testsuite for ttml2?

skynavga:
First, there is no testsuite yet for TTML2. Those files (which I will delete in a couple of minutes) were copied from the ttml1 repo when we forked TTML2, and no official work has been done on a TTML2 test suite yet. Second, any changes to the TTML1 test suite must go through the group process. If you think there are errors in the TTML1 test suite, then I suggest you discuss with the group first before posting a PR. Finally, I don't recognize your user name, could you let me know your identity?

palemieux:
@btsimonh FYI, there is a maintained [test suite for IMSC1](https://github.com/w3c/imsc-tests) that includes reference renderings (as PNG images).

nigelmegitt:
Apologies it has taken me a while to catch up with this. @btsimonh you might want to open a new PR on the w3c/ttml1 repo with these changes. I don't think the group process has any issue with this, though of course there may be feedback. Since you have done the work though, and you are willing to contribute it without any IPR constraints, I do not see why we would not accept it. The only exception is if the tests you've fixed were actually designed to identify correct behaviour in the presence of invalid content, but I doubt that is the case!

PR #317, opened 2017-05-17
Add #region-timing feature designation (#169).

### LINKED ISSUE #169: Feature designator for time expressions on region
It would be useful to have a feature designator for time expressions on `region` elements, since some profiles exclude this but cannot express the exclusion other than by an extension.

skynavga:
Closes #169.

### DIFF
--- a/spec/profiles/ttml2-full.xml
+++ b/spec/profiles/ttml2-full.xml
@@ -129,6 +129,7 @@
     <feature value="required">#region-inline</feature>
     <feature value="required">#region-inline-explicit</feature>
     <feature value="required">#region-inline-implicit</feature>
+    <feature value="required">#region-timing</feature>
     <feature value="required">#resources</feature>
     <feature value="required">#ruby</feature>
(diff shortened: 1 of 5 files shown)


skynavga:
On Fri, May 19, 2017 at 3:56 AM, Nigel Megitt <notifications@github.com>
wrote:

> *@nigelmegitt* commented on this pull request.
>
> This looks good to me as it stands but I think we need to go further,
> considering the #timing feature. That feature is not specific about which
> elements have supported timing, and we do not have the converse of
> #region-timing such as #content-timing - if we had that then we could
> make #timing be the union of #content-timing and #region-timing.
>
> @skynavga <https://github.com/skynavga> what would you think about adding
> #content-timing? (I'm not stuck on that name by the way, if you can think
> of a better one - I'm aware that "content" is a bit vague)
>
I'm operating on a presumption that wee cannot change the definitions of
existing TTML1 feature designators. So, for example, I have introduced the
feature #animation-2 to define an extended form of #animation.

Issue #30, opened 2015-11-04
How to specify aspect ratio to understand positioning that may apply for display or video


plehegar:
See TTWG Feb 14 Meeting minutes: http://www.w3.org/2013/02/14-tt-minutes.html,text. 

How to specify aspect ratio to understand positioning that may apply for display or video.

(raised by Glenn Adams on 2013-02-21)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/201

Address any outstanding comments under the following, which are hereby incorporated into this issue:

 - [x] #27 interpreting pixel measure
 - [x] #162 need parameter to express storage aspect ratio

skynavga:
Change the draft ttp:storageAspectRatio to ttp:displayAspectRatio. Deal with over-constrained parameters in the case that DAR, PAR, and extent are all specified.

nigelmegitt:
For information only right now, I'm expecting a related but hopefully orthogonal contribution from an external group on this issue.

skynavga:
Leave open pending review at upcoming F2F.

palemieux:
http://services.w3.org/htmldiff?doc1=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fttml2%2F2cc5eefe8618342d8648eb90ee259d96ab092cc8%2Fspec%2Fttml2.html&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fttml2%2Fgh-pages%2Fspec%2Fttml2.html#root-container-region-semantics

palemieux:
When compositing a document where all lengths are in "%" and which does not specified a PAR, DAR or SAR, against a related video object with DAR=16:9 but a SAR=4:3:

- the root container PAR is computed to "1" (default)
- the root container SAR is computed to "4:3" (related video object)
- the root container DAR is therefore computed to be "4:3"

I would think the default display aspect ratio should be that of the related video object, i.e. the aspect ratio against which the timed text is authored, regardless of the storage aspect ratio that the related video object might use.

skynavga:
On Fri, Jan 6, 2017 at 12:12 PM, Pierre-Anthony Lemieux <
notifications@github.com> wrote:

> When compositing a document where all lengths are in "%" and which does
> not specified a PAR, DAR or SAR, against a related video object with
> DAR=16:9 but a SAR=4:3:
>
>    - the root container PAR is computed to "1" (default)
[...]

PR #321, opened 2017-05-17
Apply improvements to aspect ratio and pixel semantics.

palemieux:
This is an initial pass at rationalizing computation of aspect ratio and pixel dimensions.

Some highlights:

- IMHO the ultimate goal is to determine (a) the intended display aspect ratio of the root container, i.e. what the viewer sees, and (b) the meaning of `px` within the document coordinate system

- the mapping of the root container to the `presentation viewport` is left to profiles and/or applications, but can use the computed PAR, SAR, and DAR.

- the document coordinate system is specified as an abstract Cartesian coordinate system, i.e. independently of the notion of pixels.

I expect additional editorial will be needed, at the very least, but wanted to get the review started.


palemieux:
diff at http://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fttml2%2Fspec%2Fttml2.html&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fttml2%2Fa70a3078b018e5609fa470231182f40e8e06f3b5%2Fspec%2Fttml2.html

skynavga:
I'm making an editing pass on this PR and will commit an update later this evening.

spec/ttml2.xml
```diff
@@ -855,7 +855,7 @@ means of a <att>designator</att> attribute or prose text in a specification of p
 </def>
 </gitem>
 <gitem id="terms-display-aspect-ratio">
-<label>[display aspect ratio]</label>
+<label>[display aspect ratio (or DAR)]</label>
```
skynavga: Acronyms go into the Acronyms section.

spec/ttml2.xml
```diff
@@ -877,10 +877,7 @@ and <loc href="#terms-pixel-aspect-ratio">pixel aspect ratio</loc>.</p>
 <gitem id="terms-document-coordinate-space">
 <label>[document coordinate space]</label>
 <def>
-<p>A logical coordinate space which determines the aspect ratio and resolution of the
-<loc href="#terms-root-container-region">root container region</loc> used to format and present a
-<loc href="#terms-document-instance">document instance</loc>, more about which see
-<specref ref="root-container-region-semantics"/>.</p>
+<p>See <specref ref="root-container-region-semantics"/>.</p>
```
skynavga: The purpose of terminology section is to provide some definition, and not just vector to a new location. So this edit is too radical.

spec/ttml2.xml
```diff
@@ -1314,7 +1311,7 @@ A <loc href="#terms-default-region">default out-of-line region</loc> is implied
 </def>
 </gitem>
 <gitem id="terms-pixel-aspect-ratio">
-<label>[pixel aspect ratio]</label>
+<label>[pixel aspect ratio (or PAR)]</label>
```
skynavga: Ditto on acronyms.

spec/ttml2.xml
```diff
@@ -1508,7 +1505,7 @@ by <specref ref="semantics-style-resolution-processing-sss"/>.</p>
 </def>
 </gitem>
 <gitem id="terms-storage-aspect-ratio">
-<label>[storage aspect ratio]</label>
+<label>[storage aspect ratio (or SAR)]</label>
```
skynavga: Ditto on acronyms.

spec/ttml2.xml
palemieux: Ok.

spec/ttml2.xml
palemieux: Wouldn't the terminology section be the right place to define acronyms?

skynavga: We have an acronyms section. I am creating an edit now to be posted later this evening.

palemieux: I missed the acronym section... so sounds good to move it there.

spec/ttml2.xml
```diff
@@ -5174,7 +5169,7 @@ numerator | demoninator
 </tr>
 </tbody>
 </table>
-<p>If not specified, then square pixels (i.e., aspect ratio 1:1) must be assumed to apply.
+<p>If not specified, then both <code>numerator</code> and <code>demoninator</code> must be assumed to be equal to 1.
```
skynavga: This change is unnecessary and possibly overconstrained. In particular, the values of the numerator and denominator may be some other value than 1/1, e.g., 3/3 is equivalent.

spec/ttml2.xml
```diff
@@ -5401,10 +5396,13 @@ implicit duration of the <el>body</el> element of the document, if the <el>body<
...
 element, then it must adhere to <specref ref="style-attribute-extent"/>, in which case it
 specifies the spatial extent of the <loc href="#terms-root-container-region">root container region</loc> in
-which <loc href="#terms-content-region">content regions</loc> are located and presented. If no <att>tts:extent</att>
+which <loc href="#terms-content-region">content regions</loc> are located and presented.
+<!-- If no <att>tts:extent</att>
 attribute is specified, then the spatial extent of the <loc href="#terms-root-container-region">root container region</loc> is
 considered to be determined by the <loc href="#terms-document-processing-context">document processing context</loc>.
-The origin of the <loc href="#terms-root-container-region">root container region</loc> is determined by the <loc href="#terms-document-processing-context">document processing context</loc>.</p>
+The origin of the <loc href="#terms-root-container-region">root container region</loc> is determined by the <loc href="#terms-document-processing-context">document processing context</loc>.
+ -->
```
skynavga: Can't dispense with this language, am revising slightly to refer to root container region semantics appendix.

spec/ttml2.xml
```diff
@@ -9120,10 +9118,7 @@ represented by this attribute are based upon that defined by <bibref ref="xsl11"
 <p>The <att>tts:extent</att> attribute may be used for the following purposes:</p>
 <olist>
 <item><p>to specify the <loc href="#terms-width"><emph>width</emph></loc> and <loc href="#terms-height"><emph>height</emph></loc>
-of the <loc href="#terms-root-container-region">root container region</loc>,
-which has the effect of defining
-the <loc href="#terms-document-coordinate-space">document coordinate space</loc>,
-about which see also <specref ref="root-container-region-semantics"/>;</p></item>
+of the <loc href="#terms-root-container-region">root container region</loc>;</p></item>
```
skynavga: Revising this change as it removes useful information, while adding nothing.

spec/ttml2.xml
```diff
@@ -14178,16 +14173,19 @@ units
...
+<p>the unit of measure <code>px</code> (pixel) is determined as follows:</p>
+<ulist>
+<item><p>if the <att>tts:extent</att> attribute is specified using <code>px</code> units on the <el>tt</el> element, then:</p>
+<ulist>
+<item><p>one <code>px</code> unit in the horizontal dimension is equal to 100/(root container width in <code>px</code>) <code>rw</code>; and</p></item>
+<item><p>one <code>px</code> unit in the vertical dimension is equal to 100/(root container height in <code>px</code>) <code>rh</code>; and</p></item>
+</ulist>
+</item>
+<item><p>otherwise, a <code>px</code> corresponds with the semantics defined by <bibref ref="xsl11"/>, &sect; 5.9.13, and its mapping to the document coordinate space is undefined.</p></item>
+</ulist>
```
skynavga: Need to vector this definition to root container region semantics sub-section on pixels rather than attempt to define in this context. I will revise accordingly.

spec/ttml2.xml
```diff
@@ -22310,146 +22308,64 @@ it requires support for the
 <div1 id="root-container-region-semantics">
 <head>Root Container Region Semantics</head>
 <p><emph>This appendix is normative.</emph></p>
-<p>This appendix specifies the semantics for determining essential characteristics of the root container region and its use, including:</p>
-<ulist>
-<item><p>aspect ratios (display, storage, pixel)</p></item>
-<item><p>resolution</p></item>
-<item><p>coordinate space</p></item>
-</ulist>
```
skynavga: Revering this removal, as this is useful information for reader.

spec/ttml2.xml
```diff
@@ -22310,146 +22308,64 @@ it requires support for the
...
-resolved resolution of the root container region.
-Lastly, the aspect ratios that apply to <emph>R</emph> are established according to the resolved values of <code>DAR</code>, <code>SAR</code>,
-and <code>PAR</code> as indicated above.</p>
-<p>Unless a <loc href="#terms-higher-level-protocol">higher level protocol</loc> applies,
-the <loc href="#terms-document-coordinate-space">document coordinate space</loc> is determined once and only once when processing a
-given document instance.</p>
+<p>The coordinate space of the root container region, also known as the <loc href="#terms-document-coordinate-space">document coordinate space</loc>, is a cartesian coordinate system with:</p>
+<ulist>
+<item><p>its origin coincident with the upper left-hand corner of the <loc href="#terms-root-container-region">root container region</loc>; and</p></item>
+<item><p>the point with coordinates (1, 1) coincident with the lower right-hand corner of the <loc href="#terms-root-container-region">root container region</loc>.</p></item>
```
skynavga: This is clearly wrong.

palemieux: What is wrong?

skynavga: If the coordinates of the lower right hand corner are <1,1>, then the entire coordinate space has only one pixel. Coordinates are in pixels, not percentages.

skynavga: I am close to committing an update on this branch, so please stand by for further changes.

spec/ttml2.xml
```diff
@@ -22310,146 +22308,64 @@ it requires support for the
...
+  </tr>
+  <tr>
+    <td>no</td>
+    <td>yes</td>
+    <td>SAR x PAR</td>
+    <td>width of tts:extent divided by height of tts:extent</td>
+    <td>computed value of ttp:pixelAspectRatio</td>
+  </tr>
+</tbody>
+</table>
```
skynavga: I was expecting only that the existing specification text would be corrected if there was a substantive difference in technical understand of the desired results as we discussed in the London F2F. I did not expect a wholesale replacement of the existing text with a table that is not only difficult to interpret, but introduces a number of ambiguities, such has how to resolve over-constrained specified values. I am restoring the previous text and then editing it as needed to incorporate the results from the F2F in a manner I hope is consistent with the semantics of your replacement table.

spec/ttml2.xml
palemieux: The point is that the coordinate system should not be in pixels since pixels are meaningless if tts:extent is not specified on `tt`.

spec/ttml2.xml
palemieux: > but introduces a number of ambiguities, 

What ambiguities?

skynavga: See "such as ...".

spec/ttml2.xml
skynavga: But it is in pixels, and has always been so. In the case there is no specifed SAR (via tts:extent), then we can still infer one that permits resolution of percentage and viewport units to those pixels, and thus permits the conceptual model to remain consistent in all scenarios.

palemieux: Glenn, it cannot be in pixels if the root container's dimensions are not in pixels! How is a coordinate of (20px, 20px) resolved if the root container dimensions are (100%, 100%).

spec/ttml2.xml
palemieux: In absence of a specific example, tt sounds like you did not even try to read it.

spec/ttml2.xml
skynavga: You may have forgotten that TTML **always** assumes the coordinate space is in pixels, and that a document can use px units even without specifying SAR, i.e., without specifying ``tts:extent`` on ``<tt>``.

spec/ttml2.xml
skynavga: There you go again, accusing me of bad faith. Of course I read it. Many times.

palemieux: > however, if PAR is also specified, then this is ambiguous in the case that PAR is not DAR/SAR. 

In London, we had decided that DAR/SAR would override the specified PAR value.

palemieux: > Your table also appears to imply that tts:extent on <tt> may be lengths that use non-pixel units,

As it stands it can also be: 1) the auto keyword or (2) the contain keyword.

skynavga: I don't disagree that in that case PAR is overridden by DAR/SAR, but your table doesn't say that, and the reader may be confused. In any case, that was what I meant by ambiguity (as well as the non-pixel lengths on extent).

Yes, I agree that auto and contain are possible values, but your table didn't say that, so a reader might have thought that extent with a non-pixel length could apply.

spec/ttml2.xml
palemieux: > You may have forgotten that TTML always assumes the coordinate space is in pixels, 

Without extent in px when dimensions are expressed in px, it is not possible to determine the relative dimensions of a region with (50%,50%) extent and one with (20px, 20px) extent.

This has been an issue with TTML1, and should be fixed in TTML2.

skynavga: Not really. In this case, the document processing context can choose a resolution, say, e.g., the resolution of a related media object, or in case of no media, an arbitrary resolution, e.g., "1000px 1000px". Having done so, it is straightforward to resolve both [50%,50%] and [20px,20px]. If the results are unexpected, then it is the authors fault for failing to specify a resolution in the first place. We can always add a strong warning against the use of pixels without specifying resolution, but for backwards interoperability, we can't change TTML1 retroactively.

skynavga:
@palemieux edit applied to branch.

skynavga:
Closes #30.

skynavga:
@palemieux do you have any comments on the PR edits? what I'm looking for primarily is if you have any technical issues with the edits with respect to the table you proposed as a replacement, and, more specifically, whether you feel that my incorporation of the table's semantics into the earlier text is consistent with your mental model of resolving aspect ratios.

palemieux:
@skynavga It is on my todo list.

spec/ttml2.xml
nigelmegitt: I don't agree that TTML assumes the coordinate space is in pixels. It may well define semantics should an implementation make that assumption, but it also seems entirely possible (as well as a good idea in some cases) to author and present TTML documents without making references to logical TTML pixels anywhere. For example `rw` and `rh` units, or cell units, can be used throughout, and mapped to a vector image format on output that also only uses relative units.

skynavga:
On Wed, May 24, 2017 at 9:56 AM, Nigel Megitt <notifications@github.com>
wrote:

> *@nigelmegitt* commented on this pull request.
>
> In spec/ttml2.xml
> <https://github.com/w3c/ttml2/pull/321#discussion_r118293175>:
>
> > -<p>The coordinate space of the root container region, also known as the <loc href="#terms-document-coordinate-space">document coordinate space</loc>,
> -is an unbounded, two-dimensional plane on which a closed set of logical pixels are defined that take the form of a rectangle <emph>R</emph> such that
> -the said pixels are interior to (inside of) the <emph>R</emph>, and where the origin (position) of this coordinate space is coincident with the upper,
> -left-hand corner of <emph>R</emph>, and where positive pixels on the vertical axis extend downwards and positive pixels on the horizontal axis
> -extend rightwards.
> -Furthermore, the <loc href="#terms-width">width</loc> and <loc href="#terms-height">height</loc> of <emph>R</emph> is set to the
> -resolved resolution of the root container region.
> -Lastly, the aspect ratios that apply to <emph>R</emph> are established according to the resolved values of <code>DAR</code>, <code>SAR</code>,
> -and <code>PAR</code> as indicated above.</p>
> -<p>Unless a <loc href="#terms-higher-level-protocol">higher level protocol</loc> applies,
> -the <loc href="#terms-document-coordinate-space">document coordinate space</loc> is determined once and only once when processing a
> -given document instance.</p>
> +<p>The coordinate space of the root container region, also known as the <loc href="#terms-document-coordinate-space">document coordinate space</loc>, is a cartesian coordinate system with:</p>
> +<ulist>
> +<item><p>its origin coincident with the upper left-hand corner of the <loc href="#terms-root-container-region">root container region</loc>; and</p></item>
> +<item><p>the point with coordinates (1, 1) coincident with the lower right-hand corner of the <loc href="#terms-root-container-region">root container region</loc>.</p></item>
>
> I don't agree that TTML assumes the coordinate space is in pixels. It may
> well define semantics should an implementation make that assumption, but it
> also seems entirely possible (as well as a good idea in some cases) to
> author and present TTML documents without making references to logical TTML
> pixels anywhere. For example rw and rh units, or cell units, can be used
> throughout, and mapped to a vector image format on output that also only
> uses relative units.
>
The original authors of TTML have always assumed a pixel based coordinate
system. From what I can tell, it is only with your recent arrival in the
group and the focus on percentage based lengths in IMSC that you have
arrived at a different position that is historically justified.

In any case, I do agree with you that it is possible to construct a
document without recourse to pixels, but that does not invalidate the
existing assumption that the document coordinate space is in pixels,
whether or not those pixels are referenced as such.

spec/ttml2.xml
```diff
@@ -5396,19 +5414,14 @@ implicit duration of the <el>body</el> element of the document, if the <el>body<
...
 specifies the spatial extent of the <loc href="#terms-root-container-region">root container region</loc> in
-which <loc href="#terms-content-region">content regions</loc> are located and presented.
-<!-- If no <att>tts:extent</att>
+which <loc href="#terms-content-region">content regions</loc> are located and presented. If no <att>tts:extent</att>
 attribute is specified, then the spatial extent of the <loc href="#terms-root-container-region">root container region</loc> is
-considered to be determined by the <loc href="#terms-document-processing-context">document processing context</loc>.
-The origin of the <loc href="#terms-root-container-region">root container region</loc> is determined by the <loc href="#terms-document-processing-context">document processing context</loc>.
- -->
-</p>
+determine by rules specified in <specref ref="root-container-region-semantics"/>.
```
skynavga: Fixed.

spec/ttml2.xml
```diff
@@ -22308,64 +22314,163 @@ it requires support for the
...
+<p>Three aspect ratios apply to the <loc href="#terms-root-container-region">root container region</loc>:</p>
+<glist>
+<gitem>
+  <label>display aspect ratio (DAR)</label>
+  <def>
+    <p></p>
+    <p>The <emph>display</emph> aspect ratio of the root container corresponds with the displayed aspect ratio of the root container, i.e.,
+    the actual (or intended) physical aspect ratio when it is (to be) presented on a visual medium;
+    this <emph>display</emph> aspect ratio may be specified by the
+    <loc href="#parameter-attribute-displayAspectRatio">ttp:displayAspectRatio</loc> attribute on the
```
skynavga: The two parts work in concert: the first part defines how specified values of aspect ratios are determined, the second part defines how resolved values of aspect ratios are determined. It is a two pass process, like determining specified and then computed values of style properties. If you believe there is a technical inconsistency, then please share that information.

spec/ttml2.xml
```diff
@@ -22308,64 +22314,163 @@ it requires support for the
...
+    <loc href="#document-structure-vocabulary-tt"><el>tt</el></loc> element, or may be inferred using other information as described below.</p>
+  </def>
+</gitem>
+<gitem>
+  <label>storage aspect ratio (SAR)</label>
+  <def>
+    <p></p>
+    <p>The <emph>storage</emph> aspect ratio of the root container corresponds with the logical aspect ratio of the root container
+    in storage or sample units;
```
skynavga: The term storage aspect ratio, as well as its synonyms or near-synonyms, frame aspect ratio, image aspect ratio, and picture aspect ratio, typically make referent to storage units and/or sample units, namely, the units of information that correspond to pixels in the context of a stored frame/image/picture. I suppose we could substitute "logical pixels" in this context, since they have the same intended meaning here.

spec/ttml2.xml
```diff
@@ -22308,64 +22314,163 @@ it requires support for the
...
+    this <emph>storage</emph> aspect ratio may be specified (indirectly) by the
+    <loc href="#style-attribute-extent">tts:extent</loc> attribute on the
+    <loc href="#document-structure-vocabulary-tt"><el>tt</el></loc> element, or may be inferred using other information as described below.</p>
+    <p>When the <loc href="#style-attribute-extent">tts:extent</loc> attribute is specified on the
+    <loc href="#document-structure-vocabulary-tt"><el>tt</el></loc> element, then the <code>SAR</code> is assumed to have been specified as
+    follows:</p>
+    <olist>
```
skynavga: Fix pending (by vectoring definitions in terminology section to Appendix H).

spec/ttml2.xml
```diff
@@ -22308,64 +22314,163 @@ it requires support for the
...
-<p>The following table specifies how the <loc href="#terms-display-aspect-ratio">DAR</loc>, <loc href="#terms-storage-aspect-ratio">SAR</loc> and <loc href="#terms-pixel-aspect-ratio">PAR</loc> of the root container region are resolved.</p>
+<head>Aspect Ratios</head>
+<p>Three aspect ratios apply to the <loc href="#terms-root-container-region">root container region</loc>:</p>
+<glist>
+<gitem>
+  <label>display aspect ratio (DAR)</label>
+  <def>
+    <p></p>
+    <p>The <emph>display</emph> aspect ratio of the root container corresponds with the displayed aspect ratio of the root container, i.e.,
+    the actual (or intended) physical aspect ratio when it is (to be) presented on a visual medium;
```
skynavga: The term *visual medium* is a term of art that goes back to CSS1 961217, where it is assumed and contrasted with *non-visual media*, expanded further in CSS2 980512 [2] as *visual media group*.

I am removing the term presentation viewport (and all other viewport related terms) since they are no longer referenced in the specification text.

[1] https://www.w3.org/TR/REC-CSS1-961217#appendix-e
[2] https://www.w3.org/TR/1998/REC-CSS2-19980512/media.html#media-groups

spec/ttml2.xml
```diff
@@ -22308,64 +22314,163 @@ it requires support for the
...
+      the <code>SAR</code> is considered to be unspecified and is resolved according to the rules specified below.</p></item>
+    </olist>
+  </def>
+</gitem>
+<gitem>
+  <label>pixel aspect ratio (PAR)</label>
+  <def>
+    <p></p>
+    <p>The <emph>pixel</emph> aspect ratio of the root container corresponds with the actual (or intended) physical aspect ratio of
+    each individual physical (as opposed to logical) pixel of the root container when it is (to be) presented on a visual medium;
```
skynavga: Fix pending, by disconnecting *display pixels* from post-TTML processing.

spec/ttml2.xml
```diff
@@ -22308,64 +22314,163 @@ it requires support for the
...
+      <item><p>the resolved value of <code>PAR</code> is <code>1</code> (one).</p></item>
+    </olist>
+  </item>
+  <item>
+    <p>if <code>PAR</code> is specified,
+    then its resolved value is its specified value, and the values of <code>DAR</code> and <code>SAR</code> are determined as follows:</p>
+    <olist>
+      <item><p>if a <loc href="#terms-related-media-object">related media object</loc> exists and has a defined <code>DAR</code>,
+      then the resolved value of <code>DAR</code> is the display aspect ratio of the
+      <loc href="#terms-related-media-object">related media object</loc> and the value of <code>SAR</code>
```
skynavga: Fix pending.

spec/ttml2.xml
skynavga: Resolution pending.

spec/ttml2.xml
skynavga: Replacing with logical pixels in pending fix.

PR #324, opened 2017-05-22
Issue 0164 default play rate

### LINKED ISSUE #164: Default play rate
In N.2 appears:

<pre>
playRate ∈ ℜ, −∞ < playRate < ∞, playRate is unit-less, and where the value of
playRate is determined by the document processing context;
</pre>

This should be enhanced by adding:

<pre>
, or the value 1.0 if so determined
<pre>

skynavga:
Closes #164.


spec/ttml2.xml
```diff
@@ -23128,6 +23128,12 @@ an epoch <code>E</code>, and where the value of <code>epochOffset</code> is the
 </tr>
 </tbody>
 </table>
+<note role="elaboration">
+<p>If the value of <code>playRate</code> is determined by the <loc href="#terms-document-processing-context">document processing context</loc>,
+i.e., it is not set to the default value of <code>1.0</code>, then that value may vary in real time over the course of document
+presentation, in which case future values of <code>R</code> for given values of <code>M</code> are expected to be re-evaluated based on new
```
nigelmegitt: The word "future" here may be misleading - I suggest removing it. It does not account for the fact that the play rate may be set to a negative value by the processing context, e.g. by the user pressing "rewind".

PR #325, opened 2017-05-22
Clarify applicability of special value of 60 in seconds component (#1…

### LINKED ISSUE #158: Clock-time form time expressions allow 60 seconds for media timebase
Equivalent issue to [TTML1 issue 210](https://github.com/w3c/ttml1/issues/210).

There appears to be no restriction on the use of a value of `60` in the seconds component of a clock-time time expression of the form hh:mm:ss. When using `ttp:timeBase="media"` it therefore appears that a minute can have 61 seconds; my understanding of the use case for this is to accommodate UTC leap seconds, so I would expect a restriction that the `60` seconds value can only apply when `ttp:timeBase="clock"`, and possibly additionally when `ttp:clockMode="utc"|"local"` since GPS clocks are not adjusted for leap seconds.

The discussed proposal is in TTML1 to disrecommend 60 seconds according to above and in TTML2 to prohibit 60 seconds as per above.

skynavga:
…58).

Closes #158.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -17330,6 +17330,11 @@ in the preceding syntax definition.</p>
 minutes, seconds, and frames less than 10. Minutes are constrained to
 [0&hellip;59], while seconds (including any fractional part) are constrained to the closed
 interval [0,60], where the value 60 applies only to leap seconds.</p>
+<note role="clarification">
+<p>The special seconds value 60 is applicable only when leap seconds are counted; namely, when
+<loc href="#parameter-attribute-timeBase"><att>ttp:timeBase</att></loc> is <code>clock</code>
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -17330,6 +17330,11 @@ in the preceding syntax definition.</p>
 minutes, seconds, and frames less than 10. Minutes are constrained to
 [0&hellip;59], while seconds (including any fractional part) are constrained to the closed
 interval [0,60], where the value 60 applies only to leap seconds.</p>
+<note role="clarification">
```
nigelmegitt: As per https://github.com/w3c/ttml1/issues/210#issuecomment-212449373 I'm expecting this to be a normative provision in TTML2 not an informative one like in TTML1.

PR #328, opened 2017-05-23
Issue 0133 ruby style inheritance

### LINKED ISSUE #133: Ruby text should used special inheritance algorithm
in order to permit auto-sizing of ruby text annotations, the font size and line height styles should not be inherited on tts:ruby textContainer and, on tts:ruby text, should inherit only from the parent textContainer;

that is, if textContainer is present and no font size or line height is specified on it, then it should not inherit these from its parent tts:ruby container; for tts:ruby text, if no font size or line height is specified, then it may inherit from its parent tts:ruby texContainer (if one exists, otherwise not)

(raised by Glenn Adams on 2015-03-21)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/379

skynavga:
Closes #133.


spec/ttml2.xml
```diff
@@ -9898,6 +9892,35 @@ represented by this attribute are based upon that defined by <bibref ref="xsl11"
...
+either <code>textContainer</code> or <code>text</code>, then, when resolving the computed value
+of <loc href="#style-attribute-fontSize"><att>tts:fontSize</att></loc>, the following special inheritance
+rules apply.</p>
+<p>A <loc href="#content-vocabulary-span"><el>span</el></loc> element for which the computed value
+of <loc href="#style-attribute-ruby"><att>tts:ruby</att></loc> is <code>textContainer</code>
+does not inherit <loc href="#style-attribute-fontSize"><att>tts:fontSize</att></loc> from its parent.
+Instead, if no <loc href="#style-attribute-fontSize"><att>tts:fontSize</att></loc> attribute is specified on the
+ruby text container <loc href="#content-vocabulary-span"><el>span</el></loc> element, a value equal to 50% of the value of the computed value
+of <loc href="#style-attribute-fontSize"><att>tts:fontSize</att></loc> on the nearest ancestor
+<loc href="#content-vocabulary-span"><el>span</el></loc> element for which the computed value
```
palemieux: This is a significant departure from the current inheritance model which does not generally require to go up the element hierarchy.

Can you provide an example of the problem this is solving?

skynavga: The handling of auto-sizing semantics for ruby text. Actually, many types of inheritance in TTML require going "up the element hierarchy", e.g., simple inheritance (goes up to at least parent), xml:lang and xml:space, etc.

palemieux: > e.g., simple inheritance (goes up to at least parent)

That is only the parent.

> xml:lang and xml:space, etc.

Yes. They are a pain to handle. It would be good to avoid other similar cases.

> he handling of auto-sizing semantics for ruby text.

Ok, but can you provide an example document where this would be an issue?

skynavga:
On Tue, May 23, 2017 at 11:52 PM, Pierre-Anthony Lemieux <

>
> In spec/ttml2.xml
> <https://github.com/w3c/ttml2/pull/328#discussion_r118167374>:
>
> > @@ -9898,6 +9892,35 @@ represented by this attribute are based upon that defined by <bibref ref="xsl11"
>  &sect; 7.9.4. The addition of a second length component to permit specifying font
>  width and height independently is an extension introduced by TTML.</p>
>  </note>
> +<div4 id="style-attribute-fontSize-special-semantics">
> +<head>Special Semantics</head>
> +<p>When applied to a <loc href="#content-vocabulary-span"><el>span</el></loc> element
> +for which the computed value of <loc href="#style-attribute-ruby"><att>tts:ruby</att></loc> is
> +either <code>textContainer</code> or <code>text</code>, then, when resolving the computed value
> +of <loc href="#style-attribute-fontSize"><att>tts:fontSize</att></loc>, the following special inheritance
> +rules apply.</p>
> +<p>A <loc href="#content-vocabulary-span"><el>span</el></loc> element for which the computed value
> +of <loc href="#style-attribute-ruby"><att>tts:ruby</att></loc> is <code>textContainer</code>
> +does not inherit <loc href="#style-attribute-fontSize"><att>tts:fontSize</att></loc> from its parent.
> +Instead, if no <loc href="#style-attribute-fontSize"><att>tts:fontSize</att></loc> attribute is specified on the
> +ruby text container <loc href="#content-vocabulary-span"><el>span</el></loc> element, a value equal to 50% of the value of the computed value
> +of <loc href="#style-attribute-fontSize"><att>tts:fontSize</att></loc> on the nearest ancestor
> +<loc href="#content-vocabulary-span"><el>span</el></loc> element for which the computed value
>
> e.g., simple inheritance (goes up to at least parent)
>
> That is only the parent.
>
> xml:lang and xml:space, etc.
>
> Yes. They are a pain to handle. It would be good to avoid other similar
> cases.
>
> he handling of auto-sizing semantics for ruby text.
>
> Ok, but can you provide an example document where this would be an issue?
>
Every document that uses ruby that doesn't explicitly specify fontSize on
every ruby textContainer and ruby text span. For auto sizing you want leaf
ruby text spans to inherit 50% of the computed font size of the nearest
ancestor ruby text container text span, or if there is none, then the
nearest ancestor ruby container span.

It is possible I could express this requirement more succinctly, but I
believe the new text does just that.

palemieux:
> It is possible I could express this requirement more succinctly, but I believe the new text does just that.

Can you provide a test case TTML document and show what happens with and without the special inheritance?

skynavga:
Without special inheritance, ruby would be sized the same as base, but it should be 50% of base (by default). If we say that textContainer and text inherit 50% of parent, then 50% will be applied twice if both ruby text container and ruby text are used, ending up with 25%. But 50% would be applied if only ruby text is used. Special rules are clearly required.

palemieux:
>  Special rules are clearly required.

Maybe, but I think it is reasonable to ask for what will ultimately be a test case since this part of the specification is extremely complex. Surely, Skynav has generated such an example for its internal testing?

skynavga:
Basically, any file with ruby that doesn't specify font size on ruby text
container or ruby text spans qualify as a test case. See, for example, [1].
The test is to verify that the font size applied to text is correct in the
auto-sizing cases, i.e., that 50% of ruby base font size is used, and not
100% or 25%.

[1]
https://github.com/skynav/ttt/blob/master/ttt-ttpe/src/test/resources/com/skynav/ttpe/ttml2/ttml2-ruby-on-mixed-orientations.xml

palemieux:
@skynavga Thanks for the example.

Is the proposal equivalent to specifying `<style xml:id="s4" tts:fontSize="50%" tts:ruby="text" tts:rubyPosition="before"/>`?

skynavga:
On Wed, May 24, 2017 at 11:40 AM, Pierre-Anthony Lemieux <

> @skynavga <https://github.com/skynavga> Thanks for the example.
>
> Is the proposal equivalent to specifying <style xml:id="s4"
> tts:fontSize="50%" tts:ruby="text" tts:rubyPosition="before"/>?
>
It depends. Consider

*Case 1 - Auto Size*

<span tts:ruby="contatiner" tts:fontSize="24px">
  ...
  <span tts:ruby="text">ruby</span>
</span>

font size on ruby should be 12px (50% of ruby container), not 24px (100% of
ruby container)

*Case 2 - Auto Size*

<span tts:ruby="contatiner" tts:fontSize="24px">
  ...
  <span tts:ruby="textContainer">
    <span tts:ruby="text">ruby1</span>
    <span tts:ruby="text">ruby2</span>
  </span>
</span>

font size on ruby should be 12px (50% of ruby container), not 24px (100% of
ruby container) or 6px (25% of ruby container)

*Case 3 - Partial Auto Size*

<span tts:ruby="contatiner" tts:fontSize="24px">
  ...
  <span tts:ruby="textContainer" tts:fontSize="18px">
    <span tts:ruby="text">ruby1</span>
    <span tts:ruby="text">ruby2</span>
  </span>
</span>

font size on ruby should be 18px (100% of ruby text container), not 24px
(100% of ruby container) or 12px (50% of ruby container)

*Case 4 - Partial Auto Size*

<span tts:ruby="contatiner" tts:fontSize="24px">
  ...
  <span tts:ruby="textContainer">
    <span tts:ruby="text" tts:fontSize="18px">ruby1</span>
    <span tts:ruby="text">ruby2</span>
  </span>
</span>

font size on ruby1 should be 18px, font size on ruby2 should be 12px (50%
of ruby container)

PR #333, opened 2017-05-25
Add rendering model that defines rendering (painting) order (#228).

### LINKED ISSUE #228: What is the painting order for image, text, backgroundImage and backgroundColor?
In the case that a `span`, for example has more than one of:

* non-transparent `tts:backgroundColor`
* text content
* a tts:backgroundImage
* an `<image>` child (i.e. a "foreground" image)

what is the painting (z) order? Or better yet, where is the painting order specified?

skynavga:
Closes #228.


spec/ttml2.xml
```diff
@@ -17009,6 +17009,17 @@ href="http://www.unicode.org/reports/tr14/#Algorithm">Line
...
 </div3> <!-- semantics-line-layout -->
+<div3 id="semantics-rendering-model">
+<head>Rendering Model</head>
+<p>The rendering order employed by a <loc href="#terms-presentation-processor">presentation processor</loc> must be consistent with
+the rendering model specified by <bibref ref="xsl11"/>, &sect;4.9; in addition, the following constraints apply:</p>
+<ulist>
+  <item><p>inline area descendants of a line area that are generated by ruby text spans are rendered
+  after other inline area descendants of the line area;</p></item>
+  <item><p>the intrinsic marks of a block area or an inline area generated by an
+  <loc href="#embedded-content-vocabulary-image"><el>image</el></loc> element consist of the raster of the associated image.</p></item>
```
nigelmegitt: Does this apply to an image used in an image definition context and referenced by a backgroundImage or only to image used in an image presentation context?

skynavga: We don't make any normative reference to CSS rendering model, so have to define in terms of XSL-FO. However, if someone thinks there is a discrepancy, they may point it out.

Images in an image definition context have no presentation semantics, so clearly it does not apply in that context.

nigelmegitt: That doesn't really answer the question though - this could be read as saying that when a backgroundImage is specified (via indirect reference to an `image` element in a definition context) then no other marks are present. I think it needs to say that when `backgroundImage` is used that defines the background area, drawn before other intrinsic marks and replacing any `backgroundColor`, but when `image` is used in an image presentation context then the current wording applies - the area consists only of the raster of the associated image.

skynavga: I can't fathom how anyone could read it that way. The referenced material in XSL-FO clearly states that

>Areas generate three types of marks: (1) the area background, if any, (2) the marks intrinsic to the area (a glyph, image, or decoration) if any, and (3) the area border, if any.

which makes it clear that the area's background (including a background image) and the area's intrinsic marks are distinct. How a background image or a foreground image are referenced, obtained, determined to be visible, etc., are outside the scope of the rendering model, i.e., those decisions come before the rendering model is employed.

PR #336, opened 2017-05-26
Improve descriptive prose and elaborate computed value semantics of t…

### LINKED ISSUE #200: Two value percentage fontSize not fully defined
If a two value percentage based `tts:fontSize` is used and an ancestor element specifies a single value absolute size, the horizontal component is undefined. For example:

``` xml
<div tts:fontSize="1c">
   <p tts:fontSize="100% 100%">...</p>
</div>
```

In this case the `div` sets a font size defined by height alone, so width is undefined. Then the first (horizontal) 100% value in the `p` element's `tts:fontSize` attribute is 100% of an unknown value. This cannot meaningfully be transformed by a transformation processor.

Solution proposals:
1. Prohibit two value percentage based sizes (simplest solution)
2. Prohibit two value percentage based sizes when the parent element's font size has a single value (harder to verify)
3. Define the applicable size to use when calculating the horizontal component of a two value percentage based font size if the parent element font size has a single value. (most complicated solution)

I would prefer the first of these on first look.

skynavga:
…ts:fontSize (#200).

Closes #200.


spec/ttml2.xml
```diff
@@ -9829,6 +9848,17 @@ represented by this attribute are based upon that defined by <bibref ref="xsl11"
...
 </note>
+<div4 id="style-attribute-textSize-special-semantics">
+<head>Special Semantics</head>
+<p>The computed value of the property associated with this attribute consists of a 2-tuple
+the entries of which denote the computed values of the width and height components of the font size respectively.
+Furthermore, when applying inheritance semantics, each of these entries is considered to
+be inherited separately from the other.</p>
+<note role="elaboration">
+<p>Notwithstanding the separate inheritance of width and height components of the computed font size,
+there is, at present, no mechanism that permits specifying only one component.</p>
```
nigelmegitt: I don't understand the value of stating that each entry in the tuple is inherited separately given that no meaningful use can be made of it. Is this for some future compatibility, or to guide implementations, or to allow for a single mechanism for tuple inheritance to be defined once and reused both here and for example in `tts:textDecoration`?

spec/ttml2.xml
```diff
@@ -9742,33 +9742,61 @@ then relative to nearest ancestor <loc href="#terms-styled-element">styled eleme
...
-span.</p>
+<p>If the value of this attribute specifies a <code>c</code> unit on a <loc href="#style-value-length">&lt;length&gt;</loc> component,
+then it is resolved in terms of the height (only) or (both) the width and height of a
+<emph>cell</emph> length unit as defined by <specref ref="style-value-length"/>, which is to say, in terms of
+the <loc href="#terms-computed-cell-size">computed cell size</loc>.</p>
+<note role="example">
+<p>For example, consider a paragraph (<code>p</code>) element <code>P</code>.
+If the <loc href="#terms-computed-cell-size">computed cell size</loc> is (24px,36px), and if
+<code>tts:fontSize="1c"</code> is specified on <code>P</code>, then the computed value of <code>tts:fontSize</code>
+resolves to (36px,36px), while if <code>tts:fontSize="1c 1c"</code> is specified on <code>P</code>,
```
nigelmegitt: I do not think this example is correct; specifically in the width component. My expectation here is that a single length value results in a tuple of `{unresolved, height}`, not `{width, height}`. In other words the computed value of `tts:fontSize` is not necessarily a logical pixel equivalent of the em square in width and height.

This is important because the PAR might not be 1. For example, if the PAR is 1.5 and the font size height is 36px then it would be reasonable for the implementation to deduce that the equivalent width of the em square in logical pixels is 24px; note that PAR is not guaranteed to be known at document authoring time. Instead the document author is saying "render some text without anamorphic scaling, so that it is this high." That's an important semantic that we remove by saying that the em square is _always_ scaled to a specific width in logical pixels.

spec/ttml2.xml
```diff
@@ -9742,33 +9742,61 @@ then relative to nearest ancestor <loc href="#terms-styled-element">styled eleme
...
 <p>A glyph's EM square is conventionally defined as the EM square of the font that contains the glyph. That is,
 glyphs do not have an EM square that is distinct from their font's EM square.</p>
 </note>
-<p>If horizontal and vertical sizes are expressed independently, then the
-units of the <loc href="#style-value-length">&lt;length&gt;</loc> values must be the same.</p>
+<p>Relative <loc href="#style-value-length">&lt;length&gt;</loc> values
+that appear in this attribute, i.e., values expressed in percentage (<code>%</code>), cell (<code>c</code>), or EM (<code>em</code>) units,
+are resolved in relation to a referenced 2-dimensional size value consisting of two components, a width component and
+a height component. If the relative unit is a percentage, then the referenced value is either the computed value
+of the font size of the nearest styled ancestor element or the <loc href="#terms-computed-cell-size">computed cell size</loc>.
```
skynavga: Let E be a content element like ``p`` or ``span``. Assuming we are talking about percentages, if no ancestor of element E specifies a font size, then the most distant styled ancestor, namely ``<tt>``, will have been assigned a computed value by resolving the initial value of font size ``1c`` into a specific length in pixels as determined by the height of the computed cell size (CCS). In other words, the computed value of ``tts:fontSize`` on the root element will be the tuple (height(CCS),height(CCS)) in pixels.

For example, if the root container region semantics resolves the document coordinate space to a resolution of (640px,480px), and if ``ttp:cellResolution`` is (32,20) then CCS is equal to (20px,24px). Consequently, the computed value of ``tts:fontSize`` on ``<tt>`` is resolved to (24px,24px).

This computed value is eventually inherited by the nearest styled ancestor A of E. Now, if E doesn't specify a font size, then it inherits this same value. If E specifies a font size in percentage, say the single value 100% (or an equivalent single value 1em), then this single percentage length is resolved by reference to the nearest ancestor styled element, namely A, and since the specified value consists of a single length, then that length is resolved in reference to the height component of the computed value of font size on A, which is also 24px, so the computed value of font size on E is (24px,24px).

The referenced value for resolving percentages is already defined in the *Percentages* line of the table, so we know it is resolved in reference to the nearest styled ancestor. The only time that CCS is used directly is when resolving a specified length that uses cell units or when resolving the computed value of font size on ``<tt>`` in the case that the initial value of font size is specified in cell units (as is the case when the default initial value applies).

spec/ttml2.xml
skynavga: I have no idea how you came to that conclusion, as it is justified by nothing in TTML, XSL-FO, or CSS. TTML1 (and earlier) clearly say:

>If a single <length> value is specified, then this length applies equally to the horizontal and vertical size of a glyph's EM square.

How can you read that and conclude that the horizontal size (width) of the EM square is unresolved. Furthermore, what does PAR have to do with this discussion. We are speaking of logical pixels here only. Discussing PAR (and corresponding display pixels) is entirely orthogonal.

The bottom line is that if one length is specified or if two lengths are specified whether the values are the same, then the EM Square is, well, square. However, if two unequal lengths are specified, than an EM Square is really a non-square rectangle. Mapping this square to display pixels (based on pixels having a PAR aspect ratio) is unrelated to font sizes or lengths based on font sizes.

As for your example of PAR of 1.5 with font size 36px, then the EM square is 36px by 36px in logical pixels which maps to a 36px by 36px rectangle in display pixels. In both cases, the EM square is 36px by 36px. However, in logical pixels we can't talk about aspect ratio (or if you insist, then you can just declare they are square), but in display pixels, now we can certainly talk about aspect ratio of pixels (=PAR). So a 36px by 36px box (in hypothetical square logical pixels) now becomes a 36px by 36px rectangle with a width to height ratio of 1.5 in display pixels.

spec/ttml2.xml
```diff
@@ -9742,33 +9742,61 @@ then relative to nearest ancestor <loc href="#terms-styled-element">styled eleme
...
+resolves to (36px,36px), while if <code>tts:fontSize="1c 1c"</code> is specified on <code>P</code>,
+then the computed value resolves to (24px,36px).</p>
+</note>
 <p>If the value of this attribute specifies an <code>em</code> unit on a <loc href="#style-value-length">&lt;length&gt;</loc> component,
 then it is treated as if an equivalent percentage value were specified, where <code>1em</code> is equal to <code>100%</code>.</p>
+<note role="example">
+<p>For example, consider a span element <code>S</code>, a child of a paragraph (<code>p</code>) element <code>P</code>.
+If the computed value of <code>tts:fontSize</code> on <code>P</code> is (18px,24px), and if
+<code>tts:fontSize="1em"</code> is specified on <code>S</code>, then this is equivalent to specifying <code>100%</code>,
+which resolves to (24px,24px). However, if <code>tts:fontSize="1em 1em"</code> is specified on <code>S</code>,
```
skynavga: The definition of ``em`` going back to early DFXP days, still found in TTML2 says:

>when specified relative to a font whose size is expressed as two length measures of non-equal lengths, then one em is equal to the inline progression dimension of the anamorphically scaled font when used to specify lengths in the inline progression direction and equal to the block progression dimension of the scaled font when used to specify lengths in the block progression direction

So, any value of ``tts:fontSize`` that consists of two values is resolved in terms of a reference width and reference height. All computed font sizes as well as the computed cell size consist of a width and height component. The meaning of ``1em 1em``, ``100% 100%``, and ``1c 1c``, all refer separately to these width and height components. This is also consistent with the use of cells in inline and block progression dimensions, although, with font sizes, we always equate inline dimension with width and block dimension with height, i.e., horizontal and vertical axes, regardless of writing mode.

spec/ttml2.xml
skynavga: I was attempting to be consistent with language found under ``tts:textDecoration``, but I would be willing to remove the second sentence, "Furthermore, ..." and the note if you prefer.

spec/ttml2.xml
```diff
@@ -12919,9 +12949,9 @@ represented by this attribute are based upon that defined by <bibref ref="xsl11"
 <div4 id="style-attribute-textDecoration-special-semantics">
 <head>Special Semantics</head>
 <p>The computed value of the property associated with this attribute consists of a 3-tuple,
-each of which denotes, respectively, whether an underline, line through, or overline
+where each ewntry denotes, respectively, whether an underline, line through, or overline
```
skynavga: Thanks.

spec/ttml2.xml
skynavga: Removed the second sentence and not.

spec/ttml2.xml
skynavga: Fixed.

skynavga:
@nigelmegitt please suggest any specific clarifications or improvements you would like to see made here so we can merge this PR

spec/ttml2.xml
nigelmegitt: > ... the most distant styled ancestor, namely `<tt>`

It is surprising to refer to the `tt` element as a styled element - possibly this can be clarified by simply listing the styled elements somewhere and referencing that list in the definition of `styled element`.

My expectation was that the `region` element is the top of the style element tree based on the style resolution section.

> In other words, the computed value of tts:fontSize on the root element will be the tuple (height(CCS),height(CCS)) in pixels.

Which pixels are you referring to here, logical or display?

spec/ttml2.xml
nigelmegitt: > How can you read that and conclude that the horizontal size (width) of the EM square is unresolved.

It is unresolved at _authoring_ time. Also it is a length here, not a "number of pixels" and they are different things depending on PAR. At best the meaning of "length" differs depending on perspective - a person using non-technical terms would think that you might measure the horizontal length and the vertical length with a ruler on the display and find that they come out the same. In other words, it is not clear that it is meant in a technical sense of "number of pixels".

Reading your comment above, it means that when PAR is not 1, even a single value font size, say `"100%"` always results in an anamorphically scaled em square on the presentation device. This is a surprise to me and not at all what I think most people would understand from TTML1.

spec/ttml2.xml
nigelmegitt: OK thanks for the elaboration.

spec/ttml2.xml
skynavga: You seem to have forgotten about Root Style Inheritance [1]. The root element is now the top level styled element from which all regions inherit.

[1] https://w3c.github.io/ttml2/spec/ttml2.html#semantics-style-inheritance-root

Logical pixels of course.

spec/ttml2.xml
skynavga: First, we are only talking about presentation time here. Authoring time is irrelevant. Second, we are talking about logical pixels here and PAR has no relevance. Length in TTML2 never means a physical, measurable distance.

An "EM square" (in TTML2 measurement terms, i.e., logical pixels) is indeed non-square when PAR is not 1 (unless another transformation to aspect ratio occurs post-TTML output that inverts PAR (but that is out of scope).

spec/ttml2.xml
nigelmegitt: True, I had forgotten about root style inheritance.

spec/ttml2.xml
nigelmegitt: > Authoring time is irrelevant. 

Clearly for every document author authoring time is completely relevant. You can't just sweep it out of scope of the semantics we specify. By the logic put forward here it is actually _impossible_ for a document author to specify "I only want non-anamorphic font display", which is the usual semantic on the web platform; further almost all implementations now would fail to be conformant unless PAR is 1, which cannot be relied upon unfortunately. These two things can't be what we want.

> Length in TTML2 never means a physical, measurable distance.

That is why I preferred to refer to the horizontal dimension as "undefined".

skynavga: First, it is a general truth that an author does not know whether the eventual display will be anamorphic or not, and we cannot address this problem. As we now note in H.1 [1]:

>The actual, physical presentation of the pixels of the root container may be subject to numerous transformations in aspect ratio, sample resolution, and color space subsequent to all defined TTML presentation processing. Such post-TTML processing is wholly out of scope of this specification.

However, the author **can** state they want display pixels that are square. They merely need to specify ``ttp:pixelAspectRatio="1 1"``. Whether the actual presentation device requests this or not is out of scope.

Regarding

>That is why I preferred to refer to the horizontal dimension as "undefined".

You keep insisting on thinking in display pixels. It is most definitely not undefined in logical pixels and that is what every discussion of length and measure assumes in TTML2.

[1] https://w3c.github.io/ttml2/spec/ttml2.html#root-container-region-semantics-aspect-ratios

nigelmegitt: > However, the author **can** state they want display pixels that are square. They merely need to specify `ttp:pixelAspectRatio="1 1"`. Whether the actual presentation device repsects this or not is out of scope.

No, that's orthogonal to the semantic of "just set the height"; furthermore since PAR is given the least priority in H.1.4, it is possible that even if the author intends a particular PAR it will not be the actual one that applies.

What is the impact of anamorphically scaling (ideographic) glyphs in top to bottom writing modes? It's not my area of expertise but it seems as though the width is not nearly as important as the height in that case, but an unintended horizontal squeeze or stretch could adversely impact the readability.

nigelmegitt:
I see this has been merged mid-discussion. That's rather unfortunate to say the least.

nigelmegitt:
Discussed during meeting today 2017-06-01, agreed no further change needed on this pull request, however @skynavga has an entry on his to do list to clarify that logical pixels are intended, and review all use of the term pixel to make sure the correct meaning is used.

spec/ttml2.xml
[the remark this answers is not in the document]
nigelmegitt: See also https://github.com/w3c/ttml1/issues/214#issuecomment-345306822 in which I work an example that shows that the mere introduction of root style inheritance to an otherwise unmodified document means that if processed using TTML2 semantics the result can be different from that using TTML1 semantics. Not all documents define which set of semantics is expected to be used.

spec/ttml2.xml
```diff
@@ -9774,26 +9774,57 @@ then relative to nearest ancestor <loc href="#terms-styled-element">styled eleme
...
 <p>The <loc href="#style-value-length">&lt;length&gt;</loc> value(s) used to express font size must be non-negative.</p>
-<p>For the purpose of determining applicability of this style property,
-each character child of a <el>p</el> element is considered to be enclosed in an anonymous
-span.</p>
+<p>If the value of this attribute specifies a <code>c</code> unit on a <loc href="#style-value-length">&lt;length&gt;</loc> component,
+then it is resolved in terms of the height (only) or (both) the width and height of a
+<emph>cell</emph> length unit as defined by <specref ref="style-value-length"/>, which is to say, in terms of
+the <loc href="#terms-computed-cell-size">computed cell size</loc>.</p>
+<note role="example">
+<p>For example, consider a paragraph (<code>p</code>) element <code>P</code>.
```
nigelmegitt: This example is inconsistent in presentation with the one on span, at line 9822 - why are "paragraph" elements augmented with a "(`p`)" whereas "span" elements are not?

spec/ttml2.xml
```diff
@@ -9774,26 +9774,57 @@ then relative to nearest ancestor <loc href="#terms-styled-element">styled eleme
...
-<loc href="#style-value-length">&lt;length&gt;</loc> values are specified, then the first expresses the horizontal
-scaling and the second expresses vertical scaling.</p>
-<note role="example">
-<p>Use of independent horizontal and vertical font sizes is expected to be used
-with cell based units in order to denote fonts that are two rows in height and
-one column in width.</p>
-</note>
+equally to the horizontal and vertical size of a glyph's EM square;
+if two <loc href="#style-value-length">&lt;length&gt;</loc> values are specified, then the first expresses the horizontal
+size and the second expresses vertical size.</p>
```
nigelmegitt: This loses the sense that the font-defined glyph generally needs scaling to the precise display dimensions of the em square, and specifically always needs scaling anamorphically to achieve a different width to the height. I think this needs to be brought back in somehow, though I like the idea of specifying the resultant size of the em square unambiguously.

spec/ttml2.xml
```diff
@@ -9774,26 +9774,57 @@ then relative to nearest ancestor <loc href="#terms-styled-element">styled eleme
...
-<p>For the purpose of determining applicability of this style property,
-each character child of a <el>p</el> element is considered to be enclosed in an anonymous
-span.</p>
+<p>If the value of this attribute specifies a <code>c</code> unit on a <loc href="#style-value-length">&lt;length&gt;</loc> component,
+then it is resolved in terms of the height (only) or (both) the width and height of a
+<emph>cell</emph> length unit as defined by <specref ref="style-value-length"/>, which is to say, in terms of
+the <loc href="#terms-computed-cell-size">computed cell size</loc>.</p>
+<note role="example">
+<p>For example, consider a paragraph (<code>p</code>) element <code>P</code>.
+If the <loc href="#terms-computed-cell-size">computed cell size</loc> is (24px,36px), and if
```
nigelmegitt: As per the discussion in https://github.com/w3c/ttml2/issues/200#issuecomment-305547161 the sizes here are in logical pixels - in fact they are dimensionless until multiplied by the size of the pixel unit, whatever that turns out to be. Consider adding a parenthetical _(in logical pixels)_ to avoid confusion with "real" pixels.

spec/ttml2.xml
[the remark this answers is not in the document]
nigelmegitt: Note for posterity that root style inheritance was removed, thus dealing with the concern raised in this thread following https://github.com/w3c/ttml2/pull/336#discussion_r118701776.

PR #337, opened 2017-05-26
Clarify that height of referenced value is used to resolve relative l…

### LINKED ISSUE #217: Cell metric in tts:textOutline
In Section 10.2.47:

> When a <length> expressed in cells is used in a tts:textOutline value, the cell's dimension in the block progression dimension applies. 

Is the intent for for the computed value of an outline thickness specified in "c" to depend on writingMode (to determine block progression direction)? Note that the example does not provide a value for writingMode.

I would think it would be unexpected to have an outline thickness of "1c" yield different results across regions depending on writingMode, even though cellResolution remains unchanged by definition.

Should the prose read _When a <length> expressed in cells is used in a tts:textOutline value, the computed cell height applies._?

skynavga:
…engths in tts:textOutline (#217).

Closes #217.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -13170,11 +13170,13 @@ present, then it denotes the outline color; if no <emph>color</emph> term is
 present, the computed value of the <att>tts:color</att> applies.  The first
 <emph>length</emph> term denotes the outline thickness and the second length
 term, if present, indicates the blur radius.</p>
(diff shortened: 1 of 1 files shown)


nigelmegitt:
This was a substantive (normative) change so was merged against agreed process - @skynavga please stick to your own 3 day rule.

@palemieux this pull request addresses an issue you raised, so calling it to your attention.

Issue #213, opened 2016-10-30
Avoid populating non-applicable, non-inheritable styles in SSS(E).


palemieux:
_Specified Style Set Processing_ applies the initial value of styling attributes only when _P is not inheritable or E is the tt element_. There are however styling attributes that are inheritable and have initial values, e.g. color.

The prose also appears substantially different than that of TTML1, where _if the element type of E is the layout element type region, then set P′ to the initial value of property P_. 

Note: In the TTML1 case, it is not clear how initial values of non-inheritable styling attributes ever apply to elements other than regions.

skynavga:
> Specified Style Set Processing applies the initial value of styling attributes only when P is not inheritable or E is the tt element. There are however styling attributes that are inheritable and have initial values, e.g. color.

This is not a bug. Perhaps you don't recall that TTML2 introduces root style inheritance [1]. In particular, regions now inherit style from the root element (tt). As a consequence, all inheritable styles are initialized on tt and then inherited downwards toward region and thence to content elements.

[1] http://w3c.github.io/ttml2/spec/ttml2.html#semantics-style-inheritance-root

> The prose also appears substantially different than that of TTML1, where if the element type of E is the layout element type region, then set P′ to the initial value of property P.

That is correct. It has changed due to the introduction of root style inheritance which wasn't in TTML1.

> Note: In the TTML1 case, it is not clear how initial values of non-inheritable styling attributes ever apply to elements other than regions.

The description in TTML1 was incomplete in this regard. We already have an issue open to cover this need for a new erratum [2].

[2] https://github.com/w3c/ttml1/issues/200

skynavga:
Closing as this is not a bug or actionable issue.

palemieux:
> As a consequence, all inheritable styles are initialized on tt and then inherited downwards toward region and thence to content elements.

Ok. In 10.4.4.2(6.c), the initial value of non-inheritable P should not be applied to E if P does not apply to E, right?

In other words, shouldn't 10.4.4.2(6.c) read:

`if either (P is not inheritable and P applies to E) or (P is inheritable and E is the tt element), perform the following`

skynavga:
[...]

PR #338, opened 2017-05-26
Issue 0221 writing mode and direction

### LINKED ISSUE #221: writingMode and direction interaction
Style inheritance propagates `direction="ltr"` (the initial value) on all elements (unless explicitly specified as "rtl" somewhere), thereby overriding any right-to-left direction specified using `writingMode` on region, since, as specified in Section 7.29.1 of XSL, `direction` takes precedence over `writingMode`.

As specified in Section 7.29.1 of XSL, the TTML style resolution process needs to specify that, whenever specified, `writingMode` also sets the value of `direction` to the horizontal direction implied by `writingMode`.

See also https://lists.w3.org/Archives/Public/public-tt/2016Nov/0027.html

skynavga:
Closes #221.


spec/ttml2.xml
```diff
@@ -8834,6 +8834,26 @@ The title of the book is
 represented by this attribute are based upon that defined by <bibref ref="xsl11"/>,
 &sect; 7.29.1.</p>
 </note>
+<div4 id="style-attribute-direction-special-semantics">
+<head>Special Semantics</head>
+<p>When performing style resolution processing (see <specref ref="semantics-style-resolution-processing"/>),
+the resolution of the computed value of <att>tts:direction</att> on a <loc href="#layout-vocabulary-region"><el>region</el></loc> element
```
palemieux: even though it applies to `region` only, `tts:writingMode` can be specified on any element, so this should read _on an element_ instead of _region element_.

skynavga:
On Fri, May 26, 2017 at 1:52 PM, Pierre-Anthony Lemieux <

>
> In spec/ttml2.xml
> <https://github.com/w3c/ttml2/pull/338#discussion_r118778049>:
>
> > @@ -8834,6 +8834,26 @@ The title of the book is
>  represented by this attribute are based upon that defined by <bibref ref="xsl11"/>,
>  &sect; 7.29.1.</p>
>  </note>
> +<div4 id="style-attribute-direction-special-semantics">
> +<head>Special Semantics</head>
> +<p>When performing style resolution processing (see <specref ref="semantics-style-resolution-processing"/>),
> +the resolution of the computed value of <att>tts:direction</att> on a <loc href="#layout-vocabulary-region"><el>region</el></loc> element
>
> even though it applies to region only, tts:writingMode can be specified
> on any element, so this should read *on an element* instead of *region
> element*.
>
No. See my immediately previous comment. tts:writingMode would be ignore in
that context: it doesn't apply and isn't inherited.

palemieux:
> No. See my immediately previous comment. tts:writingMode would be ignore in
> that context: it doesn't apply and isn't inherited.

Can you point to the provisions that state this?

skynavga:
On Fri, May 26, 2017 at 1:58 PM, Pierre-Anthony Lemieux <

> No. See my immediately previous comment. tts:writingMode would be ignore in
> that context: it doesn't apply and isn't inherited.
>
> Can you point to the provisions that state this?
>
Are you asking what "applies to" means? I would think it obvious, as
indicated by the title of your issue #213:

Avoid populating non-applicable, non-inheritable styles in SSS(E)


"Applies to" is used in TTML (and XSL-FO) in the common English sense of
"is used with or on", etc., Its negation means it "is not used with or on"
meaning "is ignored" for the purpose of ascribing some characteristic or
behavior.

But if your are asking for textual evidence, I see the following in XSL-FO:

The first step in refinement of a particular formatting object is to obtain
> the effective value of each property that applies to the object.


Note that TTML, having been based on XSL-FO semantics, draws widely from
the language used there. For example, in [1], which enumerates which
elements each property "applies to". The same terminology is used in CSS.

[1] https://www.w3.org/TR/xsl/#prapply

Frankly, I am a little surprised by your question.

skynavga:
On Fri, May 26, 2017 at 2:51 PM, Glenn Adams <glenn@skynav.com> wrote:

>
>
> On Fri, May 26, 2017 at 1:58 PM, Pierre-Anthony Lemieux <
> notifications@github.com> wrote:
>
>> No. See my immediately previous comment. tts:writingMode would be ignore
>> in
>> that context: it doesn't apply and isn't inherited.
>>
>> Can you point to the provisions that state this?
>>
> Are you asking what "applies to" means? I would think it obvious, as
> indicated by the title of your issue #213:
>
> Avoid populating non-applicable, non-inheritable styles in SSS(E)
>
>
See also my comment
https://github.com/w3c/ttml2/issues/213#issuecomment-257441803. I have just
started working
on a resolution of this issue (#213), so I will go out of my way to make it
clear that if a property does not apply to an element and if it is not
inheritable, then if it is specified on that element, it is ignored and not
populated into SSS(E).

palemieux:
> so I will go out of my way to make it
clear that if a property does not apply to an element and if it is not
inheritable, then if it is specified on that element, it is ignored and not
populated into SSS(E).

That is not the right solution since it changes the style resolution rules.

It is much simpler to simply state that the exception applies to all elements.

skynavga:
On Fri, May 26, 2017 at 3:43 PM, Pierre-Anthony Lemieux <

> so I will go out of my way to make it
> clear that if a property does not apply to an element and if it is not
> inheritable, then if it is specified on that element, it is ignored and not
> populated into SSS(E).
>
> That is not the right solution since it changes the style resolution rules.
>
> It is much simpler to simply state that the exception applies to all
> elements.
>
I will give this some thought. It would definitely be the wrong approach to
call out tts:writingMode specifically in any such language that clarifies
what the negation of "applies to" means.

palemieux:
> It would definitely be the wrong approach to
call out tts:writingMode specifically in any such language that clarifies
what the negation of "applies to" means.

Yes -- I had actually deleted my comment since addressing #213 might clarify this :)

Regardless of the disposition of #213, it might be safer to extend the special resolution semantics of `tts:direction` + `tts:writingMode` to all elements (I do not see how it can hurt).

skynavga:
On Fri, May 26, 2017 at 4:29 PM, Pierre-Anthony Lemieux <

> It would definitely be the wrong approach to
> call out tts:writingMode specifically in any such language that clarifies
> what the negation of "applies to" means.
>
> Yes -- I had actually deleted my comment since addressing #213
> <https://github.com/w3c/ttml2/issues/213> might clarify this :)
>
> Regardless of the disposition of #213
> <https://github.com/w3c/ttml2/issues/213>, it might be safer to extend
> the special resolution semantics of tts:direction + tts:writingMode to
> all elements (I do not see how it can hurt).
>
ok, i'll consider it and do so if it doesn't conflict with general language
to handle #213

palemieux:
> ok, i'll consider it and do so if it doesn't conflict with general language
to handle #213

#213 is specifically about "Initial values of non-inheritable styles should not be applied to elements to which the style does not apply."

It did not consider specified but non-applicable and non-inheritable styles.

I would very careful with changing the core style resolution algorithm to ignore specified but non-applicable and non-inheritable styles given the inter-dependencies of some style attributes.

I was happy with this PR, and was just asking to extend the special resolution semantics of tts:direction + tts:writingMode to all elements (I do not see how it can hurt).

PR #339, opened 2017-05-26
Clarify that xml:space (and xml:lang) association occurs before ISD c…

### LINKED ISSUE #220: xml:space inheritance rules during ISD construction
XML specifies _this declared intent [of xml:space] is considered to apply to all elements within the content of the element where it is specified, unless overridden with another instance of the xml:space attribute._

TTML specifies _the xml:space attribute may be used with any element in the core vocabulary catalog except parameter vocabulary._

TTML does not mention processing of `xml:space` in the ISD construction process.

Will an `xml:space` value specified on the `head`, `layout` and `region` elements apply (unless overridden) to all descendants of region elements resulting from the ISD construction process (and override `xml:space` values placed on the `body` element)?

skynavga:
…onstruction (#220).

Closes #220.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -5951,6 +5951,9 @@ in the core vocabulary catalog except parameter vocabulary.</p>
 applies to all of that element's descendants unless overridden by a descendant. In other words,
 the language that is associated with an element is either specified on the element or is effectively
 inherited from that element's nearest ancestor element.</p>
+<p>This language association process is based solely on the structure of the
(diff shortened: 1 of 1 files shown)


nigelmegitt:
By the way, what's happening with your pull request initial comments @skynavga ? There have been a few that begin with an ellipsis and don't otherwise make much sense other than including an issue reference.

skynavga:
Github is creating the initial title, which if too long (in its opinion) is truncated and an ellipse with the truncation placed in the comment. I generally just add a single sentence to that: "Closes #N".

PR #345, opened 2017-05-28
Add version information to normative references (#293).

### LINKED ISSUE #293: Reference versioning
I note that a number of references are non-versioned.  That is, no specific version/edition/date is provided in the reference list.  Is this intentional?  If so, what does that mean in W3C?

[FYI, ISO last year started allowing non-versioned normative references, which means basically the latest in force edition.  For normative references this has some downsides, including creating non-interoperable implementations depending on what changed in the underlying specifications versus when in time one implements it. They should only be used with some confidence that updates will have no adverse consequences to implementers.]

Also, why are some references versioned, e.g. 608 and 708 (the letter suffix denotes the version) and some are not versioned, e.g. 12M (no date)?

And, some references are out of date, e.g. "12M" hasn’t been called 12M or approved by ANSI in a really long time. Although there may be cases where the latest version is not desired, in general the latest published version should be cited.

skynavga:
Closes #293.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -23740,7 +23740,7 @@ Mail Extensions (MIME) Part Two: Media Types</titleref>,
 RFC 2046, November 1996, IETF. (See
 <xspecref href="http://www.rfc-editor.org/rfc/rfc2046.txt">http://www.rfc-editor.org/rfc/rfc2046.txt</xspecref>.)
 </bibl>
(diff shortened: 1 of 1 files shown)


nigelmegitt:
Although this is post-merge, I've requested review from @mikedo - Mike, you might as well just look at the ED and if the references have versions correctly, then add a review comment here saying all is good.

PR #346, opened 2017-05-29
Issue 0150 inline background height

### LINKED ISSUE #150: Ways to make span background height be lineHeight
Address any outstanding comments under the following, which are hereby incorporated into this issue:

- [ ] #101 span background height
- [ ] #131 span line height
- [x] #163 clarify bpd

Many (most?) CSS implementations draw the background box behind inline areas with a height less than the line height, meaning that successive lines are drawn with gaps between their background areas, if there's no block level background color set. We want to be able to avoid that situation, since it negatively impacts readability and hence accessibility when background colors are applied to spans.

See also the discussions minuted [here](https://www.w3.org/2016/02/18-tt-minutes.html#item05) and [here](https://www.w3.org/2016/02/25-tt-minutes.html#item05) and [this thread](https://lists.w3.org/Archives/Public/public-tt/2016Feb/0104.html).
## First question: is the height actually defined?

[CSS 2 §10.6.1](https://www.w3.org/TR/CSS2/visudet.html#inline-non-replaced):

> The height of the content area should be based on the font, but this specification does not specify how. A UA may, e.g., use the em-box or the maximum ascender and descender of the font. (The latter would ensure that glyphs with parts above or below the em-box still fall within the content area, but leads to differently sized boxes for different fonts; the former would ensure authors can control background styling relative to the 'line-height', but leads to glyphs painting outside their content area.)
> 
> Note: level 3 of CSS will probably include a property to select which measure of the font is used for the content height.

See also §10.6.6 and §10.6.7 in that specification.

I'm still trying to work out what this means and whether it provides flexibility to implementers to make the inline box height be the same as the line height, or if there's some other calculation that applies.
## Second question: do we need syntax and semantics for clarifying it in TTML2?

Pending the answer to the first question.

skynavga:
Closes #150.


skynavga:
Can you produce a point document that uses more than one value for filling
line gap? It is significantly more complex to make this a style attribute.

nigelmegitt:
@skynavga There are implementations that vary the line gap behaviour now based on a lineHeight heuristic, so that the following two paragraphs render with different backgrounds:

```xml
<style xml:id="line_120" tts:lineHeight="120%"/>
<style xml:id="line_130" tts:lineHeight="130%"/>
<style xml:id="span_style" tts:backgroundColor="black" tts:color="white"/>
...
<p style="line_120">
<span style="span_style">First line<br/>Second line</span>
</p>
<p style="line_130">
<span style="span_style">First line<br/>Second line</span>
</p>
```

A conformant HbbTV 2.0.1 system closes the gaps between background areas in the first paragraph but not the second one. The threshold is set at 125%. Rather than duplicating this heuristic it would be better to allow authorial control at the `p` level.

PR #347, opened 2017-05-30
Issue 0128 conditional styling

### LINKED ISSUE #128: Support for conditional styling.
Consider the use case in which an author wishes to permit the viewer of a TTML2 document to select from one of a number of style choices, either depending on a parameter or a media query, for example choices that vary tts:fontSize and tts:extent to accommodate 'normal size font', 'large size font' and 'small size font' options.

The condition attribute can only be used to omit an element from semantic processing, not to change its behaviour. One might imagine that the following is a way to proceed:

<pre>
&lt;layout&gt;
&lt;region xml:id="r1" condition="parameter(text_size)=='large size font'" tts:extent="95vw 30vh" .../&gt;
&lt;region xml:id="r1" condition="parameter(text_size)=='normal size font'" tts:extent="80vw 20vh" .../&gt;
&lt;region xml:id="r1" condition="parameter(text_size)=='small size font'" tts:extent="60vw 15vh" .../&gt;
&lt;/layout&gt;
&lt;styling&gt;
&lt;style xml:id="sFontSize" condition="parameter(text_size)=='large size font'" tts:fontSize="15vh"/&gt;
&lt;style xml:id="sFontSize" condition="parameter(text_size)=='normal size font'" tts:fontSize="10vh"/&gt;
&lt;style xml:id="sFontSize" condition="parameter(text_size)=='small size font'" tts:fontSize="7.5vh"/&gt;
&lt;style xml:id="sDefaultFont" style="sFontSize" tts:fontFamily="myFontFamily"/&gt;
&lt;/styling&gt;
...
&lt;body&gt;
&lt;div region="r1" style="sDefaultFont"&gt;
...
&lt;/div&gt;
&lt;/body&gt;
</pre>

However this construct, which requires use of xml:id for style and region reference, breaks xml:id uniqueness rules, resulting in invalid documents. What options are there for achieving this use case? I can see:

a) repeating all the content in the document with different style and region references and specifying condition only on the content,

b) basing everything on the initial element and making that conditional (since nothing needs to refer to initial by xml:id), and specifying all regions inline - unfortunately this may be very verbose in terms of repeating regions on many content elements, but it could work for cases where there are only a few regions and they can be associated with body or div elements.

Neither of these two options is particularly attractive - a) is highly repetitious and offers no advantage over the provision of multiple documents with any associated costs for asset management and distribution there. b) is limited in basing style on initial so it is a 'one chance' condition, and it is potentially repetitious in region definition.

By the way, there are at least three audience groups for which this use case exists: 1) Those who have reading difficulties with normal size text; 2) users of different devices, where it has been established that text needs to be rendered at different sizes on large screen televisions from smartphones for example; 3) those who just want to be able to customise the display.

It would be great if the condition construct could be used to allow some predefined viewing options to be authored into the document, i.e. in a controlled way by the document author. I can't see how this can be achieved at present though.

What solution choices are there? Perhaps the easiest is to redefine the condition construct so that it also includes an 'if then else if' syntax in which attributes can be defined, so you might end up with, for example:

<pre>
&lt;region xml:id="r1" condition="if parameter(text_size)=='large size font' then (tts:extent='95vw 30vh' elseif parameter(text_size)='normal size font' then (tts:extent='80vw 20vh') elseif parameter(text_size)='small size font' then (tts:extent='60vw 15vh') else (tts:extent='80vw 20vh')"/&gt;
</pre>

Then xml:id rules are not broken and region r1 can be referenced safely with the attribute evaluation only being conditional. I'd advocate retaining the ability to specify a condition that can be used to exclude the entire element from semantic processing, as now.

Another solution to this problem might be to define some preprocessing using XPath to select specific elements and/or attributes and set values on the basis of the same condition functions that have already been specified, i.e. parameter, media, supports. Something like:

<pre>
&lt;tt [parameters etc]&gt;
&lt;preprocess xmlns="..."&gt;
 &lt;rule condition="parameter(text_size)=='large size text'" path="//region[@xml:id='r1']"&gt;
   &lt;attributes tts:extent="95vw 30vh"&gt;
 &lt;/rule&gt;
&lt;/preprocess&gt;
&lt;head&gt;
  &lt;layout&gt;&lt;region xml:id="r1"/&gt;&lt;/layout&gt;
&lt;/head&gt;
...
&lt;/tt&gt;
</pre>

It would be an error for a path attribute to refer to anywhere except <head> or <body> or their descendants.

This option would also have the incidental effect that it would provide similar functionality to declarative styling. All the rules would be executed in document order prior to processing the <head>. Preprocessing could of course also be performed externally to the document before processing, if a 'user style' is desirable (as is the case for any XML document) .

(raised by Nigel Megitt on 2015-01-16)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/366

skynavga:
Closes #128.


spec/ttml2.xml
```diff
@@ -5886,14 +5886,28 @@ in the core vocabulary catalog:</p>
 </ulist>
 <div3 id="content-attribute-condition">
 <head>condition</head>
-<p>The <att>condition</att> attribute is used to conditionally exclude an element from semantic processing.</p>
+<p>The <att>condition</att> attribute is used to conditionally exclude an element from semantic processing,
+in which case the element is said to be <emph>conditionalized</emph>.</p>
```
nigelmegitt: This seems to say that "conditionalized" is equivalent to "has a condition that evaluates to false" but that is not the intuitive meaning nor is it how it is used later in this pull request. Suggest an edit to:

> The condition attribute is used either to include or exclude an element from semantic processing depending on the result of evaluating the associated logical expression. Elements with a condition attribute are said to be _conditionalized_.

skynavga:
ok, i'll fix the editorial comment

PR #348, opened 2017-05-30
Issue 0215 extent auto

### LINKED ISSUE #215: tts:extent="auto" vs tts:extent="100% 100%"
Is there any difference between specifying tts:extent="auto" and tts:extent="100% 100%" on a region other than the root container region?

If not, suggest rephrasing:

`If applied to a region other than the root container region and the value of this attribute is auto, then the computed value of the style property must be considered to be the same as the extent of the root container region.`

to

`If applied to a region other than the root container region and the value of this attribute is auto, then the specified value of the style property must be considered to be "100% 100%".`

skynavga:
Closes #215.


skynavga:
That's a different (new) issue, unrelated to this one.

nigelmegitt:
OK I've raised #351 on the subsidiary point.

PR #350, opened 2017-05-31
Issue 0168 implied region targets current region

### LINKED ISSUE #168: Make inline region semantics set values on current region
The current semantic when inline style attributes applicable to regions are used in content elements is to create a new implied inline region. 
### Options
1. Keep existing semantic where new anonymous regions are created.
2. Modify the semantic such that when region style attributes are encountered on content elements they are used to modify the current region for that content element while the element is active.

Option 2 would be syntactic shorthand for a `<set>` element on a region that modifies the region's `tts:origin`, `tts:extent` or `tts:position`.
### History

I raised this discussion with the thread beginning at https://lists.w3.org/Archives/Public/public-tt/2015Oct/0068.html however it was not concluded.

This issue satisfies tracker [action 472](https://www.w3.org/AudioVideo/TT/tracker/actions/472).

skynavga:
Closes #168.


spec/ttml2.xml
```diff
@@ -16272,9 +16272,11 @@ the following style attributes apply exceptionally to the <el>region</el> <emph>
 </ulist>
 <p>In such cases, the application of these style attributes are considered to be
 equivalent to an anonymous <loc href="#terms-inline-animation">inline animation</loc> taking
-the form of a <loc href="#animation-vocabulary-set"><el>set</el></loc> element, where the
+the form of a <loc href="#animation-vocabulary-set"><el>set</el></loc> animation element <emph>A</emph>, where the
 animation applies to <emph>R</emph> over an active time interval equivalent to the active time
-interval of <emph>E</emph>.</p>
+interval of <emph>E</emph>, and where, if multiple animation elements
+(<emph>A<sub>i</sub></emph>, ..., <emph>A<sub>j</sub></emph>) already apply to <emph>R</emph>,
+then <emph>A</emph> applies subsequent to these the last of these existing animations, namely, immediately after <emph>A<sub>j</sub></emph>.</p>
```
nigelmegitt: Sorry, I don't understand this sentence - maybe a grammar error? 

"subsequent to these the last of these" -> "subsequent to the last of these" ?

If I understand this right this means that if there are two temporally overlapping anonymous `set`s, call them s_1 and s_2, then s_2 only becomes active after s_1 has ended. That's the opposite from my expectation, but I'm happy to have _any_ rule here rather than none! It isn't clear what the behaviour is if multiple animation elements can not be distinguished by timing alone, e.g. if they have the same `begin` and `end` attribute values.

PR #352, opened 2017-05-31
Add border examples

### LINKED ISSUE #189: Add support for adjacent background area merging
Follow-on from #176 which was closed since the core issue, corner rounding, was resolved. This is the new home for the additional requirement to merge background areas such that the corner rounding results in a nice looking output.

Extracted comments from: https://github.com/w3c/ttml2/issues/176#issuecomment-246140753 and https://github.com/w3c/ttml2/issues/176#issuecomment-248003223

None of the examples of rounded borders I have seen use different radii for the two directions, so I'd advocate simplifying to a single value for now, until a need for two radii is presented.

However one thing that may be underspecified at the moment is the flow of rounded rectangles between adjacent areas. I've seen this kind of thing:
image

where I would rather have seen something like this:
image

or even perhaps:

image

In other words when creating a composite background shape from congruent area backgrounds the radial centre should be offset in the direction of the hypotenuse of the right angle triangle formed between the vertex being drawn and the two adjacent vertices, not away from it. There are probably more elegant ways to describe this!

Further examples from f2f meeting 2016-09-19:

image

and 

image

These require a little more thought in terms of how to specify them

nigelmegitt:
Fixes #189.

One image for the existing example, the other with an XML example for
rounded borders as per #189.

Would appreciate a review especially of the border example with the red, green and blue borders - I'm not at all sure that the rendering is correct - two queries:
1. Should the span border be included in the size of the inline area, and therefore increase the height of the line area, and bump the text up a bit? (related to this: should the border of the text flow into the border of the region?)
2. What is the z order of the borders? I have made the region border in front of the p border which in turn is in front of the span border; possibly it should be the reverse?


nigelmegitt:
@skynavga Sure. I was just adding an image to reflect the existing example fragment, I didn't modify that fragment.

Issue #302, opened 2017-04-28
Incorporate resolutions of TTML1 issues.


skynavga:
This is an umbrella issue to cover the tasks of incorporating resolutions of TTML1 issues into TTML2.

- [x] [w3c/ttml1/#205](https://github.com/w3c/ttml1/issues/205) padding percentage
- [x] [w3c/ttml1/#221](https://github.com/w3c/ttml1/issues/221) padding and extent
- [x] [w3c/ttml1/#235](https://github.com/w3c/ttml1/issues/235) tab semantics
- [x] [w3c/ttml1/#237](https://github.com/w3c/ttml1/issues/237) references to CSS2
- [x] [w3c/ttml1/#239](https://github.com/w3c/ttml1/issues/239) tts:overflow

skynavga:
Completed sub-tasks.

PR #355, opened 2017-06-01
Clarify nature of region overflow in terms of region's content, paddi…

skynavga:
…ng, and border rectangles (#302).

See #302.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -10970,6 +10970,13 @@ as a style property only to those element types indicated in the following table
 not be clipped outside of the affected region.
 If the value is <code>hidden</code>, then content should
 be clipped outside of the affected region.</p>
+<note role="clarification">
+<p>Marks produced by content associated with a region's content rectangle may be rendered outside that
+content rectangle, thus intersecting with the region's padding and border rectangles, without any affect from
+the <att>tts:overflow</att> attribute. Such marks only become a candidate for region overflow when they would
+extend outside the region's extent which is coincident with the region's border rectangle, which contains the
+region's padding rectangle, which contains the region's content rectangle.</p>
+</note>
 <note role="explanation">
 <p>Unless a manual line break element <el>br</el> is used by the content author,
 a paragraph of a given region will generate no more than one line area in that


spec/ttml2.xml
```diff
@@ -10970,6 +10970,13 @@ as a style property only to those element types indicated in the following table
 not be clipped outside of the affected region.
 If the value is <code>hidden</code>, then content should
 be clipped outside of the affected region.</p>
+<note role="clarification">
+<p>Marks produced by content associated with a region's content rectangle may be rendered outside that
+content rectangle, thus intersecting with the region's padding and border rectangles, without any affect from
+the <att>tts:overflow</att> attribute. Such marks only become a candidate for region overflow when they would
+extend outside the region's extent which is coincident with the region's border rectangle, which contains the
+region's padding rectangle, which contains the region's content rectangle.</p>
+</note>
```
nigelmegitt: One potential point of confusion here - whilst this may be true, would we not normally expect line wrapping to occur within the region's content rectangle? The asymmetry seems strange - why do we not have the same applicable rectangle for the line wrapping boundary as for the overflow clipping boundary?

skynavga: Line wrapping does occur within the region's content rectangle. Did you think otherwise?

nigelmegitt: No, I agree, I would expect line wrapping to occur within the region's content rectangle. But I would also expect the overflow clipping boundary to be the same as the line wrapping boundary. This is not the case in the current wording, which is the point I'm making.

skynavga: I would not expect that. An area is bounded by its border rectangle, so, if clipping occurs on the area, I expect it to clip outside the border rectangle. In general, neither XSL-FO nor CSS precludes the appearance of marks (or some portion of a mark) outside of the content rectangle with which it is associated. TTML follows the same precept. Since we define overflow only on region, but not on content elements, then we are only concerned with clipping on a region area.

nigelmegitt: Hmm. It still seems inconsistent to me but I'll back off on this one.

PR #356, opened 2017-06-01
Revert deprecation of profile parameter attribute (#331).

### LINKED ISSUE #331: Revert deprecation of ttp:profile on root element.
Upon further consideration, I believe my earlier suggestions to deprecate ``ttp:profile`` (in favor of ``ttp:processorProfiles``), which suggestion as already been implemented, may have been premature. I recommend the group consider restoring the use of ``ttp:profile`` which can serve as a shorthand for ``ttp:processorProfiles`` in the common case that only one processor profile applies.

skynavga:
Closes #331.


spec/ttml2.xml
```diff
@@ -4265,13 +4265,16 @@ when specified on the <el>tt</el> element.</p>
 </div3>
 <div3 id="profile-attribute-profile">
 <head>ttp:profile</head>
-<p>The <att>ttp:profile</att> attribute is <phrase role="deprecated">deprecated</phrase>. If creating a TTML document instance for TTML2 (or later versions),
```
nigelmegitt: Even if not deprecating `ttp:profile` we still need the precedence rule for TTMLv>=2 in case both `ttp:profile` AND `ttp:processorProfiles` are present.

mikedo: Hmm.  Consider that we not permit both.  The purpose of retaining ttp:profile is only for backwards compatibility with TTML1 instance documents, not authoring of new TTML2-conformant-only instance documents. For the latter, there is no real reason to allow both.

skynavga:
[GA - from 06/01/17 meeting] Add note to conformance section that indicates intention that TTML1 documents that are identified as such remain conformant in TTML2.

spec/ttml2.xml
skynavga: @nigelmegitt I confirmed that there is already language that establishes precedence, about which see [1]

[1] file:///Users/glenn/work/ttml2/spec/ttml2.html#semantics-procedure-construct-effective-processor-profile

spec/ttml2.xml
```diff
@@ -2026,6 +2026,24 @@ a <el>ttp:profile</el> element as a child of the
 <att>ttp:version</att> attribute on its root <el>tt</el> element if it requires
 support for a <loc href="#terms-feature">feature</loc> not defined by <bibref ref="ttml1"/>.</p>
 </div2> <!-- claims -->
+<div2 id="compatibility">
```
skynavga: @mikedo I added this new non-normative text on compatibility in the context of conformance; let me know if you have any concerns

mikedo: @skynavga The one case I was most concerned about (A conforming TTML2 processor should be able to process a TTML1 document instance; ...) contains a conformance term, "should", and (using its English definition) implies a softer condition that the other 3 "is intended to" statements. I think all 4 cases should have the same sentence construction.

skynavga: See the qualification under "Document Conventions":

>Within normative prose in this specification, the words may, should, and must are defined as follows:

So they are not conformance keywords in a non-normative context.

mikedo: I think that's a confusing practice, but OK.  That's not really my concern.  Why should the sentence construction be different on this case and not the other 3?

skynavga: The "is intended to" makes a statement about the specification, while the "should be able to" makes a statement about an implementation. We really can't say what the creator's of the implementation intend.

mikedo: OK, I see that.

spec/ttml2.xml
nigelmegitt: @skynavga OK. It would be useful to modify the new informative note introduced at line 4296 pointing to that section to indicate where the profile resolution happens and that `ttp:processorProfiles`, if present, always takes precedent over `ttp:profile`. Suggested wording in comment below.

spec/ttml2.xml
```diff
@@ -4265,13 +4288,16 @@ when specified on the <el>tt</el> element.</p>
...
-<p>If used in a document, the <att>ttp:profile</att> attribute
+<p>The <att>ttp:profile</att> attribute
 denotes a <loc href="#terms-processor-profile">processor profile</loc>
 of the Timed Text Markup Language (TTML) that applies when processing
 a <loc href="#terms-document-instance">document instance</loc>.</p>
+<note role="clarification">
+  <p>If an author wishes to specify that one or more of a collection of processor profiles apply
+  rather than a single processor profile, then the content author is advised to use the
+  <att>ttp:processorProfiles</att> as specified by <specref ref="profile-attribute-processorProfiles"/>,
+  which attribute is introduced by this version of the TTML specification.</p>
```
nigelmegitt: e.g. append:

[[
If both `ttp:profile` and `ttp:processorProfiles` are present in the same document instance then the semantics defined in 5.2.4.3 [construct effective processor profile] result in `ttp:processorProfile` taking precedence over `ttp:profile`, with the latter effectively being ignored by a conforming TTML2 processor.
]]

spec/ttml2.xml
nigelmegitt: I agree that using conformance language in informative sections is confusing and probably poor practice, even if there's text to declare that it does not introduce conformance rules. It is normal practice in other standards bodies to avoid doing this always - certainly we don't do it in the EBU-TT specifications and it would break ETSI rules too I believe - maybe ISO too? Instead similar words are used that are not defined as conformance terms, such as "could" instead of "may" for example.

skynavga: Our rules are our rules. The W3C does not have a standard in this regard, so we are fine.

spec/ttml2.xml
```diff
@@ -2026,6 +2026,29 @@ a <el>ttp:profile</el> element as a child of the
 <att>ttp:version</att> attribute on its root <el>tt</el> element if it requires
 support for a <loc href="#terms-feature">feature</loc> not defined by <bibref ref="ttml1"/>.</p>
 </div2> <!-- claims -->
+<div2 id="compatibility">
+  <head>Compatibility</head>
+<p><emph>This sub-section is non-normative.</emph></p>
+  <div3 id="compatibility-forward">
+    <head>Forward Compatibility</head>
+    <p>A conforming TTML1 processor may be able to process a TTML2 document instance, provided it ignores unknown vocabulary
```
skynavga: ok, i'll qualify as such

spec/ttml2.xml
```diff
@@ -2026,6 +2026,29 @@ a <el>ttp:profile</el> element as a child of the
...
+    and ignores known vocabulary that appears in a context where it is not permitted in a conforming TTML1 document instance;
+    however, such processing may discard or fail to present content, or may present content differently than intended when processed
+    by a conforming TTML2 processor.</p>
```
skynavga: Implementations could do many things to facilitate forward and/or backward compatibility, but it is going too far to suggest that TTML1 implementations be retroactively modified to use TTML2 features. Of course they can, a but that is up to implementors to decide, and not for the specification to suggest.

spec/ttml2.xml
```diff
@@ -2026,6 +2026,29 @@ a <el>ttp:profile</el> element as a child of the
...
+      if a TTML1 processor encounters a <el>style</el> element as a child of another <el>style</el> element, then the former
+      nested style may be ignored (since this facility was not defined for TTML1).</p>
+    </note>
+  </div3>
+  <div3 id="compatibility-backward">
+    <head>Backward Compatibility</head>
+    <p>A conforming TTML1 document instance is intended to be a conforming TTML2 document instance.</p>
+    <p>A conforming TTML2 processor should be able to process a TTML1 document instance; however, it may emit warnings if it
+    encounters deprecated features.</p>
+  </div3>
```
skynavga: Add a note under presentation processor conformance noting that two conforming presentation processors do not have to produce same presentation given a single conforming document as input.

spec/ttml2.xml
skynavga: Add note at end of 6.2.7.

spec/ttml2.xml
skynavga: DRY applies. Note at end of 6.2.7 will suffice.

nigelmegitt: This was already in an informative note. Anyhow the new note at the end of 6.2.7 is okay.

PR #390, opened 2017-06-10
Add justify keyword value to tts:displayAlign (#370).

### LINKED ISSUE #137: The tts:displayAlign property may need to apply to content elements
especially if their children specify explicit bpd

(raised by Glenn Adams on 2015-04-10)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/384

### LINKED ISSUE #370: Add justify to tts:displayAlign.
Add ``justify`` value to support justification in block progression dimension.

skynavga:
Add justify keyword value to tts:displayAlign (#370).
Apply tts:displayAlign to block area generating content elements (body, div, p) (#137).
Add and update feature definitions as needed for #displayAlign* features.

Closes #137.
Closes #370.


spec/ttml2.xml
```diff
@@ -9198,7 +9194,8 @@ as a style property only to those element types indicated in the following table
 <td>
 <code>before</code> |
 <code>center</code> |
-<code>after</code>
+<code>after</code> |
+<code>justify</code>
```
nigelmegitt: The reference for displayAlign is XSL 1.1 §7.14.4 "display-align". That defines the values:

> auto | before | center | after | inherit

Hence with this pull request as it is we have no defined semantic for the new `justify` keyword. It may be "obvious" but it still needs to be defined somewhere.

skynavga: Adding normative prose defining justify.

PR #391, opened 2017-06-11
Make values of tts:rubyPosition consistent with <emphasis-position> (…

### LINKED ISSUE #374: Add auto and outside values to tts:rubyPosition.
Add ``auto`` and ``outside`` values as defined by ``<emphasis-position>`` expressions.

skynavga:
…#374).

Closes #374.


spec/ttml2.xml
```diff
@@ -14229,6 +14225,101 @@ the computed opacity and the supported opacity is minimized. If there are multip
 the computed value, then the value most distant from 0, i.e., the greatest opacity, is used.</p>
 </note>
 </div3>
+<div3 id="style-value-annotation-color">
```
nigelmegitt: Why is this needed? Since it is applied to a `span` anyway, wouldn't `tts:color` apply and serve the same purpose, when specified on a span containing annotation text?

skynavga: Both ruby and text emphasis need a way to say that their color follows the color of the base text, which is done using the ``current`` value defined on ``<annotation-color>``.

nigelmegitt: OK thanks.

PR #468, opened 2017-10-14
Add document processing context override semantics for effective con…

### LINKED ISSUE #467: Allow detection of the EBU-TT-D content profile.
The _[construct effective content profile]_ does not allow the _effective content profile_ to be set by inspection of the document, beyond the values of `ttp:contentProfiles` and `ttp:profile`.

This prevents TTML2 profiles such as IMSC1.1 to determine that a document conforms to EBU-TT-D (and thus IMSC1.1) since EBU-TT-D signals content profiles using the `ebuttm:conformsToStandard` element.

Propose to add step at _[construct effective content profile]_ allowing the _effective content profile_ to be determined by the document processing context.

skynavga:
…tent and processor profile construction (#467).

Closes #467.


spec/ttml2.xml
```diff
@@ -1358,6 +1358,22 @@ A <loc href="#terms-default-region">default out-of-line region</loc> is implied
 <loc href="#terms-out-of-line-region">out-of-line region</loc> is specified explicitly.</p>
 </def>
 </gitem>
+<gitem id="terms-override-content-profile">
+<label>[override content profile]</label>
+<def>
+<p>A <loc href="#terms-content-profile">content profile</loc> that is determined to override all other
```
palemieux: `ebuttm:conformsToStandard` allows multiple content profiles to be specified, like `ttp:contentProfiles` with the semantic `all`

skynavga:
In that case, it will be up to the document processor context to decide how
to map those profiles to a single TTML2 compliant content profile. No
change in language is required.

spec/ttml2.xml
nigelmegitt: That's what I thought too, but I'm not really sure why the concept of an "override" is needed for a content profile. Maybe if I understood that then it would help...

spec/ttml2.xml
```diff
@@ -1358,6 +1358,22 @@ A <loc href="#terms-default-region">default out-of-line region</loc> is implied
...
+possible determinations of a <loc href="#terms-content-profile">content profile</loc>.</p>
+<p></p>
+</def>
+</gitem>
+<gitem id="terms-override-procesdsor-profile">
```
nigelmegitt: typo: procesdsor -> processor

spec/ttml2.xml
```diff
@@ -2725,7 +2741,16 @@ in order to maintain content profile invariants.</p>
...
 <olist>
 <item>
-<p>if a <att>ttp:contentProfiles</att> attribute is specified on the root <el>tt</el> element, then</p>
+<p>if the <loc href="#terms-document-processing-context">document processing context</loc> determines that a specific
+<loc href="#terms-override-content-profile">override content profile</loc> <emph>P<sub>O</sub></emph> is to override all other determinations, then the
+<loc href="#terms-effective-content-profile">effective content profile</loc> is <emph>P<sub>O</sub></emph>;</p>
+<note role="example"><p>For example, an <bibref ref="ebuttd"/> document that complies with this specification
+may specify an <code>ebuttm:conformsToStandard</code> attribute to signal compliance with the <bibref ref="ebuttd"/> content profile.
+In such a case, a <loc href="#terms-content-processor">content processor</loc> that supports that profile may use the presence of
+that attribute as an override to determine the <loc href="#terms-effective-content-profile">effective content profile</loc>.</p></note>
```
nigelmegitt: This example doesn't work since the cardinality of `ebuttm:conformsToStandard` is `0..*`, so multiple values with different content profiles can be signalled, when they all apply.

skynavga: Actually, it does work. An example does not have to cover all scenarios. In the example listed here, it refers to one profile (the EBU-TT-D profile). It is up to a processor that uses this mechanism (override content profile) to determine how to obtain the override profile, e.g., based on the presence of one profile referenced by ``ebuttm:conformsToStandard``, based on the presence of multiple profiles refenced by ``ebuttm:conformsToStandard``, based on whatever criteria the processor deems relevant, etc.

spec/ttml2.xml
skynavga: Thanks, will fix.

skynavga:
For some reason, I can't reply to your comment about why override is needed above (there is no Reply... field shown), so I will reply here. The original ask was

>add step at [construct effective content profile] allowing the effective content profile to be determined by the document processing context

from which I discern two scenarios, such that, in the absence of such a new step, we have either (1) the effective content profile is null or (2) the effective content profile is different from what the processor (or the entity operating the processor) wants. The first of these scenarios we might call the trivial (or null) case, while the second is the true override case.

To address these scenarios, I originally crafted two new steps, a new step 1 that supports the override scenario, and a new step just before the last step that supports the trivial (null) case. After doing that, I realized that the semantics of both steps could be combined into a single override case, where, in the first scenarios, it is overriding null, and in the second, it is overriding non-null.

I suppose we can ask whether we want to support the second scenario, i.e., whether to allow the processor to override a specified content profile or not. One case I can imagine is where we want to allow a processor to use a modified or extended form of the algorithm we have specified to construct an effective content profile. For example, in the current algorithm, the profiles referenced by ``ttp:contentProfiles`` has a higher priority than any ``<ttp:profile type="content"/>`` profiles, such that, if the former is specified, the latter is ignored. I can imagine a processor that might want to give the latter priority, or combine multiple content profiles from both the attribute and the top-level profiles. Or perhaps the entity controlling the processor simply wants to treat the document differently than the author specified (in terms of a content profile).

spec/ttml2.xml
nigelmegitt: OK, I see. However it actually still doesn't work for another reason I just spotted: `ebuttd:conformsToStandard` is an element not an attribute.

spec/ttml2.xml
nigelmegitt: I've explained this to myself thus: a document interchange context may require that a specific profile A is used when interchanging documents, and the participant in that context who is receiving documents chooses to validate them. It makes no sense to use any content profile other than A for validating. If a document is provided that claims conformance to B instead, the validator should continue to validate against A regardless.

PR #470, opened 2017-10-18
Issue 0406 style semantics

### LINKED ISSUE #406: Add equivalent CSS properties to style attributes
CSS WG and TAG review feedback: semantics defined only in terms of XSL-FO should be replaced with CSS semantics where possible; if similar but different semantics are available in CSS properties mention them at least informatively.

Use the most recent CSS Snapshot as the source of usable CSS property definitions. This is currently https://www.w3.org/TR/css-2017/

nigelmegitt:
Fixes #406 and most of the sub-issues by adding CSS derivations. Raising as a PR for review now before all the sub-issues are resolved since so many of them _are_ done, and now would be a good time to get any feedback on, for example, the formatting of XSL and CSS derivations.


cconcolato:
1) The following sentence reads strangely to me:
>The section lists derivations or equivalents, references and any differences compared to the derivation source. 

What about:
> The section lists references from which a style attribute was derived and any difference compared to that reference.

2) Regarding the following:
> The CSS equivalence assumes that each  Intermediate Synchronic Document  is mapped to an equivalent  [HTML5]  DOM structure whose elements are given the stated style properties. 

I think this sentence could be improved. "equivalent" is unclear. I think it could say:
> Regarding the CSS equivalence, when the TTML properties used in a TTML document simply map to similar CSS properties (possibly changing the name of the property or the name of the value), then the structure of the TTML elements in the TTML document can be used to create the structure of HTML elements. In some more complex cases, a simple mapping does not exist (e.g. when the property is sensitive to the client environment (e.g. font-size), other strategies for creating an HTML document, possibly using JavaScript, should be used.

3) I would prefer having a section related to CSS only and another one related to XSL only. Readers will not probably be interested in both at the same time.

4) Some conventions in the tables are unclear:
- what is the difference between Note and Details? I would propose to merge them into "Difference wrt model".
- what does it mean when they are empty? I would propose to add "none"
- what does "-inherit" mean in the details line?

5) I think also that having a single table with 3 columns (TTML property / CSS (XSL) Reference / Difference) and one row per TTML property might be easier to read. It currently does not look like the text content would be too big to fit in the 3-column table.

nigelmegitt:
Thanks for the review @cconcolato.

1. derivations etc

>> What about:
>
>    The section lists references from which a style attribute was derived and any difference compared to that reference.

The trouble is that strictly the style attributes were not all derived from multiple sources. How about:

> The section lists references from which a style attribute was derived, or functionally equivalent references, and any differences compared to those references.

2. HTML equivalent elements

I like the idea there, and was conscious when I wrote the pull request that the wording was vague. I just wasn't 100% certain that the approach you suggest actually works! If you can validate that, then I'd be happy to go with your wording, or maybe a slightly edited version for conciseness.

3. XSL section vs CSS section

The approach you suggest would work, but would need quite a big re-edit. I would have to change the derivation links in every style attribute to be "XSL derivation" and "CSS derivation" to point to the relevant subsection. I agree it might be an improvement but since it will be a lot of work I'd like other views before I start doing that. Especially your idea under 5. might work.

4. Table conventions

Note and Details were in the previous Derivation table, so I kept the information split similarly. I agree they could probably be merged.

Similarly "-`inherit`" came from there too - it means that the keyword `inherit` is removed.

That gives me an idea: change "Details" to "Value differences" and put in that row: 
* removed keywords,
* added keywords,
* value mappings

I would keep notes if there are any. I'd be happy just to remove any empty Notes rows rather than adding "None". I don't strongly mind either way.

5. One table, an extra column.

I agree this could be a good approach - I didn't do it because I wasn't sure if it would be clear, but it would certainly be more concise and as you say, it would probably fit fine. I'd like views on this as well as 3. above before making the changes.

nigelmegitt:
@cconcolato I've pushed some changes - I went with your option 5 because it was more concise and looked fine. I think it's the better option.

I also pushed a change to reference `skewX()` and `skewY()` - @cconcolato and @skynavga might want to take a look at that.

All of the style attributes with no derivation now have an ednote on them.

nigelmegitt:
Built version can be viewed at https://rawgit.com/w3c/ttml2/issue-0406-style-semantics-build/index.html

cconcolato:
There are still some issues I think with the text:
- what I proposed in point 2 is missing
- the "-inherit" notation is still unclear
- some properties have 2 column, some just one. That's not entirely clear.
- I would merge Reference and Model; and also merge Values and Notes.

Anyway, this is very low priority.

PR #486, opened 2017-11-07
Add clipBegin and clipEnd parameter attributes

### LINKED ISSUE #483: Add parameters to define the temporal extent
Split out from #323 - add parameters analogous to the `clipBegin` and `clipEnd` attributes of the `<audio>` element but applicable to the TTML document, whose effect is to define the period of time in the document's timebase for which the document defines some behaviour.

Initial value of the `clipBegin` parameter would be 0s, which is the current behaviour.
There would be no initial value of the `clipEnd` parameter: if present it would specify a clipping behaviour, or if absent the document contents would define the end time.

The behaviour is as follows:
* The computed value of `clipBegin` is the specified value if present or `0s` if absent.
* When generating ISDs any ISDs ending before the computed value of `clipBegin` shall be pruned.
* Any remaining ISDs beginning before  the computed value of `clipBegin` shall have their `/isd:isd/@begin` attribute set to `clipBegin`.
* If `clipEnd` is specified, any remaining ISDs beginning after `clipEnd` shall be pruned.
* If `clipEnd` is specified, any remaining ISDs ending after `clipEnd` shall have their `/isd:isd/@end` attribute set to `clipEnd`.
  * For the purposes of comparison any ISD with an unspecified `end` time or an end equivalent to "indefinite" is considered to end after any specified `clipEnd` time.

Incidentally, it is possible to construct a TTML document with an ISD that has an indefinite end time but it is not possible to express this in an `isd:isd` document, because the `end` attribute on the root element is required and `<timeExpression>` does not permit a value equivalent to the SMIL keyword "indefinite" - I will raise that as a separate issue.

The rationale for _not_ using `begin` and `end` attributes for this on the `tt:tt` element is that the `begin` attribute would create an offset for the descendant elements' timing, and the `tt:tt` element is not formally considered to be a time container.

nigelmegitt:
Closes #483.

Semantics defined but ISD generation algorithm not updated yet
because there’s an issue note there saying that it needs to be revised
for other reasons. I’ve added a similar note for #483 to include the
ISD pruning/truncation/extension semantics when that is done.

Also added to schemas and regenerated ED.


palemieux:
What problem is this intended to solve? It is not clear from #483

spec/ttml2.xml
```diff
@@ -17537,6 +17619,15 @@ and the resulting generated out-of-line region is shown by the following example
...
 </issue>
+<p/>
+<issue id="issue-ttml2-483">
+<head>Truncate/prune/extend to deal with
+ttp:clipBegin and ttp:clipEnd</head>
+<source><loc href="https://github.com/w3c/ttml2/issues/483">https://github.com/w3c/ttml2/issues/483</loc></source>
+<p>When revising, also include truncation/pruning/extension to deal with the
+<loc href="#parameter-attribute-clipBegin">ttp:clipBegin</loc> and 
+<loc href="#parameter-attribute-clipEnd">ttp:clipEnd</loc> semantics.</p>
+</issue>
```
palemieux: `clipBegin` and `clipEnd` should be authorial hints only and not affect processing. More specifically, it is not clear why they need to affect processing since the application already controls presentation of the document, including starting and stopping presentation/processing at timeline locations outside of ISD boundaries.

nigelmegitt: I disagree. This needs to be normative and to affect processing. It has no effect when the attributes are absent, so it has no backward compatibility issue.

palemieux: You have simply not motivated at all whatsoever why the processing model needs to be modified. The TTWG really needs to stop modifying TTML2 unless necessary.

nigelmegitt: Yes, I have, at https://github.com/w3c/ttml2/issues/483#issuecomment-345283633 perhaps you did not spot that?

nigelmegitt: @palemieux ping?

palemieux: @nigelmegitt Thanks for the reminder. See additional thoughts in the ticket. For the record, you have not convinced me that creating additional ISDs is necessary.

nigelmegitt:
> Clip begin and end sound a lot like query parameters

@skynavga They're nothing to do with query parameters. See the issue for the use case and motivation - in summary it's about the ability for a document author to define the temporal period during which the document defines behaviour. 

If the parameter attribute names seem misleading, could you please propose alternative names?

nigelmegitt:
In order to make progress on TTML2, in the absence of a clear consensus, I'm closing this pull request, which can be reopened or moved to another repo if we want to pick it up again in the future.

PR #488, opened 2017-11-09
Coalesce references to URI syntax into a new <uri> value expression (…

### LINKED ISSUE #474: Delimiters in URI references possibly ambiguous
The delimiter used to separate profile designators in `ttp:contentProfiles` and `ttp:processorProfiles` is `<lwsp>` which includes space characters, however according to the reference https://www.w3.org/TR/xmlschema-2/#anyURI the `xsd:anyURI` type can contain non-percent-encoded spaces in their lexical representation, albeit they are discouraged.

Therefore the delimiter as defined at present may result in false separation of URIs where it is not intended.

I would resolve this either by changing the delimiter, or preferably, by requiring percent encoding of the URIs which results in them no longer containing space characters.

skynavga:
…#474).

Closes #474.


nigelmegitt:
@skynavga please could you merge from `master` so this is up to date with the new build process?

nigelmegitt:
Also, the pull request shouldn't include `spec/ttml2.html` since that is now a transient build artefact.

skynavga:
@nigelmegitt regarding the request to change the normative reference for [URI] from RFC2396 to RFC3986, I am not comfortable making this change in the context of this issue, as it may have wider ramifications; can you accept this PR as is and file another issue if you want to upgrade the [URI] reference? The references in this PR are synchronized with 2396 section numbering, and not 3986.

PR #492, opened 2017-11-15
Added note specifying the mapping of tt:image to smpte:backgroundImage

### LINKED ISSUE #460: SMPTE backgroundImage annotation.
A request from SMPTE made on a call today:

In the context of SMPTE-TT being specified as a "safe harbor" format for captions by the FCC, and TTML2 adopting an equivalent semantic of `smpte:backgroundImage`, add an annotation to the effect that the motivation for including image support in TTML2 is at least in part derived from the use of images in SMPTE-TT, and describe the equivalence to `smpte:backgroundImage`.

Note that it may be that the mapping is not from `tts:backgroundImage` but from `<image>` instead. This needs to be checked.

palemieux:
Closes #460


spec/ttml2.xml
```diff
@@ -7241,6 +7241,13 @@ constraints apply:</p>
 </tbody>
 </table>
 <p></p>
+<note role="clarification">
+<p>The <code>image</code> element incorporates and expands the capabilities of the 
+<code>#image</code> extension introduced in <bibref ref="smpte2052-1"/>. In particular, 
+a <code>smpte:backgroundImage</code> attribute that references <loc href=
```
skynavga: needs an "a" after "references"

skynavga:
There seems to be a problem with the ttml2.html file generated here, as it appears to delete/undo prior changes. Please remove both the ttml2.html and schema.zip artifacts from this PR before merging.

palemieux:
@skynavga Can you be more specific? This branch was split from the `HEAD` of `gh-pages` and `ttml2.html` was generated using `ant sg`

skynavga:
Click on show diffs on ttml2.html under files changed, and look at the
changes that are unrelated to your changes to ttml2.xml.

spec/ttml2.xml
```diff
@@ -25137,6 +25144,11 @@ Signal &ndash; NTSC for Studio Applications</titleref>, SMPTE Standard.
 href="http://www.smpte.org/">Conversion from CTA-708 Caption Data to SMPTE-TT</titleref>,
 SMPTE Recommended Practice.
 </bibl>
+<bibl id="smpte2052-1" key="SMPTE ST 2052-1">SMPTE ST 2052-1,
```
skynavga: It would be better to make the value of key "SMPTE 2052-1" for consistency with the existing bibliographic entry with key "SMPTE 2052-11".

palemieux:
@skynavga `ant build-escapes` needed to be run. Why isn't this part of `ant sg`?

spec/ttml2.xml
palemieux: SMPTE 2052-1 is an incorrect reference to a SMPTE standard, which should be referenced as SMPTE ST 2052-1

skynavga: the form of keys we use in our document are not constrained by other usage; however, if you want to make use of SMPTE ST 2052-1 as our key, then perhaps the other key should be changed to SMPTE RP 2052-11? the problem is that they are now inconsistent

skynavga:
You need to use `ant generate` if you want to build escapes as well. The `ant sg` is a shortcut generation the intentionally does not build them. Normally, they don't need to be built unless one of the escaped entities has been changed.

palemieux:
> You need to use ant generate if you want to build escapes as well. 

Ok.

spec/ttml2.xml
skynavga: thanks, that works

palemieux:
@skynavga Please approve the PR when all your comments have been addressed.

mikedo:
The SMPTE title is incorrect for 2052-1:2013. It should be "Timed Text Format (SMPTE-TT)". 2052-11 that defines the conversion of 708, but it's off topic in this context.

mikedo:
In TTML2, 2052-1 is informative as far as I can tell.  I think undated informative references are fine.  (In contrast, in IMSC1.1 is it normative.)

PR #505, opened 2017-12-23
Update [URI] to reference RFC3986 (#504).

### LINKED ISSUE #504: Change URI reference to RFC3986
[RFC3986](https://www.ietf.org/rfc/rfc3986.txt) superseded [RFC2396](https://www.ietf.org/rfc/rfc2396.txt) for the definition of URI in January 2005. We should reference it.

Two references to URI are specifically numbered. They may be mapped as follows:

| RFC2396 | RFC3986 | Notes |
|----------|----------|-------|
|4.1 Fragment Identifier (syntax) | 3.5 Fragment | |
|2.4.3 Excluded US-ASCII Characters| 2.4 When to Encode or Decode| Also 2.5 Identifying Data is relevant - the note in [pull #488 §8.3.10](https://rawgit.com/w3c/ttml2/issue-0474-uri-references-build/index.html#content-value-uri) should be updated to refer to any UCS character values not just US-ASCII control characters|

skynavga:
Closes #504.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -6589,7 +6589,9 @@ designator is (semantically) supported by the <loc href="#terms-content-processo
 <p>The value syntax of the <code>xsd:anyURI</code> data type must adhere to the definition specified by <bibref ref="xsd-2"/>, &sect;3.2.17.</p>
 <note role="clarification">
   <p>For avoidance of doubt, characters designated to be excluded in a URI by <bibref ref="uri"/>, &sect;2,
-  <emph>Characters</emph>, and, more specifically, SPACE (U+0020) characters,
+  <emph>Characters</emph>, and, more specifically,
+  the XML whitespace characters &amp;#9; (HT), &amp;#10; (LF), &amp;#13; (CR), and &amp;#20; (SPACE),
+  as specified in <bibref ref="xml10"/>, &sect; 3.3.3, step (3), first bullet,
   do not appear in an unescaped form in the lexical representation of a <code>&lt;uri&gt;</code> value.</p>
 </note>
 </div3>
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -6588,10 +6587,12 @@ designator is (semantically) supported by the <loc href="#terms-content-processo
 </tbody>
 </table>
 <p>The value syntax of the <code>xsd:anyURI</code> data type must adhere to the definition specified by <bibref ref="xsd-2"/>, &sect;3.2.17.</p>
-<note role="clarification"><p>For avoidance of doubt, characters designated to be excluded in a URI by <bibref ref="uri"/>, &sect;2.4.3,
-<emph>Excluded US-ASCII Characters</emph>, do not appear in an unescaped form in the lexical representation of a <code>&lt;uri&gt;</code> value.</p></note>
+<note role="clarification">
+  <p>For avoidance of doubt, characters designated to be excluded in a URI by <bibref ref="uri"/>, &sect;2,
+  <emph>Characters</emph>, and, more specifically, SPACE (U+0020) characters,
```
nigelmegitt: Why call out SPACE characters particularly?

skynavga: In RFC2396, the following were explicitly excluded in 2.3.4 (meaning they are escaped)

```
control     = <US-ASCII coded characters 00-1F and 7F hexadecimal>
space       = <US-ASCII coded character 20 hexadecimal>
```

while these are not explicitly excluded in the RFC3986 in section 2 (though perhaps one may 
consider this implied).

But I see that calling out SPACE alone is not sufficient, and I really want all XML whitespace characters, so I am elaborating this point in the newest commit in this branch, which perhaps you can approve.

nigelmegitt: I never got the chance to approve this or not since you merged it so quickly, but I'm fine with it as luck has it. If you're going to ask me to approve though, it'd be polite to give me the opportunity to do so!

PR #506, opened 2017-12-23
Add @ttm:alt as shorthand for ttm:item name="altText" (#490).

### LINKED ISSUE #490: Clarify use of altText named metadata item.
See also #479.

skynavga:
Closes #490.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -3337,6 +3337,7 @@ Metadata Attributes
 </td>
 <td>
 <loc href="#metadata-attribute-agent">ttm:agent</loc>,
+<loc href="#metadata-attribute-alt">ttm:alt</loc>,
 <loc href="#metadata-attribute-role">ttm:role</loc>
 </td>
 </tr>
@@ -17203,7 +17204,7 @@ region that is temporally inactive must not produce any visible marks when
 presented on a visual medium.</p>
 <note role="motivation">
 <p>An <loc href="#terms-out-of-line-region">out-of-line</loc> <el>region</el> element may be associated with a time interval for two
-purposes: (1) in order to temporally bound the presentation of the region and
+purposes: (1) in order to temporally bind the presentation of the region and
 its content, and (2) to provide a temporal context in which animations of region
 styles may be effected.</p>
 <p>For example, an author may wish to specify an <loc href="#terms-out-of-line-region">out-of-line</loc> <el>region</el> element that is otherwise empty, but
@@ -19544,6 +19545,7 @@ for use with the <el>metadata</el> element and with certain
 <loc href="#terms-content-element">content element</loc> types:</p>
 <ulist>
 <item><p><specref ref="metadata-attribute-agent"/></p></item>
+<item><p><specref ref="metadata-attribute-alt"/></p></item>
 <item><p><specref ref="metadata-attribute-role"/></p></item>
 </ulist>
 <note role="explanation">
@@ -19565,6 +19567,36 @@ involved in the performance of the content.</p>
 <p>An example of the <att>ttm:agent</att> attribute is shown above in
 <specref ref="metadata-vocabulary-agent-example-1"/>.</p>
 </div3>
+
+<div3 id="metadata-attribute-alt">
+<head>ttm:alt</head>
+<p>The <att>ttm:alt</att> attribute serves as a shorthand expression for an equivalent
+<code>&lt;ttm:item name="altText"/&gt;</code> element, where the value of the attribute
+adheres to <loc href="http://www.w3.org/TR/xmlschema-2/#string">xsd:string</loc>.</p>
+<p>If both forms are applied to the same element, then the longhand form, i.e., the <code>&lt;ttm:item name="altText"/&gt;</code> element,
+applies and the shorthand form is ignored.</p>
+<p>The following example depicts equivalent specifications of alternate text to be applied to a given <code>image</code> element.</p>
+<table id="metadata-attribute-alt-example-1" role="example">
+<caption>Example Fragment &ndash; Alternate Text Equivalents</caption>
+<tbody>
+<tr>
+<td>
+<eg xml:space="preserve">
+...
+&lt;image <phrase role="strong">ttm:alt="Alternate Text"</phrase>&gt;
+  &lt;metadata&gt;
+    <phrase role="strong">&lt;ttm:item name="altText"/&gt;Alternate Text&lt;/ttm:item&gt;</phrase>
+  &lt;/metadata&gt;
+&lt;/image&gt;
+...
+</eg>
+</td>
+</tr>
+</tbody>
+</table>
+<p></p>
+</div3>
+
 <div3 id="metadata-attribute-role">
 <head>ttm:role</head>
 <p>The <att>ttm:role</att> attribute may be used by a content author
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -19570,9 +19570,17 @@ involved in the performance of the content.</p>
 
 <div3 id="metadata-attribute-alt">
 <head>ttm:alt</head>
-<p>The <att>ttm:alt</att> attribute serves as a shorthand expression for an equivalent
-<code>&lt;ttm:item name="altText"/&gt;</code> element, where the value of the attribute
+<p>The <att>ttm:alt</att> attribute serves as a shorthand expression for a semantically equivalent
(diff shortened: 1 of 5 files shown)


nigelmegitt:
Built version can be found at https://rawgit.com/w3c/ttml2/issue-0490-add-ttm-alt-build/index.html

spec/ttml2.xml
```diff
@@ -19565,6 +19567,36 @@ involved in the performance of the content.</p>
 <p>An example of the <att>ttm:agent</att> attribute is shown above in
 <specref ref="metadata-vocabulary-agent-example-1"/>.</p>
 </div3>
+
+<div3 id="metadata-attribute-alt">
+<head>ttm:alt</head>
+<p>The <att>ttm:alt</att> attribute serves as a shorthand expression for an equivalent
```
nigelmegitt: Do we need to state that the applicable `xml:lang` on the element applies, and that `xml:space` is considered to be `"preserve"` on attribute values?

Also probably worth noting that if these values need to be set explicitly, or indeed any other attribute on an `ttm:item  name="altText"` then the element form needs to be used.

nigelmegitt:
@palemieux I don't agree that the altText attribute prose must match IMSC 1.0.1 ittm:altText, because a) it can't - see https://github.com/w3c/ttml2/issues/490#issuecomment-353752537 and b) it is reasonable for TTML2 to be more general and for profiles to be more specific.

Any modifications that are worth making to align the TTML2 structure with the IMSC one, such as usage notes, would need to be made on the `ttm:item` section that defines the `altText` name rather than the short form attribute version since the latter simply refers to the former. Would you consider raising any such issue against [14.3.2 `<named item>`](https://www.w3.org/TR/ttml2/#metadata-value-named-item) rather than requiring it on this pull request?

spec/ttml2.xml
skynavga: sounds reasonable, will add

skynavga:
@palemieux regarding

>shouldn't the[re] be a feature that includes the ttm:alt attribute?

I should normally say no, since TTML1 doesn't define a feature for each metadata vocabulary item, but instead, coalesces them into one feature (#metadata). However, I anticipate that you might want to include support only for ``@ttm:alt`` in a profile, but exclude support for ``ttm:item``, so I will define individual features.

skynavga:
@palemieux see 69867ff9f4632a249f2269fd56e4e782edec73f3

skynavga:
@nigelmegitt added schema update and note/prose on lang/space

skynavga:
@palemieux are your comments addressed?

cconcolato:
To address @palemieux, maybe we could add some text to "14.3.2 <named-item>" in the "altText" part, not in the new attribute section. What seems to be missing are:
- this sentence:
> This text equivalent may be used to support indexing of the content and also facilitate quality checking of the document during authoring.
- and the NOTE

skynavga:
@cconcolato in general, I am opposed to enumerating or describing applications of a feature, particularly when it is a feature that has a long history of various applications.

also, what do you mean by "and the NOTE"?

cconcolato:
I was talking about the NOTE in IMSC which talks about how it is different from HTML5.

skynavga:
That note is irrelevant to TTML (and IMSC for that matter), and adds nothing that clarifies or explains TTML metadata semantics. If we went down this path, we would need notes explaining that TTML {head,body,div,p,span,br} are distinct from HTML vocabulary of the same name.

palemieux:
In preparation for the F2F meeting, below is a summary of my objection.

- IMSC1 introduced `ittm:altText`. The prose around `ittm:altText` was subject of much discussion, including the note at the end of 7.7.4 and the MAY in the introductory paragraph.

- TTML2 eventually introduced `<ttm:item name="altText"/>` to serve a similar purpose. [ed.: it is still not clear to me what motivated using a generic `ttm:item` element plus `@name`, instead of simply introducing individual elements, i.e. using native XML extension mechanisms.]

- the discussion at the TTWG F2F pointed towards using an `@ttm:alt` attribute instead of `<ttm:item name="altText"/>`.

This PR, which (i) proposes to introduce `@ttm:alt` *in addition* to `<ttm:item name="altText"/>` and (ii) does not match IMSC1 prose, has the following pitfalls:

- supporting both `@ttm:alt` and `<ttm:item name="altText"/>`  complicates implementations, especially since the mapping is not one-to-one
- both `@ttm:alt` and `<ttm:item name="altText"/>` diverge syntactically from `ittm:altText`
- additional prose will be required in IMSC1.1 to backfill any prose absent from TTML2, and constrain out `<ttm:item name="altText"/>`, thereby somewhat defeating the purpose of IMSC1.1 referencing TTML2
- the new prose in both TTML2 and IMSC1.1 may yield new comments whereas the prose in IMSC1 is presumably acceptable

Assuming there is a desire to preserve the altText element, I think the simplest path forward is to:

- move `ittm:altText` to the TTML2 namespace (as `ttm:altText`)
- deprecate `ittm:altText` in IMSC 1.1
- use `ttm:altText` in IMSC 1.1

Alternatively, IMSC 1.1 could continue supporting `ittm:altText`, regardless of what TTML2 does.

skynavga:
On Tue, Jan 2, 2018 at 3:49 PM, Pierre-Anthony Lemieux <

> In preparation for the F2F meeting, below is a summary of my objection.
>
>    -
>
>    IMSC1 introduced ittm:altText. The prose around ittm:altText was
>    subject of much discussion, including the note at the end of 7.7.4 and the
>    MAY in the introductory paragraph.
>    -
>
>    TTML2 eventually introduced <ttm:item name="altText"/> to serve a
>    similar purpose. [ed.: it is still not clear to me what motivated using a
>    generic ttm:item element plus @name, instead of simply introducing
>    individual elements, i.e. using native XML extension mechanisms.]
>
> The reason ttm:item was introduced was to avoid defining new attributes
over time, but instead, have a generic mechanism that would serve to
support all future items that might have been specified as attributes.

>
>    - the discussion at the TTWG F2F pointed towards using an @ttm:alt
>    attribute instead of <ttm:item name="altText"/>.
>
> I do not agree with this characterization, in particular, your use of
"instead of". At the time, I clearly recall saying it would be a shorthand
for the existing ttm:item form.

> This PR, which (i) proposes to introduce @ttm:alt *in addition* to <ttm:item
> name="altText"/> and (ii) does not match IMSC1 prose, has the following
> pitfalls:
>
>    - supporting both @ttm:alt and <ttm:item name="altText"/> complicates
>    implementations, especially since the mapping is not one-to-one
>
> actually, it is one-to-one (injective), in both directions, but it is not
onto (surjective), from ttm:alt to ttm:item

>
>    - both @ttm:alt and <ttm:item name="altText"/> diverge syntactically
>    from ittm:altText
>
> there is no (agreed) requirement that TTML2 incorporate IMSC1.0 features
while maintaining exact syntax or prose; the requirement is to support
equivalent functionality, which ttm:alt and its longhand form do

>
>    - additional prose will be required in IMSC1.1 to backfill any prose
>    absent from TTML2, and constrain out <ttm:item name="altText"/>,
>    thereby somewhat defeating the purpose of IMSC1.1 referencing TTML2
>    - the new prose in both TTML2 and IMSC1.1 may yield new comments
>    whereas the prose in IMSC1 is presumably acceptable
>
> it is not acceptable to Skynav, e.g., the note referencing HTML is not
acceptable, nor is the

>
>
> Assuming there is a desire to preserve the altText element, I think the
> simplest path forward is to:
>
>    - move ittm:altText to the TTML2 namespace (as ttm:altText)
>
> Skynav would object to this due to (1) there being (1) no (agreed)
requirement to incorporate this syntax directly, and (2) the extra four
characters required by the name altText as opposed to alt. The reason this
was proposed was to provide a shorthand: adopting a "middlehand" between a
reasonable shorthand (ttm:alt) and its stated longhand obviates the purpose
of this process.

>
>    - deprecate ittm:altText in IMSC 1.1
>    - use ttm:altText in IMSC 1.1
>
> Alternatively, IMSC 1.1 could continue supporting ittm:altText,
> regardless of what TTML2 does.
>
The simplest solution in my thinking is for IMSC1.1 to adopt ttm:alt and
also continue supporting ittm:altText.

nigelmegitt:
WG resolves not to merge this and not to add the ttm:altText attribute. See the issue.

skynavga:
Per https://github.com/w3c/ttml2/issues/490#issuecomment-356469536, have resolved not to add ``@ttm:alt``; therefore, closing this PR.

PR #510, opened 2017-12-24
Improve tts:rubyAlign prose syntax (#489).

### LINKED ISSUE #489: The "otherwise" clauses on tts:rubyAlign apply too broadly
In §10.2.3.37 rubyAlign, the "otherwise" clauses seem to apply too broadly - they're not supposed to be e.g. "otherwise if rubyAlign is not `auto`" etc.

Needs an edit to specify subsections of each case more clearly.

skynavga:
Closes #489.


nigelmegitt:
Built version available at https://rawgit.com/w3c/ttml2/issue-0489-ruby-align-build/index.html

PR #511, opened 2017-12-24
Add indefinite value to @end on isd:isd (#484).

### LINKED ISSUE #484: All ISDs must end, but not all TTML documents must end
It is possible to construct a TTML document with an ISD that has an indefinite end time but it is not possible to express this in an isd:isd document, because the end attribute on the root element is required and <timeExpression> does not permit a value equivalent to the SMIL keyword "indefinite" - I will raise that as a separate issue.

For example what ISD should be created for a TTML document with no `begin` or `end` attributes present at all?:

```xml
<tt>
<head/>
<body>
<div>
<p>some text that is visible from 0s forever</p>
</div>
</body>
</tt>
```

I propose:

* Make the `isd:isd/@end` attribute optional, so that its absence means "no end time".
* Add a note that the restriction on `isd:sequence` that prevents overlapping ISDs means that an ISD with no `end` attribute must be the final ISD in the sequence.

skynavga:
Closes #484.


nigelmegitt:
Built version available at https://rawgit.com/w3c/ttml2/issue-0484-isd-indefinite-end-build/index.html

PR #514, opened 2017-12-25
Improve prose semantics for tts:showBackground (#498).

### LINKED ISSUE #498: Does showBackground="always" apply outside a region's begin and end times?
The following recently was raised in face to face discussions. In an example like:

```xml
...
<region xml:id="r1" tts:showBackground="always" tts:backgroundColor="red" 
        tts:position="..." tts:extent="..." 
        begin="10s" end="20s"/>
...
```

does the red background show only during 10s to 20s or also at other times?

skynavga:
Closes #498.


nigelmegitt:
Built version available at https://rawgit.com/w3c/ttml2/issue-0498-show-background-build/index.html

cconcolato:
The first part:
> the region's background is rendered only when content is selected into the region, where that content is also temporally active;

could probably be reduced to:
>  the region's background is rendered only when temporally active content is selected into the region

The second part:
> the region's background is rendered irrespective of whether content is selected into the region.

could be:
> the region's background is rendered irrespective of whether temporally active content is selected into the region.

nigelmegitt:
@cconcolato Re the second part, the temporal activity state of the content selected into the region is genuinely not relevant - even if no content is selected into the region at all the background would still be rendered. So I'm happy to leave that one as is. I agree the first change you proposed would be more concise, but I'm neutral about whether it is worth doing or not - for me, I'd be happy with it as is, given how much other work there is to do.

PR #515, opened 2017-12-27
Remove support for ruby{Overflow,Overhang} (#494).

### LINKED ISSUE #257: Make it clearer that tts:rubyOverflow only applies to text that is at the end of a line
10.2.36 tts:rubyOverflow
https://www.w3.org/TR/2016/WD-ttml2-20161117/#style-attribute-rubyOverflow

> If specified, the value of tts:rubyOverflow expresses how to resolve potential overflow scenarios where ruby text may otherwise overflow its containing block in line start or end edge contexts.

That sentence clarifies the intended application of this property iiuc: ie. to do with line start/end situations and _not_ about overhanging adjacent text (which is dealt with by rubyOverhang). Such scenarios are mentioned as 'an' example of use in the first para in the section, leading to ambiguity about the applicability of this property when you read the section.

I think the sentence quoted above should come at the start of the section, and the text clarified to more clearly indicate that this is all about scenarios where the annotation at the start/end of a line is wider than the base text.

### LINKED ISSUE #259: Simplify description of shiftRuby (editorial)
10.2.36 tts:rubyOverflow
https://www.w3.org/TR/2016/WD-ttml2-20161117/#style-attribute-rubyOverflow

Editorial (i think, unless i'm missing something – which is certainly possible).

> If the value of this attribute is shiftRuby, then overflow is prevented by relaxing the ruby text to base text alignment by inseting the ruby text while preserving base text to line edge alignment. When inseting ruby text, the semantics of tts:overHang apply for the purpose of determining whether and how much overhang is permitted between the inset ruby text and adjacent base text.
> 
> When inseting ruby text or base text in order to resolve a potential overflow scenario, only the minimum (non-negative) inset is applied.

Isn't this rather long description about insetting the ruby text just a question of saying align both ruby text and base text with the line edge?

### LINKED ISSUE #260: Should rubyOverhang:allow overlap ideographs in Japanese?
10.2.38 tts:rubyOverhangClass
https://www.w3.org/TR/2016/WD-ttml2-20161117/#style-attribute-rubyOverhangClass

My reading of jLReq is that by default it does not allow overlap of kanji characters. See the [bullet points just below Fig 136](https://w3c.github.io/jlreq/#adjustments_of_ruby_with_length_longer_than_that_of_the_base_characters).

> a. Ruby text shall not hang over the ideographic characters (cl-19) adjacent to the base characters.

Notes further down admit alternative styles, one of which allows overhang of kanji, the other forbids any overhang at all, but these seem not to be mainstream.

Therefore, i suspect that 

> If the language of the base text is specified or can be inferred as Japanese, then auto should be interpreted as if the following were specified: ideographic hiragana punctuationMarks.

should drop the `ideographic`.

### LINKED ISSUE #494: Defer ruby overflow and ruby overhang to TTML3
Having discussed the currently specified features tts:ruby{Overhang,Overflow}* with a number of potential users of these features, it appears that there is insufficient demand for them in TTML2 to justify the necessary refinement of their specification, implementation, and testing. Therefore, I propose they be removed from TTML2 and marked as a possible addition for TTML3.

skynavga:
Closes #494.
Closes #260.
Closes #259.
Closes #257.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -3430,9 +3430,6 @@ Styling Attributes
 <loc href="#style-attribute-ruby">tts:ruby</loc>,
 <loc href="#style-attribute-rubyAlign">tts:rubyAlign</loc>,
 <loc href="#style-attribute-rubyOffset">tts:rubyOffset</loc>,
-<loc href="#style-attribute-rubyOverflow">tts:rubyOverflow</loc>,
-<loc href="#style-attribute-rubyOverhang">tts:rubyOverhang</loc>,
-<loc href="#style-attribute-rubyOverhangClass">tts:rubyOverhangClass</loc>,
 <loc href="#style-attribute-rubyPosition">tts:rubyPosition</loc>,
 <loc href="#style-attribute-rubyReserve">tts:rubyReserve</loc>,
 <loc href="#style-attribute-showBackground">tts:showBackground</loc>,
@@ -7961,9 +7958,6 @@ that support inline style specifications:</p>
 <item><p><specref ref="style-attribute-ruby"/></p></item>
 <item><p><specref ref="style-attribute-rubyAlign"/></p></item>
 <item><p><specref ref="style-attribute-rubyOffset"/></p></item>
-<item><p><specref ref="style-attribute-rubyOverflow"/></p></item>
-<item><p><specref ref="style-attribute-rubyOverhang"/></p></item>
-<item><p><specref ref="style-attribute-rubyOverhangClass"/></p></item>
 <item><p><specref ref="style-attribute-rubyPosition"/></p></item>
 <item><p><specref ref="style-attribute-rubyReserve"/></p></item>
 <item><p><specref ref="style-attribute-showBackground"/></p></item>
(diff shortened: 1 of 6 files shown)


nigelmegitt:
Built version available at https://rawgit.com/w3c/ttml2/issue-0494-defer-ruby-overflow-build/index.html

spec/ttml2.xml
```diff
@@ -12067,6 +12061,26 @@ base delimiter text delimiter =
 <p>The inter-character positioning mode for ruby annotations, such as used with <emph>bopomofo</emph> characters,
 is not supported by this version of this specification due to lack of market requirements.</p>
 </note>
+<div4>
+<head>Special Semantics</head>
```
nigelmegitt: I'm not able to review the special semantics section in detail - would appreciate e.g. @cconcolato 's review of this please.

palemieux:
Diff at https://services.w3.org/htmldiff?doc1=https%3A%2F%2Frawgit.com%2Fw3c%2Fttml2%2Fgh-pages%2Findex.html&doc2=https%3A%2F%2Frawgit.com%2Fw3c%2Fttml2%2Fissue-0494-defer-ruby-overflow-build%2Findex.html#isd-vocabulary-css

palemieux:
I do not understand why "10.2.36.1 Special Semantics" was introduced as a separate section: the semantics do not seem anymore special than other ruby semantics.

skynavga:
@palemieux because we do not in general specify default presentation behavior that, strictly speaking, is not related to tts:ruby; here, we essentially define default behavior that would have been derived from the initial values we had assigned to tts:{rubyOverflow,rubyOverhang}

palemieux:
@skynavga The reader has no idea that `tts:{rubyOverflow,rubyOverhang}` used to exist, and so the semantics directly apply to rendering of rubies, so there is really nothing special. Anyway I have approved the PR.

skynavga:
well, you can consider it special in the sense that this behavior was not described in [1]

[1] http://www.w3.org/TR/2001/REC-ruby-20010531/

PR #516, opened 2017-12-28
Remove auto value from annotation position (#481).

### LINKED ISSUE #481: Remove annotation-position "auto" used with ruby position and text emphasis.
According to https://lists.w3.org/Archives/Public/public-tt/2017Nov/0007.html , `rubyPosition="auto"` should take same behavior as `'outside'`.

skynavga:
Closes #481.

### DIFF
--- a/spec/rnc/ttml2-datatypes.rnc
+++ b/spec/rnc/ttml2-datatypes.rnc
@@ -18,7 +18,6 @@ TTAF.AnimationValueList.datatype =
   string
 
 TTAF.AnnotationPosition.datatype =
-  "auto" |
   "before" |
   "after" |
   "outside"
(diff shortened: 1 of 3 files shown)


palemieux:
Diff at https://services.w3.org/htmldiff?doc1=https%3A%2F%2Frawgit.com%2Fw3c%2Fttml2%2Fgh-pages%2Findex.html&doc2=https%3A%2F%2Frawgit.com%2Fw3c%2Fttml2%2Fissue-0481-remove-ruby-position-auto-build%2Findex.html

nigelmegitt:
I observe that the appendix N mapping of emphasis position to CSS appears to be incorrect, in that the CSS values are `[ over | under ] && [ right | left ]`. That's for another issue though.

PR #518, opened 2017-12-29
Add #{presentation,profile,transformation}-version-2 feature designat…

### LINKED ISSUE #473: ttp:profile has been extended from TTML1, but no new feature has been defined
A feature that expresses support for the following attributes (not present in TTML1) is necessary:

- combine
- designator
- type

skynavga:
…ors (#473).

Closes #473.

### DIFF
--- a/spec/profiles/ttml2-full.xml
+++ b/spec/profiles/ttml2-full.xml
@@ -142,8 +142,10 @@
     <feature value="required">#pixelAspectRatio</feature>
     <feature value="required">#position</feature>
     <feature value="required">#presentation</feature>
+    <feature value="required">#presentation-version-2</feature>
     <feature value="required">#processorProfiles</feature>
     <feature value="required">#profile</feature>
+    <feature value="required">#profile-version-2</feature>
     <feature value="required">#region-inline</feature>
     <feature value="required">#region-inline-explicit</feature>
     <feature value="required">#region-inline-implicit</feature>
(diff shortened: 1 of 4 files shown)


cconcolato:
I would suggest avoiding "version-2" in the naming of the feature designators, as this could be misinterpreted as being related to the "version" attribute. But I don't have good alternative names: profile-extended/presentation-profile-extended/transformation-profile-extended
or with 'attributes' instead of 'extended'.

palemieux:
Diff at https://services.w3.org/htmldiff?doc1=https%3A%2F%2Frawgit.com%2Fw3c%2Fttml2%2Fgh-pages%2Findex.html&doc2=https%3A%2F%2Frawgit.com%2Fw3c%2Fttml2%2Fissue-0473-profile-version-2-build%2Findex.html#feature-profile-version-2

skynavga:
I'm a big believer in RTFM. Since "-2" would mean precisely "-version-2", I see no problem here. Further, the only designator with a "-N" suffix is #padding-2, where the 2 refers to the number of length specifications. So the -N proposal is not good.

cconcolato:
@skynavga where is the FM?

skynavga:
@cconcolato that would be the definitions of the individual ``-version-2`` designators, where it is clear it has nothing to do with the ``ttp:version`` per se.

skynavga:
@cconcolato btw, I would not object to adding a note at the end of the prologue of section E.1 that says something like

>Designators with a suffix ``-version-2`` are feature designators define by this (second) version of TTML, i.e., TTML2, that extend or restrict the a related feature defined by TTML1.

cconcolato:
@skynavga adding a note to E.1 would help avoid the confusion.

nigelmegitt:
I'd be happy with the note, or even changing `-version-2` suffixes to `-ttml2` which would probably be even clearer. Using `-2` would be confusing since that is already in `padding-2` to mean the 2 value version of the feature.

nigelmegitt:
In reviewing this I realised that there are two attributes not covered here, that possibly should be: see https://github.com/w3c/ttml2/issues/473#issuecomment-354897466

PR #520, opened 2017-12-29
Add display alignment and fix timing in examples in Appendix S (#459).

### LINKED ISSUE #459: Examples in section S use an unusual tts:displayAlign setting
See w3c/ttml1#254.

skynavga:
Closes #459.


nigelmegitt:
Built version is at https://rawgit.com/w3c/ttml2/issue-0459-appendix-s-build/index.html

Diff is at https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fttml2%2Findex.html&doc2=https%3A%2F%2Frawgit.com%2Fw3c%2Fttml2%2Fissue-0459-appendix-s-build%2Findex.html

spec/examples/ex2-pop-on.xml
```diff
@@ -2,8 +2,10 @@
   xmlns:ttp="http://www.w3.org/ns/ttml#parameter" xmlns:tts="http://www.w3.org/ns/ttml#styling">
   <head>
     <layout>
-      <region xml:id="r1" tts:color="white" tts:origin="10c 4c" tts:extent="40c 1c"/>
-      <region xml:id="r2" tts:color="yellow" tts:origin="10c 8c" tts:extent="40c 1c"/>
+      <region xml:id="r1" tts:color="white" tts:origin="10c 4c" tts:extent="40c 1c"
+              tts:displayAlign="after"/>
+      <region xml:id="r2" tts:color="yellow" tts:origin="10c 8c" tts:extent="40c 1c"
+              tts:displayAlign="after"/>
```
nigelmegitt: Modifying the Pop On example makes no difference to this example, and we agreed not to change this one. I think it better to revert the changes to this example. The descriptive paragraph above includes:

> Each paragraph is non-overlapping in time and appears in the same single row of its targeted region.

spec/examples/ex2-roll-up.xml
```diff
@@ -2,16 +2,19 @@
...
   </head>
   <body>
     <div region="r1">
-      <p begin="0s" end="8s">Lorem ipsum dolor sit</p>
-      <p begin="4s" end="12s">Amet consectetur adipiscing elit</p>
-      <p begin="8s" end="18s">Sed do eiusmod tempor incididunt labore</p>
-      <p begin="14s" end="25s">et dolore magna aliqua</p>
-      <p begin="18s" end="29s">Ut enim ad minim veniam quis, nostrud</p>
+      <p>
+        <span begin="0s" end="8s">Lorem ipsum dolor sit<br/></span>
```
nigelmegitt: Shouldn't the additional `<br/>` elements be at the beginning of all but the first `<span>` rather than at the end of all of them? Otherwise the first `<span>` is never on the bottom line, since the `<br/>` always bumps it up one. The text is only ever bumped up by the next line becoming active.

spec/examples/ex2-paint-on.xml
```diff
@@ -2,7 +2,8 @@
   xmlns:ttp="http://www.w3.org/ns/ttml#parameter" xmlns:tts="http://www.w3.org/ns/ttml#styling">
   <head>
     <layout>
-      <region xml:id="r1" tts:color="white" tts:origin="10c 4c" tts:extent="40c 5c"/>
+      <region xml:id="r1" tts:color="white" tts:origin="10c 4c" tts:extent="40c 5c"
+              tts:displayAlign="after"/>
```
nigelmegitt: I think the consensus was not to modify the paint on example. The presentation does depend on the value of `displayAlign` but demonstrating the paint on effect does not. It works equally well with `before` and `after` but arguably, since the region is defined near the top, the default `before` makes sense here. Otherwise the example gets mixed with roll up. I think it better to revert this change.

spec/examples/ex2-roll-up.xml
skynavga: @nigelmegitt an empty line area at the after edge of a block is always elided, so using ``<br/>`` at the ends of lines suffices (and looks better in example code).

skynavga:
@nigelmegitt please check with fixes

spec/examples/ex2-roll-up.xml
nigelmegitt: @skynavga I wondered if that is supposed to be the case. I'm aware that some documents use multiple trailing `<br/>`s in a region with `tts:displayAlign="after"` to position subtitles  some lines off the bottom of the region. Does that mean that never works? 

This technique is one of the strategies defined in [EBU Tech3360](https://tech.ebu.ch/docs/tech/tech3360.pdf) - see the example at §4.5.6.3.1 for example. This suggests that at least some implementations do not elide in the way that you say. 

Since this is unclear in the world (regardless of what is correct according to any specifications), I would prefer to avoid the confusion, by moving the `<br/>`s in this example to go at the beginning of each `<p>` other than the first one.

skynavga:
@nigelmegitt on further checking, I see that we map ``br`` to

```
<fo:character character="&#x000A;" suppress-at-line-break="retain/>
```

which has semantics

>**retain**
>The glyph area generated by the fo:character shall be placed in the area tree whether or not it first or last in a line-area.

I need to verify behavior on an XSL-FO implementation, but it may be that the start of line or end of line will produce the same behavior.

nigelmegitt:
Thanks for digging into that @skynavga - this confirms my view that we need to remove trailing `<br/>`s from the example.

skynavga:
@nigelmegitt please check

Issue #442, opened 2017-09-30
Make backward compatibility normative.


tairt:
Section 3.4.2 

> A conforming TTML1 content document instance is intended to be a conforming TTML2 content document instance.
A conforming TTML2 processor should be able to process a TTML1 content document instance; however, it may emit warnings if it encounters deprecated features.

Should be changed to

"A conforming TTML1 content document instance **is a** conforming TTML2 content document instance.

A conforming TTML2 processor **must** be able to process a TTML1 content document instance; however, a **validating processor** may emit warnings if it encounters deprecated features."

nigelmegitt:
It does seem sensible to treat this as a duplicate of #458 and deal with both sets of comments in a single pull request. @tairt would that work for you?

tairt:
This issue was opened as WR comment before #458, so it is actually the other way around. But I see that there has been a longer discussion in #458 so I am fine if this issue is resolved in a single pull request. This issue should stay open until it is resolved by that PR.

tairt:
I uncheck the label duplicate after reading through #458. The comment in this issue requests normative language. Discussion and PR for #458 do not address this.

skynavga:
I believe this issue has been addressed by #521. @tait please confirm.

tairt:
Yes, this has been addressed by #521 and closing therefore this issue.

PR #521, opened 2017-12-29
Clarify backward compatibility (#458).

### LINKED ISSUE #458: Processor backwards compatibility.
This is a proposal based on the discussion from the 5-Oct WG meeting on this topic. See also: w3c/ttml2#442 and w3c/imsc1#258.

Background: several of us are concerned about backwards compatibility of TTML1/TTML2 presentation processors, especially in the context of IMSCvNEXT. Specifically, the concern is that a conformant IMSC1 document (optionally including TTML2 features such as tts:disparity) creates the same presentation out of an IMSCvNEXT presentation processor as an IMSC1 presentation processor.  And, not because it might be the same author of the implementation.  This potentially could be solved with conformance statements in IMSCvNEXT except to the extent it references TTML2 (rather than TTML1) when TTML2 itself is potentially conflicting (at least for documents it determines to be version=2) and not subject to such a constraint.

Examining TTML2… 

Re section 3.4:

3.4.2 …A conforming TTML2 processor should be able to process a TTML1 content document instance; however, it may emit warnings if it encounters deprecated features.

Why is the above a “should” and not a “shall”?  That is, when would it be permitted/desirable for it to deviate?

There is no specific statement about presentation processor backwards compatibility. The fact that it is also a “processor” is inadequate since TTML2 is generally silent about presentation details across implementations as well as between TTML1 and TTML2, except for an informative note listing example permitted variations across implementations, inferring there are more that go unstated:

3.2.3 "…Note:
Two conforming presentation processors are not required to produce the same presentation when processing the same conforming document. For example, different presentation processors may use different fonts, different font rasterizers, different line break algorithms, as well as different implementations of behavior not prescribed by this specification."

Although this is an accurate general representation of the potential variations between implementations, the silence of the substantive statement that is desirable: “all other things being equal, the presentation shall be deterministic and consistent across implementations”. But this is testable without an exhaustive list of permitted implementation-dependent variations.

Whether we alter the above conformance provisions or not, I think I am leaning to make a broad scope statement about the intent of the TTML2 specification itself of the form: “The provisions of this specification covering TTML1 features are intended to be identical to those same provisions in TTML1. That is, any deviation is unintended.”

skynavga:
Closes #458.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -2134,25 +2134,9 @@ support for a <loc href="#terms-feature">feature</loc> not defined by <bibref re
     <head>Backward Compatibility</head>
     <p>A conforming TTML1 <loc href="#terms-timed-text-content-document-instance">content document instance</loc>
     is intended to be a conforming TTML2 <loc href="#terms-timed-text-content-document-instance">content document instance</loc>.</p>
-    <p>A conforming TTML2 processor should be able to process a TTML1
-    <loc href="#terms-timed-text-content-document-instance">content document instance</loc>;
-    however, it may emit warnings if it encounters deprecated features.</p>
-    <note role="elaboration">
-      <p>The processing semantics of TTML1 and TTML2 are not equivalent. A specific implementation of a TTML2 processor may support only TTML2
-      semantics or may, in addition, support TTML1 semantics, in which case the implementation determines which semantics apply. For example, in
-      making this determination, an implementation may consider the effective content profile or, if the effective content profile is <code>null</code>,
-      then the effective processor profile to be used with a given <loc href="#terms-timed-text-content-document-instance">content document
-      instance</loc>.</p>
-      <p>If a given TTML2 processor supports both TTML1 and TTML2 semantics, then, unless explicitly required to behave otherwise by this specification,
-      the processor should produce equivalent results when processing a strictly conforming TTML1 document regardless of whether TTML1 or TTML2 semantics
-      apply.</p>
-      <p>In the present context, (1) the term <emph>equivalent</emph> is scoped by the range of variations
-      permitted between implementations that satisfy <specref ref="conformance-processor"/>, and, in particular,
-      <emph>equivalent</emph> does not imply <emph>identity</emph>, and (2) the phrase <emph>strictly conforming TTML1 document</emph> means
-      a TTML <loc href="#terms-timed-text-content-document-instance">content document instance</loc> that
-      satisfies <bibref ref="ttml1"/>, &sect;3.1, and, after foreign namespace vocabulary has been pruned,
-      only vocabulary defined by <bibref ref="ttml1"/>, &sect;5.3.1, remains.</p>
-    </note>
+    <p>A conforming TTML2 processor is intended to process a conforming TTML1
+    <loc href="#terms-timed-text-content-document-instance">content document instance</loc> such that the output produced by
+    the TTML2 processor is within the variations allowed per TTML1; however, it may emit warnings if it encounters deprecated features.</p>
   </div3>
 </div2> <!-- compatibility -->
 </div1> <!-- conformance -->
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -2137,6 +2137,22 @@ support for a <loc href="#terms-feature">feature</loc> not defined by <bibref re
     <p>A conforming TTML2 processor should be able to process a TTML1
     <loc href="#terms-timed-text-content-document-instance">content document instance</loc>;
     however, it may emit warnings if it encounters deprecated features.</p>
+    <note role="elaboration">
+      <p>The processing semantics of TTML1 and TTML2 are not equivalent. A specific implementation of a TTML2 processor may support only TTML2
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -2137,6 +2137,22 @@ support for a <loc href="#terms-feature">feature</loc> not defined by <bibref re
     <p>A conforming TTML2 processor should be able to process a TTML1
     <loc href="#terms-timed-text-content-document-instance">content document instance</loc>;
     however, it may emit warnings if it encounters deprecated features.</p>
+    <note role="elaboration">
+      <p>The processing semantics of TTML1 and TTML2 are not equivalent. A specific implementation of a TTML2 processor may support only TTML2
```
cconcolato: Saying they are not equivalent may be true, but it does not help. The conducted investigation discussed several times in the group seems to indicate that the difference are very subtle, and that for the most part, the semantics of TTML2 and TTML1 are the same. We should emphasize on that.

spec/ttml2.xml
```diff
@@ -2137,6 +2137,22 @@ support for a <loc href="#terms-feature">feature</loc> not defined by <bibref re
...
+      semantics or may, in addition, support TTML1 semantics, in which case the implementation determines which semantics apply. For example, in
```
cconcolato: The sentence ends quite abruptly. It could say:
When an implementation supports both semantics, the determination of which semantics apply depends on the document and processing context.

skynavga:
the decision to use document interchange context is implementation specific, so it is solely the implementation that decides what information to use in making its determination

nigelmegitt:
Built version available at https://rawgit.com/w3c/ttml2/issue-0458-backward-compatibility-build/index.html

Diff at https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fttml2%2Findex.html&doc2=https%3A%2F%2Frawgit.com%2Fw3c%2Fttml2%2Fissue-0458-backward-compatibility-build%2Findex.html

spec/ttml2.xml
nigelmegitt: I'm yet to be convinced that the processing semantics of TTML1 Third Edition and TTML2 will indeed not be equivalent. It seems to me that many of the changes we are making to TTML1 Third Edition will have the effect of lessening the delta between the processing semantics, to the point that it may disappear altogether. 

There's possibly a point missing in this section, which is that in some cases the TTML2 semantic is a permitted subset of the TTML1 semantics but the TTML1 semantics may be more broadly defined, such that a conformant TTML1 processor might not be a conformant TTML2 processor even for a TTML1 document.

spec/ttml2.xml
nigelmegitt: @cconcolato Doesn't that belong in the following paragraph?

spec/ttml2.xml
```diff
@@ -2137,6 +2137,22 @@ support for a <loc href="#terms-feature">feature</loc> not defined by <bibref re
     <p>A conforming TTML2 processor should be able to process a TTML1
     <loc href="#terms-timed-text-content-document-instance">content document instance</loc>;
```
palemieux: This is missing the key sentence from https://github.com/w3c/ttml2/issues/458#issuecomment-339822234 :

_a conforming TTML1 content document instance not containing any new TTML2 feature would be rendered by a TTML2 presentation processor in an acceptable way, conformant to TTML1, within the variations allowed per TTML1._

skynavga:
I just pushed an update to to the proposed text based on discussion between myself, @cconcolato and @palemieux.

spec/ttml2.xml
```diff
@@ -2134,9 +2134,9 @@ support for a <loc href="#terms-feature">feature</loc> not defined by <bibref re
     <head>Backward Compatibility</head>
     <p>A conforming TTML1 <loc href="#terms-timed-text-content-document-instance">content document instance</loc>
     is intended to be a conforming TTML2 <loc href="#terms-timed-text-content-document-instance">content document instance</loc>.</p>
-    <p>A conforming TTML2 processor should be able to process a TTML1
-    <loc href="#terms-timed-text-content-document-instance">content document instance</loc>;
-    however, it may emit warnings if it encounters deprecated features.</p>
+    <p>A conforming TTML2 processor is intended to process a conforming TTML1
```
nigelmegitt: I don't mind using "intended" language but this sentence as written doesn't make sense - it isn't the conforming TTML2 processor that is intended to do something (that intent is defined by the implementer of the processor, not us); rather it is the WG's intent that a conforming TTML2 processor is able to process a TTML1 document etc.

skynavga:
The editors team discussed this issue and language very thoroughly yesterday, and arrived at the proposed language after careful consideration. In particular, we agreed to use "is intended to" and not "is", since this section is non-normative, and sense it is making a statement about design intention as opposed to prescribing normative behavior.

skynavga:
Merged this editorial PR with two approvals and no requested change. If @nigelmegitt has further issue, please create new issue to address and refer to this issue and/or pr.

skynavga:
@tairt this merge should also address #442; please close #442 or indicate what additional change you require

PR #526, opened 2017-12-30
Refactor repeated prose regarding the specification of style property…

### LINKED ISSUE #438: Duplicated text for attributes in TT Style Namespace.
The following text is repeated for all 55 attributes of the TT Style Namespace:

>This attribute may be specified by any element type that permits use of attributes in the TT Style Namespaces; however, this attribute applies as a style property only to those element types indicated in the following table.

This text should be just defined once in e.g. as below:

"Attribute from the TT Style Namespace may be specified by any element type that permits use of attributes in the TT Style Namespaces; however, attributes apply as a style property only to those element types indicated in the corresponding table for that attribute."

This should also be changed for attributes in the TT Audio Style Namespace.

skynavga:
… attributes (#438).

Closes #438.


skynavga:
in order to cover 10.3 as well

cconcolato:
@skynavga indeed, makes sense then.

spec/ttml2.xml
```diff
@@ -14470,15 +14287,15 @@ that support inline style specifications:</p>
 <item><p><specref ref="audio-style-attribute-voice"/></p></item>
 -->
 </ulist>
-
+<p>Unless explicitly stated otherwise, linear white-space (LWSP) must
+appear between adjacent non-terminal components of a value of a TT Audio Style property
+value unless some other delimiter is permitted and used, in which case linear white-space may appear
+between a delimiter and a non-terminal component.</p>
```
nigelmegitt: This sensible change appears to be in the wrong pull request.

spec/ttml2.xml
```diff
@@ -7782,6 +7782,14 @@ to implement features it has in common with this model.</p>
 <p>No normative use of an
 <code>&lt;?xml-stylesheet ... ?&gt;</code> processing instruction is defined
 by this specification.</p>
+<p>The styling attributes defined in this section may be specified by any element type
+that permits use of attributes in the TT Style Namespaces; however, these attributes apply
```
nigelmegitt: I think we need to be explicit that this applies to both the TT Style Namespace and the TT Audio Style Namespace.

skynavga:
On Tue, Jan 2, 2018 at 4:38 PM, Nigel Megitt <notifications@github.com>
wrote:

> *@nigelmegitt* requested changes on this pull request.
>
> Minor fixes requested.
>
> In spec/ttml2.xml
> <https://github.com/w3c/ttml2/pull/526#discussion_r159339476>:
>
> > @@ -14470,15 +14287,15 @@ that support inline style specifications:</p>
>  <item><p><specref ref="audio-style-attribute-voice"/></p></item>
>  -->
>  </ulist>
> -
> +<p>Unless explicitly stated otherwise, linear white-space (LWSP) must
> +appear between adjacent non-terminal components of a value of a TT Audio Style property
> +value unless some other delimiter is permitted and used, in which case linear white-space may appear
> +between a delimiter and a non-terminal component.</p>
>
> This sensible change appears to be in the wrong pull request.
>
> In spec/ttml2.xml
> <https://github.com/w3c/ttml2/pull/526#discussion_r159339631>:
>
> > @@ -7782,6 +7782,14 @@ to implement features it has in common with this model.</p>
>  <p>No normative use of an
>  <code>&lt;?xml-stylesheet ... ?&gt;</code> processing instruction is defined
>  by this specification.</p>
> +<p>The styling attributes defined in this section may be specified by any element type
> +that permits use of attributes in the TT Style Namespaces; however, these attributes apply
>
> I think we need to be explicit that this applies to both the TT Style
> Namespace and the TT Audio Style Namespace.
>
It is. Note the 's' on "TT Style Namespaces" above.

nigelmegitt:
> It is. Note the 's' on "TT Style Namespaces" above.

@skynavga Sorry, but that wording is too subtle. I did see it, but I couldn't see anything that lists the "TT Style Namespaces" explicitly anywhere.

skynavga:
On Tue, Jan 2, 2018 at 5:08 PM, Nigel Megitt <notifications@github.com>
wrote:

> It is. Note the 's' on "TT Style Namespaces" above.
>
> @skynavga <https://github.com/skynavga> Sorry, but that wording is too
> subtle. I did see it, but I couldn't see anything that lists the "TT Style
> Namespaces" explicitly anywhere.
>
See 2nd para under table 5-1.

nigelmegitt:
> See 2nd para under table 5-1.

@skynavga thanks, I hadn't spotted that.

spec/ttml2.xml
nigelmegitt: Specifically I think it should be in #525. Would you be able to move it to there please @skynavga ?

spec/ttml2.xml
nigelmegitt: Perhaps we could resolve the possible confusion by wrapping "TT Style Namespaces" in a `<loc>` element that points to the place it is defined?

skynavga:
[Note to Self] Add link to (informal) definition of TT Style Namespaces.

skynavga:
@nigelmegitt please review fixes

spec/xmlspec-ttml2.xsl
```diff
@@ -1212,4 +1212,16 @@ width: 100%;
...
 </xsl:template>
 
+<!-- term: the actual mention of a term within a termdef -->
+<xsl:template match="term">
+  <em><xsl:apply-templates/></em>
+</xsl:template>
+
+<!-- termdef: sentence or phrase defining a term -->
+<xsl:template match="termdef">
+  <a id="{@id}" title="{@term}">
```
nigelmegitt: I'm not sure why but the latest changes did not trigger a new travis build, so the effect of this change is not visible on the version served by gh-pages... I've merged from master and resolved conflicts so hopefully that will trigger the build again.

tairt:
Latest Changes work for me.

Issue #76, opened 2015-11-04
Appendix N assumption that root temporal extent corresponds with the beginning of a related media object


plehegar:
From Appendix N.2:

"The above formalisms assumes that the Root Temporal Extent corresponds with the beginning of a related media object. If this assumption doesn't hold, then an additional offset that accounts for the difference may be introduced when computing media time M."

In a streaming environment it is very likely, especially for live-created subtitles, that a set of TTML samples will be produced each with a different Root Temporal Extent that starts at a different point relative to a single Related Media Object (which may itself be packaged up into an independent set of samples). However I would not in general account for this with an offset and assume (or hope) that every sample's begin and end times are zero-based; rather I'd expect the begin of the TT to be coincident with the point within the Related Media Object at which that sample begins. As a side-effect this should reduce the complexity of the transformations needed to accumulate the [short] samples together to create a longer one.

For this reason I propose that we reverse this assumption.

(raised by Nigel Megitt on 2013-08-12)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/270

nigelmegitt:
(admin point: at the current time, Appendix N in TTML1 maps to Appendix H in TTML2)

Having looked at this more, and after discussions, a couple of points need to be added - some but not all of this is restating other discussions elsewhere, for completeness here:

**Firstly**, the definition of Root Temporal Extent affects the treatment of the statement. If it is taken that the default for `begin` if unstated is effectively zero, then one might surmise that the Root Temporal Extent for e.g. the `tt` element always begins at zero. There are (at least) two other options however:

1. An undefined `begin` can be resolved when the document instance playback begins, which if it is on a media timeline might involve using some externally provided value, e.g. from an ISOBMFF box. In other words in this case the document processing context provides a syncbase which defines the beginning of the Root Temporal Extent, which may not be at the origin of the temporal coordinates. See below for a use case.
2. Treat the earliest computed begin time in the document as the beginning of the Root Temporal Extent.

A use case for 1 above  is when the document is one of a sequence of "live" TTML documents, where the sender cannot provide a reliable time coordinate that can be successfully compared to a clock at the receiver, and where the author wants to signal "display this as soon as you can, but not for longer than 5s after beginning". To support this use case (which for reference is present in EBU Tech3370) the active begin time needs to be _resolved_ downstream, where the resolution for that begin provides a syncbase value. An example document with no timing on regions might have:

```xml
<body dur="5s">
```

[...]

PR #536, opened 2018-01-06
Remove ttp:mediaOffset (#323).

### LINKED ISSUE #323: Remove ttp:mediaOffset.
See #125 for background. Raising this to capture my thoughts on mediaOffset that I have previously discussed during face to face meetings.

There are multiple problems with `ttp:mediaOffset` as currently specified:

#### Multiple presentation contexts

`ttp:mediaOffset` appears to require that exactly one related media object is relevant when deciding what media offset to apply at presentation time, in other words it fails if there is more than one presentation context, and different presentation contexts have different offset requirements. In general TTML documents are likely to be reused in different presentation contexts in an "author once, use many times" pattern.

#### Duplication of functionality present elsewhere

It duplicates functionality more usually put into wrapper formats - if it is only metadata from the TTML context then it should not be in the `ttp` namespace. It is unclear whether some additional time computation is required at presentation time or if the offset value should simply be transferred to the wrapper format. This ambiguity is likely to cause interoperability problems, especially if the offset is in some implementations effectively processed twice.

#### Circular definition

In Terminology:

> [root temporal extent]
> The temporal extent (interval) defined by the temporal beginning and ending of a document instance in relationship with some external application or presentation context.

In §7.2.10 ttp:mediaOffset
> The ttp:mediaOffset attribute is used to specify the temporal offset between the begin time of the root temporal extent and the begin time of a related media object ...

Expanding these, we seem to have a temporal offset between the beginning of a (temporal extent defined by the temporal beginning and ending of a document instance in relationship with some ... presentation context) and the begin time of a related media object. This doesn't appear to make any sense. If it is _defining_ the relationship between the document instance's begin time and the presentation context, then the above wording in §7.2.10 doesn't say that.

I am taking it that a presentation context includes possible playback of some related media object.

#### SMPTE timebase application problem

If `ttp:mediaOffset` applies to smpte timebase then this will usually result in playback failure - it is effectively saying although the author used SMPTE timecode to align the subtitles with the video, that timecode alignment is in fact wrong. The correct solution to this problem is to fix the times in the document instance, not to apply a media offset.

#### Media timebase application problem

It is a possible interpretation that `ttp:mediaOffset`specifies the relationship between media time and the document's timeline, in which case it cannot apply if timebase is "media", since in that timebase the relationship is already defined. 

#### Handling of negative times is undefined.

`ttp:mediaOffset`, if interpreted as being something that modifies all the computed times, can result in negative times, whose semantic is undefined.

#### Does not identify beginning of root temporal extent

`ttp:mediaOffset` does not actually support the (possibly orthogonal) need to define the beginning of intended presentation, i.e. the entry point in the document instance's timeline, or the point before which any ISDs should be truncated. This is asymmetric with the existing ability to specify the end of presentation, which can be done by putting a `dur` or `end` attribute on, say, the `body` element, or if the intent is for indefinite end, the end time can be left unspecified on some leaf elements. Currently all documents appear to begin at time zero regardless of the intended playback begin point. 

### Proposed resolution

1. Remove `ttp:mediaOffset`.
2. Add a `ttp:temporalBegin` (name can be discussed) that does not modify the syncbase of the root element but truncates all ISDs beginning before it so that the content does not appear. This deals with the "pre-roll" scenario for example, and can also safely be carried over to wrapper formats for example to derive fragment or segment times without needing to modify the computed times in the document instance. In other words, if duplicated in a wrapper it remains "safe".

skynavga:
Closes #323.


spec/ttml2.xml
```diff
@@ -24433,9 +24382,9 @@ and <code>tickRate</code> is the computed value of the <att>ttp:tickRate</att> p
 <p>If the computed <code>frameRateMultiplier</code> ratio is not integral, then <code>effectiveFrameRate</code> will be a non-integral rational.</p>
 </note>
 <note role="clarification">
-<p>The begin time of the <loc href="#terms-root-temporal-extent">root temporal extent</loc> is related to the begin time of a
-<loc href="#terms-related-media-object">related media object</loc> in accordance with the computed value of the
-<loc href="#parameter-attribute-mediaOffset"><att>ttp:mediaOffset</att></loc> property.</p>
+<p>The above formalisms assume that the begin time of the <loc href="#terms-root-temporal-extent">root temporal extent</loc> is related to the
+begin time of a <loc href="#terms-related-media-object">related media object</loc>. If this assumption doesn't hold, then an additional offset
+that accounts for the difference may be introduced when computing media time <code>M</code>.</p>
```
nigelmegitt: This matches TTML1 but I will raise a separate issue against this text on the basis that the temporal origin coordinates are assumed to be the same not just the begin times.

nigelmegitt: Raised as #550.

skynavga:
See also #76.

PR #537, opened 2018-01-06
Deprecate rather than obsolete ttp:feature value "use" (#357).

### LINKED ISSUE #357: Deprecate rather than obsolete ttp:feature value="use".
Given our decision in #331 to explicitly ensure that TTML1 instance documents remain TTML2 conformant, and not having yet deprecated this feature as a first step, it is not a good idea to obsolete this.  It should deprecated instead.

skynavga:
Closes #357.


cconcolato:
@skynavga This PR seems to close itself. What is the issue you want to close with this PR?

skynavga:
#537 != #357

PR #545, opened 2018-01-10
Correct semantics for DAR/PAR when SAR is specified (#539).

### LINKED ISSUE #539: Aspect ratios when SAR specified, but DAR and PAR not specified.
In order to be consistent with IMSC1 7.7.1, we need to update H.1.2 to read as follows (or equivalent):

>a. if a related media object exists and has a defined DAR, then the resolved value of SAR is the display aspect ratio of the related media object and the value of PAR is then resolved according to H.1.3 Two Aspect Ratios below;
>b. otherwise, the resolved value of DAR is determined by the document processing context and the value of PAR is then resolved according to H.1.3 Two Aspect Ratios below.

skynavga:
Closes #539.


skynavga:
On Sat, Jan 13, 2018 at 12:29 PM, Pierre-Anthony Lemieux <

> *@palemieux* requested changes on this pull request.
>
> I need the two weeks to review this
>
Since this PR is labeled as substantive, there is no need to request a full
two week review, since I am no longer performing early merges of
substantive PRs.

PR #553, opened 2018-01-14
Add tts:rubyAlign withBase as at risk (in placeholder status); add gl…

### LINKED ISSUE #236: Meaning of 'glyph area descendant'
[This was intended to be a quick query by email, but it is developing into a thread, so i'm copying the discussion here for ongoing discussion.]

my original question:
I'm working on review comments for the ruby section, but i'm a bit stuck because i don't understand the meaning of the term 'glyph area descendant', and i don't see it described anywhere.  Could one of you explain what that is?  I think it will be important to understand it in order to assess the ruby section.

skynavga:
…yph area descendant as defined term (#236).

Closes #236.


nigelmegitt:
Built version available at https://rawgit.com/w3c/ttml2/issue-0236-glyph-descendant-misc-build/index.html

PR #556, opened 2018-01-15
Deprecate pixels without specified extent (#330).

### LINKED ISSUE #330: Deprecate use of pixel units unless tts:extent on root element is in pixels.
At present, TTML1 allows a document to use pixel units when expressing ``<length>`` terms, even when the author has not defined the resolution of the document coordinate space by specifying a pixel based ``tts:extent`` on the ``<tt>`` element. I propose that this behavior be deprecated in TTML2, as it leads to potential interoperability problems when presenting TTML content. Note that a number of profiles, including IMSC1 and EBU-TT, require that a pixel based ``tts:extent`` be present when pixel units are used in content.

skynavga:
Closes #330.


nigelmegitt:
Built version at https://rawgit.com/w3c/ttml2/issue-0330-deprecate-pixels-sans-extent-build/index.html

PR #562, opened 2018-01-15
Clarify usages of tts:direction, including relationship to tts:unicodeBidi (#264).

### LINKED ISSUE #264: Link from tts:direction to tts:writing-mode
10.2.12 tts:direction
https://w3c.github.io/ttml2/spec/ttml2.html#style-attribute-direction

> The tts:direction attribute is used to specify a style property that defines the directionality of an embedding or override according to the Unicode bidirectional algorithm.

When i first read this, i thought that `tts:direction` was behaving like `dir` in HTML, or like (i think) `tts:writing-mode=tbrl`, and setting an embedding or override at the same time as signalling the base direction (although i couldn't see how it could do both).

Perhaps it would help to have a note that links to the tts:unicodeBidi section and explains that you need to use both together.

cconcolato:
Closes #264.


cconcolato:
@skynavga is that better?

skynavga:
@cconcolato please review update to PR I just committed

cconcolato:
@skynavga LGTM

PR #565, opened 2018-01-15
Require support for DFXP presentation and transformation profiles (#3…

### LINKED ISSUE #364: Require support for DFXP transformation and presentation profiles.
Add requirement to support DFXP transformation and presentation profiles for TTML1 backward compatibility.

Editorial note from 2014-07-31.

skynavga:
…64).

Closes #364.


skynavga:
Because (1) it must recognize the profile designator, and (2) for avoidance of doubt (in case something about the TTML2 profile superset turns out to contradict the DFXP profile).

palemieux:
> Because (1) it must recognize the profile designator

Yes.

>  (in case something about the TTML2 profile superset turns out to contradict the DFXP profile).

Ok. What it would mean for the processor to support contradicting features in the same namespace? If such a contradiction exists, shouldn't it be noted in the feature specification itself?

skynavga:
On Wed, Jan 17, 2018 at 12:08 PM, Pierre-Anthony Lemieux <

> Because (1) it must recognize the profile designator
>
> Yes.
>
> (in case something about the TTML2 profile superset turns out to
> contradict the DFXP profile).
>
> Ok. What it would mean for the processor to support contradicting features
> in the same namespace? If such a contradiction exists, shouldn't it be
> noted in the feature specification itself?
>
Yes, if we know them. I am trying to anticipate a contradiction we don't
know yet. In any case, there is no harm in calling out independent support
for the DFXP profiles.

palemieux:
>  In any case, there is no harm in calling out independent support for the DFXP profiles.

Weren't we trying to avoid a normative reference to TTML1?

skynavga:
We were, but it is unavoidable given the resolution to mandate support, which means a normative reference.

palemieux:
@skynavga I think it would be easier to assume that DFXP is a subset of TTML2, and avoid a normative reference. Just a suggestion at this point, and I will approve the PR as-is.

skynavga:
When reviewing the matter of TTML1 reference during this process, I had already decided we needed to make the reference normative due the other citations, e.g., in sections 3.3, 5.2.3.1, and 6.2.12.

PR #569, opened 2018-01-18
Align initial values of tts:origin and tts:position (#551).

### LINKED ISSUE #551: Consider aligning initial values of tts:position and tts:origin.
Per discussion during Jan. 2018 F2F, for backwards compatibility with TTML1.

cconcolato:
closes #551
align initial values of tts:origin and tts:position
also remove mention of tt element in tts:position, as discussed
In the process of this last part, I had to move a note in the definition of the root container region.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -1539,7 +1539,7 @@ the applicable writing mode.</p>
 about which see also <specref ref="root-container-region-semantics"/>.</p>
 <note role="elaboration">
 <p>The <loc href="#terms-root-container-region">root container region</loc> has no border or padding; consequently, its border, padding, and
-content rectangles (boxes) are coterminous.</p>
+content rectangles are coterminous.</p>
 </note>
 </def>
 </gitem>
@@ -11570,7 +11570,7 @@ a vertical line located at 50% of the width of the posititoning area.</p>
 </tr>
 </tbody>
 </table>
-<p>If specified on a <loc href="#layout-vocabulary-region"><el>region</el></loc> element, then, the positioning rectangle corresponds with the content rectangle (content box) of the <loc href="#terms-root-container-region">root container region</loc>.</p>
+<p>If specified on a <loc href="#layout-vocabulary-region"><el>region</el></loc> element, then, the positioning rectangle corresponds with the content rectangle of the <loc href="#terms-root-container-region">root container region</loc>.</p>
 <p>If a computed value of the property associated with this attribute is not supported,
 then a <loc href="#terms-presentation-processor">presentation processor</loc> must use the closest supported value.</p>
 <note role="elaboration">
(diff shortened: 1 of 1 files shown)


cconcolato:
As a side point, in my view, the semantics of tts:position could be improved. It does not define normatively how the positioning is done, only through examples. It talks about percentages, what they are relative to, but not how they are used once converted to absolute value. 

Also, the introductory text to the example fragment says that the bottom edge of the region will be 10% above the root container region. That is not correct, it will be the line that is 10% (relative to the region height) above the bottom edge of the region that will be 10% (relative to the root container region) above the bottom of the root container region. 

I can update this PR or raise another issue and prepare another PR. Let me know what you prefer.

spec/ttml2.xml
```diff
@@ -1537,6 +1537,10 @@ the applicable writing mode.</p>
 <p>A unique, logical, rectangular region that establishes a coordinate system into which
 <loc href="#terms-content-region">content regions</loc> are placed and optionally clipped,
 about which see also <specref ref="root-container-region-semantics"/>.</p>
+<note role="elaboration">
+<p>The <loc href="#terms-root-container-region">root container region</loc> has no border or padding; consequently, its border, padding, and
+content rectangles (boxes) are coterminous.</p>
```
skynavga: Please remove "(boxes)". We have previously (years ago) decided to stick with XSL-FO terminology ("area") and not introduce CSS terminology ("box"). In this case, we need neither of these. [I note that this and the following reference to "box" were already present here, which was done in error when incorporating text from CSS. So we can fix this now.]

spec/ttml2.xml
```diff
@@ -11566,22 +11570,7 @@ a vertical line located at 50% of the width of the posititoning area.</p>
...
-<note role="elaboration">
-<p>The <loc href="#terms-root-container-region">root container region</loc> has no border or padding; consequently, its border, padding, and
-content rectangles (boxes) are coterminous.</p>
-</note>
-<p>If a horizontal or vertical position offset is specified by a <att>tts:position</att> attribute in the form of a scalar value
-on a <loc href="#document-structure-vocabulary-tt"><el>tt</el></loc> element, then that value must be expressed using pixel (<code>px</code>) units,
-in which case a pixel must be interpreted as a pixel in the
-<loc href="#terms-presentation-context-coordinate-space">presentation context coordinate space</loc>
-(and not a pixel in the <loc href="#terms-document-coordinate-space">document coordinate space</loc>).</p>
+<p>If specified on a <loc href="#layout-vocabulary-region"><el>region</el></loc> element, then, the positioning rectangle corresponds with the content rectangle (content box) of the <loc href="#terms-root-container-region">root container region</loc>.</p>
```
skynavga: Remove "(content box)".

spec/ttml2.xml
nigelmegitt: I think it would probably be useful to begin introducing CSS terminology, in parenthesis, with a clear indication that it is CSS terminology. For example here, change `(boxes)` to `("boxes" in CSS terminology)`. This is an editorial change only and I don't think we should be concerned about spending a lot of time on it given other work that needs to happen.

spec/ttml2.xml
nigelmegitt: Again, I would be happy with `("content box" in CSS terminology)` as a useful alternative.

PR #570, opened 2018-01-18
Remove animate, begin, dur, end, region and timeContainer of br elements (#552).

### LINKED ISSUE #552: Remove explicit animation, styling and timing from break (br) element.
There appears to be insufficient demand to support explicit animation, styling or timing on the ``<br/>`` element. It is expected that TTML1 3e will recommend against use of animation children, so there is an opportunity to deprecate or obsolete animation children in TTML2, in which case it makes sense to remove (the only recently added - in TTML2) styling and timing as well.

See also [ttml1/#325](https://github.com/w3c/ttml1/issues/325).

cconcolato:
closes #552 

I also checked that `br` is not mentioned in the `Applies to:` section of styling properties and it is not. Quickly looking at the XSD schema it does not look like any change is needed. I am missing anything?

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -16661,7 +16661,6 @@ element types:</p>
 <item><p><loc href="#content-vocabulary-div"><el>div</el></loc></p></item>
 <item><p><loc href="#content-vocabulary-p"><el>p</el></loc></p></item>
 <item><p><loc href="#content-vocabulary-span"><el>span</el></loc></p></item>
-<item><p><loc href="#content-vocabulary-br"><el>br</el></loc></p></item>
 <item><p><loc href="#embedded-content-vocabulary-image"><el>image</el></loc></p></item>
 </ulist>
 <note role="elaboration">
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -5976,8 +5976,10 @@ an <att>xlink:href</att> attribute, and, if it does, then the latter must be ign
 <p>The <el>br</el> element denotes an explicit line break.</p>
 <p>Any metadata specified by children in the <loc href="#element-vocab-group-metadata"><code>Metadata.class</code></loc>
 element group applies semantically to the <el>br</el> element and its descendants as a whole.</p>
-<p><phrase role="deprecated">The use of <code>Animation.class</code> children is deprecated</phrase> since timing vocabulary does not apply to the 
-<el>br</el> element. Instead, animating a <el>br</el> element can be achieved by wrapping it in an animated <el>span</el> element.</p>
+<p><phrase role="deprecated">The use of <code>Animation.class</code> children is deprecated.</phrase></p>
+<note role="explanation">
+<p>No style attributes apply to <el>br</el> elements therefore animation has no effect on <el>br</el>.</p>
+</note>
 <table id="elt-syntax-br" role="syntax">
 <caption>XML Representation &ndash; Element Information Item: br</caption>
 <tbody>
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -5976,8 +5976,8 @@ an <att>xlink:href</att> attribute, and, if it does, then the latter must be ign
 <p>The <el>br</el> element denotes an explicit line break.</p>
 <p>Any metadata specified by children in the <loc href="#element-vocab-group-metadata"><code>Metadata.class</code></loc>
 element group applies semantically to the <el>br</el> element and its descendants as a whole.</p>
-<p>Any animation elements specified by children in the <loc href="#element-vocab-group-animation"><code>Animation.class</code></loc>
-element group apply semantically to the <el>br</el> element.</p>
+<p><phrase role="deprecated">The use of <code>Animation.class</code> children is deprecated</phrase> since timing vocabulary does not apply to the 
+<el>br</el> element. Instead, animating a <el>br</el> element can be achieved by wrapping it in an animated <el>span</el> element.</p>
 <table id="elt-syntax-br" role="syntax">
 <caption>XML Representation &ndash; Element Information Item: br</caption>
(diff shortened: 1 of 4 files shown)


spec/ttml2.xml
```diff
@@ -5976,23 +5976,17 @@ an <att>xlink:href</att> attribute, and, if it does, then the latter must be ign
 <p>The <el>br</el> element denotes an explicit line break.</p>
 <p>Any metadata specified by children in the <loc href="#element-vocab-group-metadata"><code>Metadata.class</code></loc>
 element group applies semantically to the <el>br</el> element and its descendants as a whole.</p>
-<p>Any animation elements specified by children in the <loc href="#element-vocab-group-animation"><code>Animation.class</code></loc>
-element group apply semantically to the <el>br</el> element.</p>
+<p><phrase role="deprecated">The use of <code>Animation.class</code> children is deprecated</phrase> since timing vocabulary does not apply to the 
+<el>br</el> element. Instead, animating a <el>br</el> element can be achieved by wrapping it in an animated <el>span</el> element.</p>
```
skynavga: Please remove the explanatory (rationale) text that follows the word "deprecated" up to the final period ending the paragraph.

skynavga: Only if it goes into a note.

cconcolato: Fine by me. I'll wait a bit to see if other reviewers are fine with that.

cconcolato:
@nigelmegitt @skynavga Is that better?

palemieux:
Build at https://rawgit.com/w3c/ttml2/issue-0552-remove-br-animation-build/index.html

palemieux:
@nigelmegitt @cconcolato See the following text at 11.3.1.4:

_for each br element that is not associated with a tts:display style property with the value none, map the element to a distinct fo:character element having the following properties:_

... so either `tts:display` applies to `br` or that note should be removed.

nigelmegitt:
@palemieux the "applies to" row in the table for `tts:display` reads:

> body, div, image, p, region, span

So _every_ br element is not associated with a `tts:display` style property regardless of its value, apparently. Could be a bug in `tts:display` though.

What say you @skynavga ?

cconcolato:
Regarding 'style', there is also this sentence in a NOTE in 8.1.3:
> Style properties that are associated with a body element in a document instance are available for style inheritance by descendant content elements such as div, p, span and br.

cconcolato:
@nigelmegitt the spec contains plenty of places where it hints that `style` on `br` can be used. It is not trivial to make that change. Please open a separate issue if you want to address that point. 

@nigelmegitt @palemieux @skynavga Added several commits to address the missing parts. Please revise your reviews.

palemieux:
@cconcolato So, it is your opinion that styles in fact apply to `br`?

cconcolato:
I did not look at the point from a technical perspective, only from an editorial one.

nigelmegitt:
I've added the agenda label here because it seems that there are several views to take into account - I hope we can resolve it here before Thursday's call though, since that would be _very_ last minute.

palemieux:
@nigelmegitt Has @skynavga gotten back to you regarding your comment at https://github.com/w3c/ttml2/pull/570#issuecomment-359477550 ? I think it is really key.

skynavga:
@palemieux TTPE uses the following code to process a ``<br/>``

```
    protected void collectBreak(Element e) {
	if (styleCollector.isDisplayed(e)) {
            text.append((char) Characters.UC_LINE_SEPARATOR);
	}
    }
```

so, ``tts:display`` is indeed significant on ``<br\>``

palemieux:
@skynavga How many more styles are expected to apply to `<br>` in your mind? Note that `tts:display` does not apply to `<br>` according to the definition of `tts:display`. Can't we just simply things and remove styles on `<br>`?

nigelmegitt:
> so, `tts:display` is indeed significant on `<br\>`

@skynavga looks like you need to raise an issue against TTPE for that, since it doesn't match the spec.

cconcolato:
@palemieux @skynavga Can you approve the PR?

skynavga:
I now believe this PR is the wrong way to go, and that style, animation, and timing should remain on ``<br/>`` for symmetry, and, further, than it was an error of omission to not indicate that ``tts:display`` applies to ``<br/>``in TTML1.

palemieux:
My input:

- Timing vocabulary was not allowed on `<br>` in TTML1 and does not need to (and should not) be added. This does not mean that timing should not apply to `<br>`. Specifically, `<br>` could have the same implicit duration as `<set>`

- Since timing vocabulary is not allowed on `<br>`, it makes no sense to allow animation.

- `tts;display` could apply to `<br>` in TTML2, but that would require a new feature designator since `tts;display` did not apply to `<br>` in TTML1

- given that the same effect of applying timing and styling vocabulary can be achieved by wrapping `<br>` in span, I suggest we stay the course.

skynavga:
On Wed, Jan 31, 2018 at 1:14 PM, Pierre-Anthony Lemieux <

> My input:
>
>    -
>
>    Timing vocabulary was not allowed on <br> in TTML1 and does not need
>    to (and should not) be added. This does not mean that timing should not
>    apply to <br>. Specifically, <br> could have the same implicit
>    duration as <set>
>    -
>
>    Since timing vocabulary is not allowed on <br>, it makes no sense to
>    allow animation.
>    -
>
>    tts;display could apply to <br> in TTML2, but that would require a new
>    feature designator since tts;display did not apply to <br> in TTML1
>    -
>
>    given that the same effect of applying timing and styling vocabulary
>    can be achieved by wrapping <br> in span, I suggest we stay the course.
>
> My position is that it was in fact allowed in TTML1 (by intention), but
that intention wasn't correctly documented. In other words, it is a error
of omission in TTML1 rather than a positive decision.

Issue #587, opened 2018-01-25
Quote all keywords in value syntax expression definitions.


skynavga:
For example, ``auto`` should be written ``"auto"``.

skynavga:
This change is intended to address the potential for ambiguous reading of identifier tokens, and, as such, is an editorial change since it does not change the intended or actual syntax.

skynavga:
In addition, change any tokens that use ``'`` delimiter to use ``"`` delimiter.

nigelmegitt:
I think this applies only to attribute value syntax, in other words there is no need to modify the XML Representations of _elements_. @skynavga do you agree?

nigelmegitt:
Arguing for modifying XML Representations of elements as well as attribute value syntax, take is example: [§6.1.5 ttp:extension](https://w3c.github.io/ttml2/#profile-vocabulary-extension) has an XML Representation that includes the adjacent lines:

```
  value = (optional|required|use|prohibited) : see prose below
  xml:id = ID
```

I suppose there could be ambiguity about whether `xml:id` should have the value `ID` or be an identifier, given that the `value` attribute takes one of four string keywords that are listed.

As part of this issue it could be modified to:

```
  value = ("optional" | "required" | "use" | "prohibited") : see prose below
  xml:id = ID
```

Doing this across the whole spec will take a while...

nigelmegitt:
I've answered the question for myself: yes, it should include XML Representations of elements. Will push a fix for that in a moment.

PR #588, opened 2018-01-25
Explicitly mark for the mandatory or optional presence of linear whit…

### LINKED ISSUE #315: LWSP in attribute and value expressions?
I apparently missed the three day window on https://github.com/w3c/ttml2/pull/310.

As I understand it, and please correct me if I am wrong, expressions of the form `rgba(        233,    1,0          )` are now permitted.

If so, this will require TTML1 implementations to be revised and tested, without stated benefits as far as I can tell.

skynavga:
…espace (#315).

Closes #315.


nigelmegitt:
Built version at https://rawgit.com/w3c/ttml2/issue-0315-lwsp-rules-build/index.html

skynavga:
@nigelmegitt yes, we can do that, but under #587, not #315

skynavga:
@palemieux I expect to merge this tomorrow or the next day unless I see a change requested review; are you ok with this as it stands?

skynavga:
@palemieux No. See ``{r,g,b,a}-value``. I did it a bit differently here so it wouldn't overflow the column width, but it is equivalent to putting inside the rgb/rgba.

palemieux:
Alright. Was `II` used instead of `<lwsp> | <lwsp>` for reasons other than conciseness?

skynavga:
@palemieux what do you mean? I don't follow

palemieux:
@skynavga I mean that `<lwsp>` is explicit in expressions, but around `||` operators where the `<lwsp>` are built-in the definition of the `||`, right? Same for `&&`, although these are not used anywhere as far as I can tell.

skynavga:
@palemieux ``|`` (call this the _CHOOSE1_ operator) and `||` (call this the _CHOOSEN_ operator) are not equivalent; _CHOOSE1_ admits only one choice, while _CHOOSEN_ admits one or more choices, with ``<lwsp>`` separating multiple choices; none of the _CHOOSE1_ constructs specify ``<lwsp>`` at the outer boundary, i.e., just before or after the choice operator; however, _CHOOSEN_ clearly requires ``<lwsp>`` between options;

skynavga:
@palemieux I will remove reference to the ``&&`` rule, since, as you point out, it is not presently used

skynavga:
@palemieux removed ``&&`` rule

palemieux:
@skynavga Got it. Can an option listed by the `||` operator appear more than once? The definition does not prevent it.

skynavga:
@palemieux I would think that duplication of options should not be permitted; however, I find that neither CSS2.1 nor CSS3 prohibits duplicates; however,  I think we should proscribe duplicates, which I will add shortly.

skynavga:
@palemieux proscribed duplication via ``||``

nigelmegitt:
@skynavga If you're going to make significant changes like this please do not merge without giving folk a chance to look. As it happens in this case I think I'm okay with the changes, but they are not just editorial.

Issue #435, opened 2017-09-29
Is ttp:version="2" required if ttp:contentProfiles is present? 


palemieux:
TTML2 specifies that:

_a ttp:version attribute must be specified on the root tt element of a document instance if it requires support for a feature not defined by [TTML1]._

Is this a processor or document requirement, and what does "require" mean? Specifically, does the mere presence of the `ttp:contentProfiles` attribute in an otherwise TTML1 conformant document trigger the need for `ttp:version="2"`.

I do not think so, but just confirming.

skynavga:
On Fri, Sep 29, 2017 at 10:53 AM, Pierre-Anthony Lemieux <
notifications@github.com> wrote:

> TTML2 specifies that:
>
> *a ttp:version attribute must be specified on the root tt element of a
> document instance if it requires support for a feature not defined by
> [TTML1].*
>
> Is this a processor or document requirement, and what does "require" mean?
> Specifically, does the mere presence of the ttp:contentProfiles attribute
> in an otherwise TTML1 conformant document trigger the need for
> ttp:version="2".
>
A content (document) requirement. Require means the author requires the
implementation to support the TTML2 feature. In other words, if this is
advisory, the content author could drop version, and assume that on an
implementation that doesn't support the feature (here ttp:contentProfiles),
the feature will be ignored; but, if the author wants to require support
(for ttp:contentProfiles) then the author should mark as ttp:version="2"
since it is going to require some version 2 feature.

skynavga:
To elaborate further, if the author expects to require a TTML2 feature to
be supported, e.g., make reference (directly or indirectly) to a processor
profile that will require some TTML2 feature, then the author should
include ttp:version="2"; however, if the author does not expect to require
support for a TTML2 feature by an implementation, then they need not
include ttp:version="2".

From the processor perspective, the presence of ttp:version="2" should be
considered a warning (by the author) that support for some version 2
feature is going to be required (via a processor profile).

palemieux:
So it looks like ttp:version="2" signals a processor requirement. In other words, the author indicates that the processor must understand one or more TTML2 features to be able to process the document accurately. Is that right?

skynavga:
No, that isn't the intent. It is more of an advisory. Any signaling of
required support must be via a processor profile. The ttp:version is more
of a heads up that something is coming.

palemieux:
> No, that isn't the intent. It is more of an advisory. 

The current prose uses conformance language that makes it more than an advisory:

[...]

PR #594, opened 2018-01-28
Specify missing validation semantics (#362, #363, #367, #541).

### LINKED ISSUE #362: Invoke abort if unsupported processor profile.
Add requirement to invoke abort if unsupported processor profile procedure during profile processing.

Editorial note from 2014-05-21.

### LINKED ISSUE #363: Invoke validation processing procedure.
Add requirement to invoke validation processing procedure. Add and use definition of 'validating processor'.

Editorial note from 2014-05-21.

### LINKED ISSUE #367: Add validation processing procedures.
Add validation processing procedures in 5.2.4.2.

Editorial note from 2014-05-21.

### LINKED ISSUE #541: Removal of validation level "strict"?
Section 8.1.1 and 10.2 introduce the additional validation level "strict". It is not clear how a document that fails strict validation relates to validation as defined in section 4. An option may be to remove this validation level completely.

skynavga:
Closes #362.
Closes #363.
Closes #367.
Closes #541.


nigelmegitt:
Built version at https://rawgit.com/w3c/ttml2/issue-0362-0363-validation-build/index.html (note that schemas are also modified by this pull request)

spec/ttml2.xml
```diff
@@ -3034,6 +3040,121 @@ a future version of a TTML presentation or transformation profile.</p>
...
+<glist>
+<gitem id="semantics-procedure-construct-effective-validation-profile">
+<label>[construct effective validation profile]</label>
+<def>
+<olist>
+<item><p>if a document's <loc href="#terms-effective-content-profile">effective content profile</loc> is not <code>null</code>, then the
+<loc href="#terms-effective-validation-profile">effective validation profile</loc> is the
+<loc href="#terms-effective-content-profile">effective content profile</loc>;</p></item>
+<item><p>otherwise, the <loc href="#terms-effective-validation-profile">effective validation profile</loc> is the
+<loc href="#terms-effective-processor-profile">effective processor profile</loc>.</p></item>
```
nigelmegitt: What does it mean to specify that a processor profile is used for validating a document instance?

spec/ttml2.xml
```diff
@@ -3034,6 +3040,121 @@ a future version of a TTML presentation or transformation profile.</p>
...
+</olist>
+</def>
+</gitem>
+<gitem id="semantics-procedure-construct-effective-validation-schema">
+<label>[construct effective validation schema]</label>
+<def>
+<olist>
+<item><p>if a document's <loc href="#terms-effective-validation-profile">effective validation profile</loc> is associated with one or more
+schemas, then select the <loc href="#terms-effective-validation-schema">effective validation schema</loc> from this set based upon the
```
nigelmegitt: This suggests selecting a single validation schema, which to be useful requires that all the schemas are combined into a single schema document. I have seen common practice of applying multiple validation schemas separately when validating a document instance, rather than being required to combine all the provisions of all of them. Can we permit validation against _all_ applicable schema rather than selecting one? Possibly by permitting synthesis of a single schema.

spec/ttml2.xml
```diff
@@ -3034,6 +3040,121 @@ a future version of a TTML presentation or transformation profile.</p>
...
+<loc href="#terms-document-processing-context">document processing context</loc>.</p></item>
+</olist>
+</def>
+</gitem>
+<gitem id="semantics-procedure-validate-document">
+<label>[validate document]</label>
+<def>
+<olist>
+<item><p>if validation is disabled by a <loc href="#terms-higher-level-protocol">higher level protocol</loc>
+or if the computed value of the <att>ttp:validation</att> parameter is <code>prohibited</code>, then exit this procedure;</p></item>
```
nigelmegitt: What is the use case for allowing a document instance to prohibit validation? If I were validating incoming document instances I would always override this. I do not think the option should be available. `"optional"` is as strong as a document instance should be able to specify. At the very least it must be possible for a higher level protocol to override `ttp:validation="prohibited"`.

spec/ttml2.xml
```diff
@@ -21787,19 +22009,32 @@ for the <code>isolate</code> value of the <loc
...
+feature if it:</p>
+<olist>
+  <item>
+    <p>is a <loc href="#terms-validating-content-processor">validating content processor</loc>;</p>
+  </item>
+  <item>
+    <p>recognizes and is capable of transforming the following element vocabulary defined by <specref ref="profile-element-vocabulary"/>:</p>
+    <ulist>
+      <item><p><loc href="#profile-vocabulary-schemas"><el>ttp:schemas</el></loc></p></item>
+      <item><p><loc href="#profile-vocabulary-schema"><el>ttp:schema</el></loc>;</p></item>
```
nigelmegitt: This is too open-ended since it is not qualified by a minimal list of schema `type` values that is supported. I would suggest that we remove the requirement for schema processing from `#validation` and create new feature designators for `#schema-xsd` and `#schema-rnc` that require support for `schemas` and `schema` qualified by the appropriate type.

This is to allow a validating processor to be created specific to a particular profile without defining how that validation must be implemented, and to allow it to validate using a schema without the document instance specifying that schema.

spec/ttml2.xml
skynavga: the same thing as it means when using content profile: find a schema from profile, then validate

spec/ttml2.xml
skynavga: I will tweak to read "select appropriate schema(s)" or something that is equivalent to this

spec/ttml2.xml
skynavga: This feature (prohibited) is not new, so we aren't going to revisit it in this PR.

spec/ttml2.xml
skynavga: No. We aren't prescribing image types, so we don't need to do so with schema types. The only reason to support the new ttp:schema elements is to support validation, so it needs to be tied to the #validation feature.

spec/ttml2.xml
nigelmegitt: That doesn't work - the spec talks about inferring a processor profile from a content profile in lots of places, but never the other way around.

spec/ttml2.xml
nigelmegitt: None of the issues linked from this pull request appears to concern the addition of schema validation so in the absence of evidence that we have consensus to add this feature it needs to be removed from this pull request.

spec/ttml2.xml
nigelmegitt: The previous text included:

> unless the end-user or application overrides this prohibition

which appears to have been removed. That needs to be added back in here in some form.

spec/ttml2.xml
nigelmegitt: Thank you.

spec/ttml2.xml
skynavga: you seem to assume this is trying to infer a content profile, but that is not the case; it merely says if there is no content profile, then use processor profile to find an associated schema (since the new ttp:schema{s} elements can be put into a processor profile as well, or associated by the processing context);

the issue here is how to perform validation when no content profile is specified, e.g., on TTML1 documents; TTV, e.g., does this by making use of a built-in (code) association of particular schemas with the TTML1 profiles;

spec/ttml2.xml
skynavga: ok, can do

spec/ttml2.xml
skynavga: which feature you referring to? the #validation feature was not added in this PR

skynavga: Just to be clear, requiring support for the ``ttp:schema{s}`` element types does not imply support for any particular schema validation whatsoever; it just means you support the association of schema(s) with a profile.

nigelmegitt: > which feature you referring to? 

@skynavga `schema` and `schemas`, which appear to be introduced in this pull request, but for which I can see no proposal or request in the linked issues.

spec/ttml2.xml
nigelmegitt: See comments attached to later lines in the file; I do not see any agreement from the WG to add `ttp:schema{s}`. The closest I can find based on an issue search is https://github.com/w3c/ttml2/issues/435#issuecomment-356142844 which does not mention adding new syntax (and anyway that issue appears not to be the target of this pull request). In my view this is a request for a new feature, and we have certainly agreed not to do that, so it should be raised as an issue against ttml.next.

To resolve this comment though, I think we need somewhat loose words that recommend that the processor infer any validation rules such as schemas based on the processor profile, without specifying an algorithm for doing so.

xchange11: +1

skynavga:
@palemieux this PR addresses three long standing open issues (#362, #363, and #367) as well as a more recently posted issue (#541); without this PR, the semantics of ttp:validate and ttp:validateAction are not well defined, as there was no defined procedure for [validate document]. As part of satisfying this, at least one standard mechanism is needed to satisfy "is or can be associated with one of the abstract document types defined by 4 Document Types" as found in Section 3.1, item (3). This is the purpose of the schemas/schema element children of ttp:profile.

palemieux:
@skynavga What is the use case for `ttp:validation`? Validation is not something that should be under the authors control.

spec/ttml2.xml
skynavga: @nigelmegitt I am not following you. Let me explain my premises, which perhaps will allow some clarity:
1. to validate, you need to a schema (or an equivalent) to be associated with the document (at least one such mechanism is mandated in 3.1 (3); however, we did not define a standard mechanism in TTML1);
2. we support multiple schema languages, so we can't simply choose the XSD schema association mechanism (``xsi:schemaLocation``, etc.), that is, we need something more generic;
3. we already have a (rich) mechanism for associating one or two profiles with a document, either just a process profile or both a content profile and a processor profile;
4. therefore, it makes sense to associate (one or more) schemas with a profile in order to satisfy (1) and (2) above (indeed, this is what TTV does internally, but without the benefit of an author defined association mechanism - that is, TTV has a hardwided set of associations from a "model" to a set of profiles and sets of schemas for these profiles);
5. an author may specify content profile for a document, which should take priority over determining schemas, in which case schemas associated with the effective processor profile for such a document are ignored;
6. if an author does not specify content profile it is null, in which case the effective processor profile may supply such schema associations;

If we do not specify something like the proposed ``ttp:schema`` element, then we are left with no standardized association mechanism whatsoever, which, in my mind, renders the validation functionality of TTML2 into horribly underspecified state such that TTML2 cannot claim to support validation at all.

Now, I don't have any problem with allowing an implementation to look elsewhere for schema associations (i.e., to look beyond any author specified schemas explicitly defined using ``ttp:schema``), but I have a serious problem with not providing at least one standard way to formally establish an association.

Also, I see no problem of logic or processing with respect to using associations obtained from a process profile as a backup for when an author does not specify a content profile.

skynavga:
@palemieux I don't know how to respond to such a claim. Of course validation is and should be under authorial control. What is the point of specifying a content profile? The validation features defined in TTML2 are and have been one of its key features from the very first stages of designing and specifying TTML2.

palemieux:
>  Of course validation is and should be under authorial control. What is the point of specifying a content profile?

The point of specifying a content profile is let other entities validate the document. The author does not need to provide any additional information. In other words, I do not see the need for `ttp:validate`. Can you provide a concrete workflow/use case? Surely you have thought of one?

skynavga:
@palemieux The only "change" here is
1. Defining a [validate document] procedure; and,
2. Defining a formal way to associate schemas with documents (via a profile).

This is not new, and it is assumed from the semantics of ``ttp:validate`` and ``ttp:validateAction`` and has long been documented. As you very well know, a parameter must be normatively used in some processing to be called a parameter. This PR formally provides that use.

However, perhaps you mistake my meaning of "control" here. Here, I mean having the ability to express authorial requirements or at least desires with respect to validation. Of course, the ultimate control over whether validation is performed and what actions to take is the domain of the implementation and the end user, so the latter can clearly override the author. However, the author needs to be permitted to provide guidance to this process, thus the ``ttp:validate`` and ``ttp:validateAction`` parameters.

skynavga:
@palemieux first, to correct a point, we should be talking of ``ttp:validation`` and ``ttp:validationAction`` (I used the wrong names above); next, in usage contexts where content validity is important or essential, then an author may wish to insist that a document only be processed if it is determined to be valid by the receiving entity, in which case, the author would

1. specify a processor profile that requires support for ``ttp:validation``;
2. specify ``ttp:validation='required'``;

Step (1) ensures that the document is only processed by a processor that is a validating processor, and (2) ensures that validation is actually performed. If an implementation does not support validation, then (1) would cause abort processing. If an implementation does support validation, but it is turned off, then (2) would cause abort processing. If an implementation does support validation and it is turned on, then (2) would cause validation to take place, in which case ``ttp:validationAction`` determines the action taken when a validation error or warning occurs.

This is already implemented, so there is no risk for CR1.

skynavga:
I would also remind folks that we already define and or make reference to the following going back to the first WD of TTML2 (12 February 2015) [1] (and earlier [3]):

* _performing validation_ with respect to **effective content profile**
* **validating content processor** (marked as TBD)
* _validating processor_ with respect to deprecation and obsoletion processing
* _perform validation_ under **[construct effective content profile]**
* ``ttp:validation``
* ``ttp:validationAction``
* many other instances which I won't attempt to enumerate here

The current (new) TTML2 profiles and validation features can mostly be traced back to [2], which were first implemented in TTML2 in May 2014 [3].

The only thing new in this PR that isn't part of this earlier work is the addition of the ``ttp:schema[s]`` elements, which are important to realize effective semantics for satisfying  the following editor's notes added one week after posting [3]:

screen shot 2018-02-10 at 3 37 59 pm

screen shot 2018-02-10 at 3 38 25 pm

[1] https://www.w3.org/TR/2015/WD-ttml2-20150212/
[2] https://dvcs.w3.org/hg/ttml/raw-file/default/ttml2/design/TPAC2013-TTMLProfiles.pdf
[3] https://lists.w3.org/Archives/Public/public-tt/2014May/0111.html

palemieux:
> the addition of the ttp:schema[s] elements

Schemas are not sufficient to validate the document, so why even include them.

spec/ttml2.xml
skynavga: @nigelmegitt although my preference is to require support for ``ttp:schema(s)`` via ``#validation``, I am willing to move this support to an independent ``#validation-schemas`` feature designator, which, if we take this option, that leaves only implied (implementation dependent) association between the effective validation profile and any potential schemas in the absence of support ``#validation-schemas``; it is also my intention that we add these elements to our standard profiles and point at the RNC and XSD schemas we define;

skynavga:
@nigelmegitt @palemieux I have attempted to address all comments; for diff from the original PR, see https://github.com/w3c/ttml2/pull/594/commits/5c8b5278cf050b550913513b601261d7ee2389e2

spec/ttml2.xml
nigelmegitt: > to validate, you need to a schema (or an equivalent) to be associated with the document

@skynavga This is false. A validation processor does not require a schema.

> it makes sense to associate (one or more) schemas with a profile in order to satisfy (1) and (2) above

Even if this is true, the group has not discussed it and no case has been made that there is a pressing need for it to be added now.

> TTML2 cannot claim to support validation at all

I disagree with this - if it were true then it would equally have been true for TTML1. Yet there are validation processors for TTML1 documents and leaving the implementation mechanism for validation to the implementer has worked fine so far.

> I am willing to move this support to an independent `#validation-schemas` feature designator,

No, sorry, this is a new feature and it has to be deferred. We are too late to have this in TTML2 now. Giving the feature a designator does not mitigate the fact that there is no consensus to add the feature.

nigelmegitt:
@skynavga Sorry, no, you have not addressed the point that we cannot introduce `schema` and `schemas` at this time, and that needs:

1. An issue
2. Consensus to proceed

Such an issue must, at this time, be targeted at ttml.next and so must any pull request that introduces the feature. We do not have time to go ahead with it in TTML2 now.

This pull request needs to be modified to revert the addition of `schema` and `schemas`. We're just too late for this one. This is a process point. The alternative is that we defer all of the issues addressed in this pull request because there's no acceptable pull request that addresses them.

skynavga:
We disagree on three essential points:

   - on whether a schema is required, I contend it is, but I am using a
   perhaps broader definition of schema than you; in particular, anything
   (whatsoever) that embodies the constraints specified by the normative XML
   representation syntax in TTML can be construed as a schema, including
   hard-wired code;
   - on whether this feature (associating schemas with a document for
   performing validation) is either useful or required; I hold that it is;
   - on whether this feature is already implied by the validation semantics
   placeholder that has been in the spec for 3+ years;
   - on whether this feature produces a risk for CR1: I contend it does
   not, but we can mark it at risk to make it easy to remove if we aren't
   satisfied at the end of CR1;
   - on whether it is appropriate to define at this juncture: I clearly
   hold that it is;

You should focus the following:

   - is it a unacceptable risk to include in CR1 now, and, if you argue
   that it is, then indicate why an at-risk statement does not successfully
   mitigate any process risk;

skynavga:
@tairt that this change is substantive is not a sufficient cause to exclude from CR; many other PRs are marked as substantive as well; we have a well accepted process to address removal of a feature, and that is identifying it as at risk, and then dropping it at the end of CR when it is judged to be unfixable, unimplemented, etc

skynavga:
@nigelmegitt @palemieux @tairt @cconcolato please see [1] for my argument that this PR is covered by an already approved resolution by the WG

[1] https://lists.w3.org/Archives/Public/public-tt/2018Feb/0142.html

nigelmegitt:
[Meeting 2018-02-15] @skynavga agrees to remove `ttp:schema` and `ttp:schemas` within the next day.

skynavga:
@nigelmegitt @palemieux please review now that ttp:{schemas,schema} are removed

skynavga:
@cconcolato please review

spec/ttml2.xml
```diff
@@ -21787,19 +21920,25 @@ for the <code>isolate</code> value of the <loc
...
-<item><p><loc href="#profile-attribute-validation"><att>ttp:validation</att></loc></p></item>
-<item><p><loc href="#profile-attribute-validationAction"><att>ttp:validationAction</att></loc></p></item>
-</ulist>
+feature if it:</p>
+<olist>
+  <item>
+    <p>is a <loc href="#terms-validating-content-processor">validating content processor</loc>;</p>
+  </item>
+  <item>
+    <p>recognizes and is cable of transforming the following attribute vocabulary defined by <specref ref="profile-attribute-vocabulary"/>:</p>
```
cconcolato: typo: cable -> capable

palemieux:
Is it accurate that _[validate document]_ only depends on schema validation, i.e. does not include any semantic validation not included in the schema validation?

nigelmegitt:
This hasn't built properly. Going to do some fiddling.

skynavga:
No. But I can see you one might read that in. I am using schemas in
[construct effective profile schemas] to mean not only formal schema
definitions, e.g., XSD and RNC, but also "programmatic schemas" that encode
other semantic constraints. I will add a note making this clear.

nigelmegitt:
@skynavga I'm seeing a note to that effect at line 995 already, but not in the built version. I'm going to close and reopen to try to trigger a build.

palemieux:
@skynavga Validation should be against the entire specification. I do not understand why validation needs to be tied to schemas, and then expand to term schema to mean the entire specification. Anyway, I guess this could be fixed post-CR1.

spec/ttml2.xml
```diff
@@ -3034,6 +3044,127 @@ a future version of a TTML presentation or transformation profile.</p>
 </div4>
 </div3>
 </div2> <!-- vocabulary-profiling -->
+<div2 id="vocabulary-validation">
+<head>Validation</head>
+<p>This section describes the TTML validation sub-system and high level requirements that apply to this sub-system.</p>
+<p>Validation is an optional process that may occur when a candidate <loc href="#terms-document-instance">document instance</loc> is
+consumed or otherwise processed by a <loc href="#terms-validating-content-processor">validating content processor</loc>. If such processing
+takes place and an exceptional condition occurs, then processing may be aborted, as described in further detail below.</p>
+<p>If processing is aborted during <loc href="#terms-validating-content-processor">validating content processor</loc>, then the possible side effects
```
nigelmegitt: grammar issue: "If processing is aborted during validating content processor, ..." doesn't make sense. Should this say "If processing is aborted during validation, ..."?

spec/ttml2.xml
```diff
@@ -3034,6 +3044,127 @@ a future version of a TTML presentation or transformation profile.</p>
...
+include:</p>
+<ulist>
+<item><p>only part of a non-conformant document may be transformed by a <loc href="#terms-transformation-processor">transformation processor</loc>,
+resulting in an incomplete transformation;</p></item>
+<item><p>only part of a non-conformant document may be presented by a <loc href="#terms-presentation-processor">presentation processor</loc>,
+resulting in an incomplete presentation.</p></item>
+</ulist>
+<p>If abort processing does occur as a side effect of validation processing, then the location in
```
nigelmegitt: Why "side effect"? Surely it is a direct effect. Suggest removing "side" here., or even replacing "side effect" with "consequence" or "result".

spec/ttml2.xml
```diff
@@ -3034,6 +3044,127 @@ a future version of a TTML presentation or transformation profile.</p>
...
+<glist>
+<gitem id="semantics-validation-state-validation-exception">
+<label>[validation exception]</label>
+<def>
+<p>A <loc href="#semantics-validation-state-validation-exception"><phrase role="strong">validation exception</phrase></loc>
+represents the following internal state information that corresponds to an exception condition that occurs during validation processing</p>
+<glist>
+<gitem id="semantics-validation-state-validation-exception-message">
+<label>message</label>
+<def><p>An arbitrary (implementation defined) string the describes or characterizes the exception</p></def>
```
nigelmegitt: s/the/that

spec/ttml2.xml
```diff
@@ -3034,6 +3044,127 @@ a future version of a TTML presentation or transformation profile.</p>
...
+<item><p>if the computed value of the <loc href="#profile-attribute-validation"><att>ttp:validation</att></loc> parameter is <code>optional</code> and the
+<loc href="#terms-document-processing-context">document processing context</loc> determines that optional validation is not to be performed,
+then exit this procedure;</p></item>
+<item><p>otherwise, using the <loc href="#terms-effective-validation-schemas">effective validation schemas</loc>, validate document conformance
+such that</p>
+<ulist>
+<item><p>if a <termref def="defs-validation-error">validation error</termref> condition is detected, then perform the semantics that apply
+to the computed value of the <loc href="#profile-attribute-validationAction"><att>ttp:validationAction</att></loc> parameter property;</p></item>
+<item><p>if a <termref def="defs-validation-warning">validation warning</termref> condition is detected, then perform the semantics that apply
+to the computed value of the <loc href="#profile-attribute-validationAction"><att>ttp:validationAction</att></loc> parameter property.</p></item>
```
nigelmegitt: There are two mechanisms present by which a document instance can attempt to wriggle out of validation: firstly by setting `ttp:validation` to `prohibited` and secondly by setting `ttp:validationAction` to `warn` or `ignore`. Both of these options have no place in a document instance. If a system wishes to validate a document instance then it is up to the operators of that system what behaviour should apply, not up to the author of the document instance being provided to it.

@skynavga can you give any use case where this would be helpful? I

skynavga:
You are making objections to functionality that has been in the
specification for nearly four years, has been implemented, and is well
defined (since you have not shown that it is somehow technically broken as
defined, whether or not you think that functionality should be present).

We have discussed the prohibited semantics in the past and obtained
consensus. You are not providing any new information now that would cause
us to remove it.

I will be objecting to going to CR unless we leave this functionality in
place,  so if you want to delay CR1 by insisting on removing now, then a
delay will ensue.

I suggest you hold off and allow CR1 to occur as planned (and without
removing this already agreed upon functionality). We now know that a CR2
will be required, so there is further time and decision points to revisit
this without delaying CR1.

spec/ttml2.xml
skynavga: Fixing now.

spec/ttml2.xml
skynavga: Fixing now.

spec/ttml2.xml
skynavga: Fixing now.

spec/ttml2.xml
skynavga: Actually, setting ``ttp:validationAction`` to ``warn`` or ``ignore`` does not prevent validation. The ``ttp:validationAction`` property is used only for the purpose of determining how to process a violation, not whether to detect a validation. As for having an author set ``ttp:validation`` to prohibited, we have already specified that the end-user or application may override this prohibition, so the prohibition from the author is merely advisory in an absolute sense. However, there is nothing wrong with an author saying that validation should not occur if it is there interest to do so. We cannot dictate their intended  use with respect to a document.

skynavga: @nigelmegitt a use case for an author specifying ``ttp:validation='prohibited'`` may include any of the following reasons:
* the content is of low value and the author doesn't care if invalidity causes it to fail; e.g., the author may not wish to take the time to validate the content in the first place and is willing to let the client do whatever it can with it;
* the content is known to be invalid (for whatever reason and the author would like a validating client to take no action with respect to validation;

Isn't this the model most used on the Internet today in general?

We have defined the default for ``ttp:validation`` to be ``optional``, and we have stated that the application (client) or end user can override ``ttp:validation='prohibited'``; therefore, it is truly up to the client to decide whether to validate or not. In that sense,  you can distinguish between optional and prohibited as nothing more than a hint on the part of the author as to what they prefer. However, the semantics of this are well defined as far as the stated mechanisms are concerned: you haven't shown they are not implementable!

skynavga:
I have reworked the earlier attempt to address this comment. Please check
and approve if satisfied.

spec/ttml2.xml
```diff
@@ -4663,11 +4811,15 @@ ttp:validationAction
 </tbody>
 </table>
 <p>If this parameter's value is <code>abort</code>, then, a <loc href="#terms-validating-content-processor">validating content processor</loc>
-must abort processing of a TTML <loc href="#terms-document-instance">document instance</loc> when validation processing fails.</p>
+must abort processing of a TTML <loc href="#terms-document-instance">document instance</loc> when a
+<termref def="defs-validation-error">validation error</termref> exception occurs.</p>
```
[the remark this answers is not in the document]
nigelmegitt: This comment raised as #672.

spec/ttml2.xml
```diff
@@ -2056,20 +2065,26 @@ instantiating a <loc href="#terms-reduced-xml-infoset">reduced xml infoset</loc>
...
+<item><p>the processor does not support some required
 (mandatory) feature specified or implied by a TTML profile
-declared to apply to the <loc href="#terms-document-instance">document instance</loc>.</p>
+declared to apply to the <loc href="#terms-document-instance">document instance</loc>; or</p></item>
+<item><p>a <termref def="defs-validation-error">validation error</termref> occurs when performing the
+<loc href="#semantics-procedure-validate-document"><phrase role="strong">[validate document]</phrase></loc> procedure.</p></item>
+</olist>
 </item>
 <item>
 <p>The processor supports all mandatory processing semantics defined by this specification.</p>
```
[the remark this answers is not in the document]
nigelmegitt: This raised as #673.

spec/ttml2.xml
```diff
@@ -1788,8 +1808,8 @@ and found to be valid as defined by <specref ref="doctypes"/>.</p>
 <gitem id="terms-validating-content-processor">
 <label>[validating content processor]</label>
 <def>
-  <p>A <loc href="#terms-content-processor">content processor</loc> which performs validation processing on a
-  TTML <loc href="#terms-document-instance">document instance</loc> prior to performing any other type of processing.</p>
+  <p>A <loc href="#terms-content-processor">content processor</loc> which does or can perform validation of a
+  candidate <loc href="#terms-document-instance">document instance</loc> prior to performing certain other type of processing.</p>
```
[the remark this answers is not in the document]
nigelmegitt: This raised as #674.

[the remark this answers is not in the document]
skynavga: If validation occurs then it should be performed in a context where any content that might be validated is not otherwise used until validation is completed on that content. If an implementation wishes to interleave validation and post-validation content processing, that is an implementation decision. I will add an informative note to this end via #674.

spec/ttml2.xml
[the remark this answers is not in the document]
nigelmegitt: Right, my comment wasn't worded well. The point is that `#region-timing` and `#time-offset` are mandatory features for generic processors but are likely to be specified as prohibited by some content profiles. In that sense the line is too strict - it _requires_ processor support for something that in some cases is not in fact needed. 

One possible resolution is to make those features optional rather than mandatory. 

A second possible resolution is to say that generic conformance requires support for the intersection of a) those features specified as mandatory for generic processors and b) those features specified as required by the (possibly inferred) processor profile.

A third possible resolution would be to exclude mandatory features from those that may be prohibited by a content profile.

PR #595, opened 2018-01-29
Remove support for unitless lengths (#561).

### LINKED ISSUE #561: Do not default to px units in the absence of units on a <length> expression.
TTML2 differs from TTML1 in the treatment of `<length>` units with no unit. TTML1 says:

>It is an error to omit the units component of a scalar length value.

TTML2 currently says:

> If no _units_ component is specified in a `<length>` expression, then it is to be treated as if `px` (pixels) units were specified.

However, we are in general attempting to converge with CSS, and in CSS some properties treat lengths without units (which are not `<length>` but are `<number>`) as multipliers to be evaluated on the element to which they are being applied. For example this is used in the [line-height](https://www.w3.org/TR/2011/REC-CSS2-20110607/visudet.html#line-height) property.

By creating a behaviour for when no units component is specified in a `<length>` expression it becomes indistinguishable from `<number>` and therefore prevents our future potential adoption of this CSS approach. I propose to revert this change in TTML2 in favour of the TTML1 approach so that we are able to use unitless numbers in style attributes that take lengths at some point in the future.

Further motivation for this reversion comes from #330 - the resolution to that issue means that omitting a unit generates the pixel behaviour which now means there should be a `tts:extent` on the `tt` root element, however it is much harder to test for since the document may have no explicit mention of `px` at all and still generate a warning/error condition if that attribute is absent and a unitless length is used.

skynavga:
Closes #561.


spec/ttml2.xml
```diff
@@ -14698,8 +14698,6 @@ units
 </tr>
 </tbody>
 </table>
-<p>If no <emph>units</emph> component is specified in a &lt;length&gt; expression, then it is to be treated as
-if <code>px</code> (pixels) units were specified.</p>
```
nigelmegitt: Please could you reinstate the TTML1 SE text:

> It is an error to omit the units component of a scalar length value.

?

nigelmegitt:
I just noticed that `tts:disparity` specifies a default of `0` with no units, and its type is `<length>`, so this pull request needs to modify that default value too.

spec/ttml2.xml

skynavga:
Addressed all comments. @nigelmegitt please approve

PR #603, opened 2018-01-29
Introduce anonymous span creation procedure (#358).

### LINKED ISSUE #358: Incorporate resolutions of additional TTML1 Issues.
This is an umbrella issue to cover the tasks of incorporating resolutions of additional TTML1 issues into TTML2.

- [x] [w3c/ttml1/#191](https://github.com/w3c/ttml1/issues/191) drop mode forumla
- [x] [w3c/ttml1/#192](https://github.com/w3c/ttml1/issues/192) frame rate multiplier defaulting
- [x] [w3c/ttml1/#193](https://github.com/w3c/ttml1/issues/193) implicit duration span
- [x] [w3c/ttml1/#196](https://github.com/w3c/ttml1/issues/196) attribute forward compatibility
- [x] [w3c/ttml1/#197](https://github.com/w3c/ttml1/issues/197) pruning br erroneously
- [x] [w3c/ttml1/#201](https://github.com/w3c/ttml1/issues/201) agent and actor references
- [x] [w3c/ttml1/#204](https://github.com/w3c/ttml1/issues/204) generic family name
- [x] [w3c/ttml1/#205](https://github.com/w3c/ttml1/issues/205) clarify meaning of percentage
- [x] [w3c/ttml1/#207](https://github.com/w3c/ttml1/issues/207) visibility hidden
- [x] [w3c/ttml1/#212](https://github.com/w3c/ttml1/issues/212) content bpd
- [x] [w3c/ttml1/#214](https://github.com/w3c/ttml1/issues/214) font size using 'em' unit
- [x] [w3c/ttml1/#235](https://github.com/w3c/ttml1/issues/235) tab processing
- [x] [w3c/ttml1/#239](https://github.com/w3c/ttml1/issues/239) region overflow semantics
- [x] [w3c/ttml1/#241](https://github.com/w3c/ttml1/issues/241) z index auto
- [x] [w3c/ttml1/#251](https://github.com/w3c/ttml1/issues/251) not yet defined TT attributes
- [x] [w3c/ttml1/#266](https://github.com/w3c/ttml1/issues/266) epoch
- [x] [w3c/ttml1/#275](https://github.com/w3c/ttml1/issues/275) chained referential styling
- [x] [w3c/ttml1/#304](https://github.com/w3c/ttml1/issues/304) smooth scroll preference
- [x] [w3c/ttml1/#310](https://github.com/w3c/ttml1/issues/310) implicit durations of set
- [x] w3c/ttml1#316 fix typo
- [x] [w3c/ttml1/#320](https://github.com/w3c/ttml1/issues/320) overflow example

### LINKED ISSUE #634: Specify implicit duration of <br>.
Even though no timing vocabulary applies to the `<br/>` element (see https://github.com/w3c/ttml1/issues/310). timing semantics, including an implicit duration, needs to be defined for the  `<br/>` element, otherwise the  `<br/>` element  is never displayed in the following example:

```
<p>
<span begin="1s"><br/></span>
</p>
```

whereas "hello" is displayed in the following example:

```
<p>
<span begin="1s">hello</span>
</p>
```

Suggest specifying that `<br/>` follow the same semantics as anonymous spans since `<br/>` is expected to behave like an EOL.

See also https://github.com/w3c/ttml1/issues/339

cconcolato:
Partially addresses #358.
Closes #634.


spec/ttml2.xml
```diff
@@ -17590,16 +17602,17 @@ i.e., parallel containment always applies.</p>
 </issue>
 <ulist>
 <item>
-<p>The implicit duration of an anonymous span is defined as follows: if
-the anonymous span's parent time container is a parallel time container, then
+<p>The implicit duration of an anonymous span, a <el>set</el> element, or a <el>span</el> element whose children are exclusively text nodes, 
+is defined as follows: 
+if the parent time container is a parallel time container, then
```
nigelmegitt: I see there's a problem with the equivalent TTML1 text here: it says "if the _span's_ parent time container ..." but the scope of this has been changed to it also covers anonymous spans and set elements. I think this here is better.

cconcolato: @nigelmegitt I raised that issue on TTML1 https://github.com/w3c/ttml1/issues/335 and @palemieux has a PR.

spec/ttml2.xml
```diff
@@ -17629,6 +17642,39 @@ based timing or event based timing in <bibref ref="smil3"/>, where the specific
 <specref ref="time-expression-semantics-smpte"/> apply.</p>
 </item>
 </ulist>
+<p>For the purposes of performing presentation processing, the <phrase role="strong">
```
skynavga: This is wrong.

skynavga: This is wrong.

spec/ttml2.xml
```diff
@@ -17590,16 +17602,17 @@ i.e., parallel containment always applies.</p>
...
 the implicit duration is equivalent to the <code>indefinite</code>
-duration value as defined by <bibref ref="smil3"/>; if the anonymous span's
+duration value as defined by <bibref ref="smil3"/>; if the 
 parent time container is a sequential time container, then the implicit duration
 is equivalent to zero.</p>
 </item>
 <item>
-<p>The implicit duration of a <el>body</el>, <el>div</el>, <el>p</el>,
-<el>span</el>, <el>audio</el> or <el>image</el> element is determined in accordance to (1) whether
+<p>The implicit duration of a <el>body</el>, <el>div</el>, <el>p</el>, <el>audio</el> or <el>image</el> element, or a <el>span</el> whose children
```
skynavga: Reverting.

spec/ttml2.xml
```diff
@@ -17590,16 +17602,17 @@ i.e., parallel containment always applies.</p>
...
 <ulist>
 <item>
-<p>The implicit duration of an anonymous span is defined as follows: if
-the anonymous span's parent time container is a parallel time container, then
+<p>The implicit duration of an anonymous span, a <el>set</el> element, or a <el>span</el> element whose children are exclusively text nodes, 
+is defined as follows: 
+if the parent time container is a parallel time container, then
 the implicit duration is equivalent to the <code>indefinite</code>
-duration value as defined by <bibref ref="smil3"/>; if the anonymous span's
+duration value as defined by <bibref ref="smil3"/>; if the
```
skynavga: Reverting.

spec/ttml2.xml
```diff
@@ -17590,16 +17602,17 @@ i.e., parallel containment always applies.</p>
 </issue>
 <ulist>
 <item>
-<p>The implicit duration of an anonymous span is defined as follows: if
-the anonymous span's parent time container is a parallel time container, then
+<p>The implicit duration of an anonymous span, a <el>set</el> element, or a <el>span</el> element whose children are exclusively text nodes,
```
skynavga: Reverting.

spec/ttml2.xml
```diff
@@ -16697,6 +16697,32 @@ attribute are implied. In addition, a <el>layout</el> container element is impli
 the implied <el>region</el> element.</p>
 </note>
 </div4>
+<div4 id="anonymous-span-construction">
```
skynavga: Moving this material closer to reference, and undoing bad changes.

spec/ttml2.xml
```diff
@@ -17629,6 +17642,39 @@ based timing or event based timing in <bibref ref="smil3"/>, where the specific
 <specref ref="time-expression-semantics-smpte"/> apply.</p>
 </item>
 </ulist>
+<p>For the purposes of performing presentation processing, the <phrase role="strong">
+<loc href="#procedure-construct-anonymous-spans">[construct anonymous spans]</loc></phrase> procedure is applied before resolving begin and end 
+times of content elements.</p>
+<p>The following example illustrates the timing semantics of anonymous spans that are children of sequential and parallel time containers.
```
skynavga: Putting this into new sub-section that is clearly non-normative.

skynavga:
The original PR is trying to make too many changes of different types, the substantive ones being of a nature that we do not have agreement (and I believe were semantically wrong). This now a purely editorial PR. We can do other PRs to address specific substantive changes, such as defining implicit duration for ``<set>``.

spec/ttml2.xml
```diff
@@ -17023,10 +17021,19 @@ original sequence of anonymous spans;</p></item>
 replace the anonymous span with its sequence of child text
 nodes;</p></item>
 </olist>
-</item>
+</def>
+</gitem>
+<gitem id="procedure-perform-flow-transformation">
+<label>[perform flow transformation]</label>
```
nigelmegitt: This label is not used anywhere - the process is never called by anything, so giving it a name, while not a problem, is also not a solution.

spec/ttml2.xml
palemieux: Why?

spec/ttml2.xml
palemieux: Why?

palemieux:
At this point, this PR should match the changes made and approved in TTML1. Both specifications can be fixed post CR1 if needs be.

skynavga:
@palemieux unfortunately, it cannot, since the changes in TTML are factually wrong (anonymous span creation does not occur before ISD construction)

nigelmegitt:
> anonymous span creation does not occur before ISD construction

@skynavga we already agreed that it must do, on the basis that knowledge of anonymous span timing is a requirement for ISD construction.

skynavga:
@nigelmegitt @palemieux the problem is the following language, which contradicts the language in TTML1 2e (9.33) and TTML2 ED (11.3.1.4) where anonymous span construction occurs as the initial step of the flow transformation process, which occurs AFTER ISD construction, which occurs AFTER timing resolution.

```
+<p>For the purposes of performing presentation processing, anonymous spans are created
+according to the following <phrase role="strong">[construct anonymous spans]</phrase> 
+procedure:</p>
``

**Comment by nigelmegitt** [comment]
@skynavga why can anonymous span creation not be done _both_ before _and_ after ISD construction, i.e. before to establish which elements are in the ISD and after for flow transformation?

**Comment by skynavga** [comment]
@nigelmegitt it COULD be done before, but only if we don't use the phrase _anonymous span_, i.e., if we refer to an implicit duration in terms of text fragments that will eventually be mapped to an anonymous span

**Comment by skynavga** [comment]
@nigelmegitt it looks like the problem language was introduced in https://github.com/w3c/ttml1/pull/318, which I did not review

**Comment by cconcolato** [comment]
@skynavga  I don't think there is a problem since TTML2 also says:
> Note:
> A compliant presentation processor is permitted to perform the previous step (anonymous span generation) and this step (style resolution) as post-processing steps while performing the [construct intermediate document] procedure, in which case each intermediate synchronic document produced by 11.3.1.3 Intermediate Synchronic Document Construction will contain anonymously generated spans and computed style sets, and therefore, these steps would not be performed during synchronic flow processing.

**Comment by palemieux** [comment]
Need to apply https://github.com/w3c/ttml1/pull/340/files

File: spec/ttml2.xml (hunk 8 above)
Comment (skynavga) [reply spec/ttml2.xml]: The lack of a label is an inconsistency, since we have added labels elsewhere. Also, it is needed in this particular context to separate the new construct anonymous spans procedure from the flow transformation procedure.

Comment (palemieux) [reply spec/ttml2.xml]: @skynavga Regardless of the language, please make sure the prose in TTML2 yields that same semantics as in https://rawgit.com/w3c/ttml1/issue-339-br-implicit-duration-build/index.html

**Comment by skynavga** [comment]
@nigelmegitt @palemieux i have completed my updates for this PR, which ends up with a few changes from the original PR:

* animate needed same treatment as set;
* animate and set need to use implicit duration of nearest time container ancestor;
* audio and image need to work like br and anonymous span;
* removing attempt to resolve https://github.com/w3c/ttml1/issues/193, which doesn't appear to represent the consensus expressed there, namely, to take no action at this time; and, as it turns out, the attempt to deal with this in https://github.com/w3c/ttml1/commit/62ca2217c48915a45ee2e7efc318d2e95dfe4677 doesn't quite work, for multiple reasons, one being that text fragments may have been converted to anonymous spans already, another being that span may specify ``@timeContainer``;

**Comment by palemieux** [comment]
@skynavga _[construct anonymous spans]_ must be performed before _11.3.1.3 Intermediate Synchronic Document Construction_ since determining whether text node is temporally active requires wrapping it in an anonymous span, i.e. text nodes do not have timing semantics.

**Comment by palemieux** [comment]
@skynavga The implicit duration of spans that only contain text nodes must be specified to be equivalent to that of anonymous spans (time container semantics do not apply since text nodes do not have timing semantics)

**Comment by palemieux** [comment]
@skynavga The implicit duration of a `set` element should be set to that of anonymous span, as in TTML1, unless there is a reason not to.

**Comment by skynavga** [comment]
@palemieux no (it need not), no (we did not agree to take action on https://github.com/w3c/ttml1/issues/193), and no (what you added in ttml1 is not correct)

**Comment by skynavga** [comment]
I have a feeling we will have to push PR out to CR2.

**Comment by palemieux** [comment]
> no (it need not), no (we did not agree to take action on w3c/ttml1#193), and no (what you added in ttml1 for set is not correct)

You have not indicated why it is not correct. I am waiting.

**Comment by skynavga** [comment]
By not correct, I am referring to the implicit duration of set. What does the following mean?

```
<span timeContainer='seq'>
  <set begin='1s' end='3s' tts:color='red'>
  <span dur='2s'>one</span>
  <span dur='3s'>two</span>
</span>
```

I think it means that the color of the outer span is set to red from 1 to 3s in the syncbase of the outer span, which means that the implicit duration of set should follow the implicit duration of the outer span.

**Comment by palemieux** [comment]
@skynavga In the example at https://github.com/w3c/ttml2/pull/603#issuecomment-366161023, the desired begin of `<span dur='2s'>one</span>` will be 3s according to TTML1 since:

- the explicit end of `set` is 3s into the `seq` container
- the implicit (and desired) begin of `<span dur='2s'>one</span>` is the desired end of the previous sibling.

As a result, the color will not change.

I am not saying this is idea and/or desirable, but it is what TTML1 says.

To achieve the result you describe, the author would simply do the following:

```
<span>
<set begin='1s' end='3s' tts:color='red'>
<span timeContainer='seq'>
  <span dur='2s'>one</span>
  <span dur='3s'>two</span>
</span>
</span>
```

or 

```
<span>
  <set begin='1s' end='3s' tts:color='red'>
  <span begin='0s' end='2s'>one</span>
  <span begin='2s' end='5s'>two</span>
</span>
```

Test cases below, which can be experimented with at http://sandflow.com/imsc1proc/index.html

[set-duration-examples.zip](https://github.com/w3c/ttml2/files/1732734/set-duration-examples.zip)

PR #604, opened 2018-01-30
Remove auto value for rubyAlign (#249).

### LINKED ISSUE #249: NR is less than NB and NB is equal to 1
10.2.34 tts:rubyAlign
https://www.w3.org/TR/2016/WD-ttml2-20161117/#style-attribute-rubyAlign

> otherwise, if NR is less than NB, the semantics of spaceAround or spaceBetween apply, respectively, according to whether NB is less than or equal to one (1) or is greater than one;

if NR is less than NB and NB is equal to 1, then NR has to be 0, in which case i don't see why the respectively difference is applied.  What am i missing?

cconcolato:
closes #249


cconcolato:
@skynavga I thought we had agreed, but indeed I can't find the minutes. This is related to [Netflix's whitepaper](https://lists.w3.org/Archives/Public/public-tt/2017Nov/att-0007/Japanese_Subtitles_on_the_Netflix_Service.pdf) in which we indicated that 'center' should be the default value.

@nigelmegitt can we formally agree on that?

nigelmegitt:
@cconcolato I do not recall agreeing to remove `auto` - I suggest you add a comment to the issue to highlight this proposed approach. 

Nevertheless, it does appear to be a possible resolution to the issue. If there is consensus on merging this pull request as a mechanism for resolving the issue then I think we should accept that.

PR #605, opened 2018-01-30
Add note for wide ruby annotations (#250).

### LINKED ISSUE #250: rubyAlign and wide ruby text
10.2.34 tts:rubyAlign
https://www.w3.org/TR/2016/WD-ttml2-20161117/#style-attribute-rubyAlign

> The tts:rubyAlign attribute is used to specify the position of ruby text within the inline area generated by the ruby text container annotation.

What about the (common) case where the ruby text is longer than the base text?  Note that in CSS a property such as ruby-align applies the distribution of text to the smaller of the ruby text _or_ base text, rather than just to the ruby text, ie. it operates on the shorter of the two.

cconcolato:
closes #250


spec/ttml2.xml
```diff
@@ -11819,6 +11819,13 @@ have no impact on ruby alignment.</p>
 </note>
 <p>If a computed value of the property associated with this attribute is not supported,
 then a <loc href="#terms-presentation-processor">presentation processor</loc> must use the value <code>auto</code>.</p>
+<note role="elaboration">
+<p>In this specification, ruby annotations in subtitling events are assumed to be narrow, so as to ease their reading. As a consequence,
```
nigelmegitt: I'm not happy to reference subtitling events in TTML2. That's a specific usage and should be in IMSC. Rather, we should switch "subtitling events" for "timed text applications" and "so as to ease their reading" for "since presentation space and available time are often constrained" (happy for tweaks to be made to the precise wording)

cconcolato:
@nigelmegitt I've tweaked the wording. How does it read now? Can you approve?

nigelmegitt:
Thanks @cconcolato I don't have any problem with that text now, but there's no point in merging it unless it also helps with the issue - so @r12a please could you take a look at this and see if it addresses your issue?

I note that this specific proposal for a note was not mentioned in the issue. In fact the last comment, from June, is suggesting a change to the semantics of rubyAlign, rather than a note.

nigelmegitt:
Closing and reopening to trigger travis.

skynavga:
@nigelmegitt I propose we merge the (now updated) note as an adequate response to the initial comment

Issue #546, opened 2018-01-10
Remove root style inheritance functionality.


skynavga:

skynavga:
Based on discussion in today's F2F, agreed to defer root style inheritance functionality to future work. Anyone who objects to this removal should speak up. A PR that implements this change is forthcoming.

PR #606, opened 2018-01-30
Remove ttp:version parameter attribute (#435).

### LINKED ISSUE #435: Is ttp:version="2" required if ttp:contentProfiles is present?

skynavga:
Closes #435.


spec/ttml2.xml
```diff
@@ -13006,9 +12920,6 @@ example.</p>
 <p>The <att>tts:textShadow</att> attribute is used to specify a style property that
 defines one or more text shadow decorations to apply to glyphs that are selected for <loc href="#terms-glyph-area">glyph areas</loc> generated
 by content flowed into a region.</p>
-<p>If both <loc href="#style-attribute-textOutline"><att>tts:textOutline</att></loc> and <att>tts:textShadow</att> attributes
```
cconcolato: why is this related to the version attribute?

cconcolato:
Built version at https://rawgit.com/w3c/ttml2/issue-0435-remove-version-parameter-build/index.html

cconcolato:
@skynavga, just to make sure we have covered every aspect, I'm copying the results of your study of TTT and checking how this change affects them: 
- [ttv] select between version specific default profiles
This seems covered by the use of TTML2 profiles as the default given that it is a superset of TTML1.

- [ttv] select between TTML schemas associated with different specification versions (56 instances)
Same as above. At some point, we had mentioned that a validator could choose the schema corresponding to the minimal profile that includes all required features.  I think this is not needed for now.

- [ttv] contextualize length expressions in terms of version (in order to support new units) (2 instances)
Can you clarify how this is handled? 

- [ttv] enable deprecation warnings (1 instance)
I think this is covered by the selection of a profile.

- [ttpe] select use of initial value of tts:position versus tts:origin (1 instance)
We fixed this with #551 

- [ttx] select between tt element and region element for root of style inheritance (1 instance)
This is covered by #546 

- [ttx] select between version specific transformation helper class methods used during ISD transformation processing (24 instances)
This seems internal to your software.

spec/ttml2.xml
skynavga: We had previously thought that ``tts:textShadow`` (a TTML2 feature) would replace ``tts:textOutline`` (a TTML1 feature). We now understand it isn't a replacement (driven by version) but both apply independently.

skynavga:
@cconcolato thanks for checking my TTT study; there are a few things I need to add/tweak to cover these points

skynavga:
@nigelmegitt I reviewed @version on ISD vocabulary, and decided it was appropriate there, since we don't profile the ISD vocabulary usage via features from which we might infer the ISD version.

nigelmegitt:
@skynavga OK, I'm not sure about that but no objection from me to keep it in and moving on.

spec/ttml2.xml
cconcolato: I don't see how this was driven by version and how this is related to `ttp:version` but I think the change is good. Approving

PR #607, opened 2018-01-30
Refactor support for foreign namespace attributes and elements (#439).

### LINKED ISSUE #439: Foreign namespace elements not reflected in syntax.
### Section 5.3.2 

> In addition to standardized extension vocabulary, a conforming document instance may contain arbitrary namespace qualified elements that reside in any namespace other than those namespaces defined for use with this specification.

This option seems not reflected in the content model description of TTML elements.

skynavga:
Closes #439.

Marking as editorial since it has no impact on conformance.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -18420,7 +18420,7 @@ attributes by expressing them indirectly on <el>metadata</el>
 elements.</p>
 </note>
 <note role="clarification"><p>The presence of foreign namespace attributes in a concrete encoding of a TTML
-<loc href="#terms-document-instance">document instance</loc> do not affect the formal validity of that
+<loc href="#terms-document-instance">document instance</loc> does not affect the formal validity of that
 <loc href="#terms-document-instance">document instance</loc> with respect to the applicable
 <loc href="#terms-abstract-document-type">abstract document type</loc> selected by <specref ref="doctypes"/>.</p></note>
 <p>The use of foreign element metadata is illustrated by the following example.</p>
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -18441,7 +18419,10 @@ however, in this case the author wishes to segregate certain metadata
 attributes by expressing them indirectly on <el>metadata</el>
 elements.</p>
 </note>
-<p></p>
+<note role="clarification"><p>The presence of foreign namespace attributes in a concrete encoding of a TTML
+<loc href="#terms-document-instance">document instance</loc> do not affect the formal validity of that
```
nigelmegitt: This is a little awkward grammatically, since the "presence" is a singular thing so I'm expecting "does not affect" rather than "do not affect". Suggest rewording to:

_Foreign namespace attributes that may be present in a concrete encoding of a TTML document instance do not affect..._

spec/ttml2.xml
```diff
@@ -18468,7 +18449,10 @@ elements.</p>
 <p>In the above example, a number of elements defined by the Dublin
 Core metadata vocabulary are used to express document level metadata.</p>
 </note>
-<p></p>
+<note role="clarification"><p>The presence of foreign namespace elements in a concrete encoding of a TTML
```
nigelmegitt: Similar point to that above - suggest: 

_Foreign namespace elements that may be present in a concrete encoding..._

skynavga:
@nigelmegitt addressed your comments, please approve

PR #608, opened 2018-01-30
Improve animation value syntax (#383).

### LINKED ISSUE #383: Improve animation value syntax.
Enhance detail of syntax of ``<animation-value>`` in order to prevent appearance of an unquoted or unescaped semicolon, which is used as a delimiter between animation values in ``<animation-value-list>``.

Editorial note from 2014-07-31.

skynavga:
Closes #383.


spec/ttml2.xml
```diff
@@ -18080,28 +18080,25 @@ or ending (final) of the attribute targeted by the animation.</p>
 <td>
 <eg xml:space="preserve">
 &lt;animation-value&gt;
-  : string
+  : ([^;] | escape)+
```
nigelmegitt: Is this a regexp? Do we use those anywhere else? Does it imply that _any_ code point other than `;` can be used? If so, it seems to admit use of characters that would otherwise fall within `<lwsp>` which makes processing `<animation-value-list>` rather tricky. I'm not convinced we can use a shortcut like this, sadly.

skynavga: This is standard syntax. Nothing new, and not invented by TTWG. Look at quoted string for example.

nigelmegitt: Oh no! It's broken there too. From the references I've just read about character classes, the example here means "any character not a `;`" and the example in quoted string (`[^"\\]` and similar for a single quote) both admit whitespace and control characters and are not restricted at all otherwise (apart from not permitting the quote characters that need to be escaped). `<animation-value>` even permits unescaped attribute delimiters, which would surely cause a validation error, or some other problems.

Since the `<animation-value-list>` can be used on any styling attribute, the value type for each `<animation-value>` in the list needs to be unambiguous in the context of the permitted values for the styling attribute on which it is specified. Styles like `tts:fontFamily` permit any quoted string, and look like they provide an escaping mechanism but that restriction needs to apply to `<animation-value>` too - it seems to be the most liberal style attribute that we need to handle.

skynavga: I don't see any breakage here at all. You are correct that the syntactic expression ``[;\\]`` means "any character other than ``;`` and ``\\``. You are also correct that ``[^"\\]`` means "any character other than ``"`` and ``\\``. This is by design, is part of existing implementations, and has resulted in no bug report that I know of.

The only syntactic requirement of ``<animation-value>`` from the perspective of parsing ``<animation-value-list>`` is that the former contain no unescaped ``;``.  In other words, the parsing of ``<animation-value-list>`` consists solely of splitting a string where ``;`` not preceded by ``\\`` is the separator; the result of this split operation produces an array of strings, each of which must satisfy the syntax rules of the associated style property.

nigelmegitt: @skynavga Right, so `<animation-value>` is used in attributes, but can contain XML attribute delimiters and control character codes. How is that not broken?

skynavga: Because XML attribute processing occurs before value parsing. If it is not well formed as an XML attribute in the first place, it is already invalid and will never reach value parsing.

PR #611, opened 2018-01-30
Loosen coupling between usesForced and forced bound parameter (#609).

### LINKED ISSUE #609: Loosen coupling between item name="usesForced" and forced.
The "usesForced" named metadata item definition is very tightly coupled with use of `condition` in the document instance, in relation to the `forced` bound parameter. This means it cannot be used in other contexts that may define some forced functionality. See also the thread leading up to https://github.com/w3c/imsc/issues/311#issuecomment-361318848

Additionally, only one scope of usage is defined, at the document level. It would seem a reasonable optional usage to be able to scope it to an element in the body content such as `div` etc.

Since no processing semantic is based on the metadata item, there we can be less prescriptive about its usage. This will also have the effect of reducing the implementation and test burden.

Instead of the current wording:

> A boolean that expresses whether some `<condition>` expression makes use of the `forced` bound parameter, where the value adheres to `xsd:boolean`. If this named metadata item is present in a document instance, then it must be specified as a child of the `head` element.

I propose:

> A boolean that expresses whether some content within the specifying element's scope makes use of forced display semantics, where the value adheres to `xsd:boolean`. If this named metadata item is specified as a child of the `head` element then the applicable scope is the whole document; if specified as a child of a content element, then the applicable scope is that element and its descendants.
> Note:
> Forced display semantics can be expressed by use of the `forced` bound parameter within a `<condition>` expression.

nigelmegitt:
Closes #609 by changing the definition of the usesForced named metadata item in the way proposed there.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -18912,11 +18912,14 @@ provide a textual equivalent or summary for some related image content.</p>
 <gitem id="named-item-usesForced">
 <label><code>usesForced</code></label>
 <def>
-<p>A boolean that expresses whether some <loc href="#content-value-condition">&lt;condition&gt;</loc> expression
-makes use of the <loc href="#bound-parameter-forced"><code>forced</code></loc> bound parameter, where the value
-adheres to <loc href="http://www.w3.org/TR/xmlschema-2/#boolean">xsd:boolean</loc>. If this named metadata item
-is present in a <loc href="#terms-document-instance">document instance</loc>, then it must be specified as a child
-of the <loc href="#document-structure-vocabulary-head"><el>head</el></loc> element.</p>
+<p>A boolean that expresses whether some content within the specifying element's scope makes use of forced display 
+semantics, where the value adheres to <loc href="http://www.w3.org/TR/xmlschema-2/#boolean">xsd:boolean</loc>. 
+If this named metadata item is specified as a child of the <loc href="#document-structure-vocabulary-head"><el>head</el></loc>
+element then the applicable scope is the whole document; if specified as a child of a 
+<loc href="#terms-content-element">content element</loc>, then the applicable scope is that element and its descendants.</p>
+<note role="elaboration">
+<p>Forced display semantics can be expressed by use of the <loc href="#bound-parameter-forced"><code>forced</code></loc>
+bound parameter within a <loc href="#content-value-condition">&lt;condition&gt;</loc> expression.</p></note>
 </def>
 </gitem>
 </glist>


spec/ttml2.xml
```diff
@@ -18912,11 +18912,14 @@ provide a textual equivalent or summary for some related image content.</p>
 <gitem id="named-item-usesForced">
 <label><code>usesForced</code></label>
 <def>
-<p>A boolean that expresses whether some <loc href="#content-value-condition">&lt;condition&gt;</loc> expression
-makes use of the <loc href="#bound-parameter-forced"><code>forced</code></loc> bound parameter, where the value
-adheres to <loc href="http://www.w3.org/TR/xmlschema-2/#boolean">xsd:boolean</loc>. If this named metadata item
-is present in a <loc href="#terms-document-instance">document instance</loc>, then it must be specified as a child
-of the <loc href="#document-structure-vocabulary-head"><el>head</el></loc> element.</p>
+<p>A boolean that expresses whether some content within the specifying element's scope makes use of forced display
```
skynavga: Remove all scoping semantics that are not document scoped.

nigelmegitt: I believe this change already meets the agreed relaxation of the "must be specified as a child of the `head` element" to express only significance - see https://github.com/w3c/ttml2/issues/609#issuecomment-361719364

Rather than expressing a boolean significance based on the location, it instead relates the position to the scope of significance.

nigelmegitt:
> What use case would require usesForced to be specified anywhere but `<tt>`?

@palemieux An author may wish to indicate specific content elements for example `<div>`s that contain forced content, perhaps because they have been introduced or modified as part of the authoring process and require some additional quality control, for example.

The counter-case is also relevant here. Rather than leaving a possible syntactic option open to differences in interpretation which would not be interoperable, I would prefer to state a baseline interpretation now, which would be open to review.

palemieux:
> An author may wish to indicate specific content elements for example <div>s that contain forced content, perhaps because they have
> been introduced or modified as part of the authoring process and require some additional quality control, for example.

Isnt't this true about any change? Should there be a `usesStyle` named metadata item? I really fail to see an actual use case for `usesForced` on anything but `head`. This is of course from the perspective of IMSC1.1. I do not object to adding it in TTML2 for potential future use, as long as it is prohibited in IMSC1.1.

nigelmegitt:
> Isn't this true about any change?

Maybe, and `ttm:item` is extensible to cover that.

> Should there be a `usesStyle` named metadata item?

Nobody has asked for one so far, so no.

> I really fail to see an actual use case for `usesForced` on anything but `head`. This is of course from the perspective of IMSC1.1. I do not object to adding it in TTML2 for potential future use, as long as it is prohibited in IMSC1.1.

I'm not going accept a restriction on metadata in IMSC 1.1 just because you don't see the use case for it @palemieux , when it has no impact on processing. It just isn't dangerous enough to prohibit.

However I'm not really clear about the use case for the `usesForced` metadata item being defined in TTML2 _at all_, regardless of scope. Can you or @skynavga or anyone else here recall why it was added? Could we simply remove it and resolve this discussion with "less is more"?

palemieux:
> Could we simply remove it and resolve this discussion with "less is more"?

Yes!

nigelmegitt:
Closing in favour of #621.

PR #614, opened 2018-01-31
Further constrain ruby descendants (#372).

### LINKED ISSUE #372: Constrain certain descendant content in ruby container.
Proscribe the following from appearing in a ruby container:

* use of span children of container that do not specify a ruby style, i.e., that do not play a role in ruby structure; if present, then such span is ignored for presentation purposes;
* use of #PCDATA children of container, baseContainer, and textContainer; i.e., permit #PCDATAA only in base, text, and delimiter; if used, then ignore for presentaion purposes;
* use of br element in ruby or its descendants, and, if used, treat as normal (collapsible) whitespace;
* use of U+2028 (LINE SEPARATOR) and U+2029 (PARAGRAPH SEPARATOR) in ruby or its descendants, and, if used, they are interpreted as normal (collapsible) whitespace.

Editorial notes from 2015-02-18.

skynavga:
Closes #372.


skynavga:
@cconcolato no, can't do that (change to "must"); we try to use constraint language consistently in TTML2, which uses declarative statements with "is" not "must"; see also the text under Conventions:

>If normative specification language takes an imperative form, then it is to be treated as if the term must applies. Furthermore, if normative language takes a declarative form, and this language is governed by must, then it is also to be treated as if the term must applies.
>
>Note:
>
>For example, the phrases "treat X as an error" and "consider X as an error" are to be read as mandatory requirements in the context of use. Similarly, if the specification prose is "X must apply", "X applies", or "X is mandatory", and "X" is further defined as "X is Y and Z", then, by transitive closure, this last declarative phrase is to be read as "Y is mandatory" and "Z is mandatory" in the context of use.

Doing as you suggest would require making thousands of changes throughout the text, which clearly we are not going to do.

cconcolato:
It is fine in many places to use the "imperative" form. I note that this is not the case everywhere, e.g. the use of 'must' in the following sentences:
> When an image appears in an image defining context the following additional constraints apply:
- attributes animate, begin, dur, end, region, and timeContainer must not be specified, and, if they appear, must be ignored for the purpose of presentation processing;
- elements from the Animation.class element group must not appear as child elements, and, if they appear, must be ignored for the purpose of presentation processing.

I feel it would be more appropriate in this case to use 'must'.

cconcolato:
About the conciseness, can you at least avoid repeating "if the computed value of tts:ruby is container" 5 times and others 6 or 7 times, e.g. by using sub-bullets?

skynavga:
On Wed, Jan 31, 2018 at 1:59 PM, Cyril Concolato <notifications@github.com>
wrote:

> About the conciseness, can you at least avoid repeating "if the computed
> value of tts:ruby is container" 5 times and others 6 or 7 times, e.g. by
> using sub-bullets?
>
That is strictly editorial tweaking, which we can do at any time in the
future. We have more important matters to work at present.  Please approve
the PR unless you find a substantive breakage.

skynavga:
@cconcolato to elaborate, the constraints on ruby content are mandatory for the support of tts:ruby, because:

>&**3.1.5 (5)**
>The reduced xml infoset satisfies all additional mandatory syntactic and semantic constraints defined by this specification.
>
>**3.2.1 (5)**
>If the processor supports some optional processing semantics defined by this specification, then it does so in a manner consistent with the defined semantics.
>
>**10.2.35**
>When using tts:ruby, the following nesting constraints apply:
> ...

I realize you may be accustomed to other conventions in different SDOs or WGs or documents, but this is what we have with TTML, and we aren't going to change it now.

As for making minor, editorial tweaks to improve reading: while they may be considered useful for some readers, others will find what we have sufficient. At this juncture, I prefer we focus on substantive issues, and not word-smithing.

skynavga:
@cconcolato thanks!

Issue #378, opened 2017-06-05
Merge informative media timing material into Appendix I.


skynavga:
Generalize and rename Appendix I *Time Expression Semantics* to Root Container Region Semantics. Merge informative material from 12.4.1 *Media Timing* into Appendix I, updating this material as needed.

Editorial note from 2016-03-13.

nigelmegitt:
I would prefer to move the appendix into section 12 rather than the other way around, since it is both normative and useful.

nigelmegitt:
I think this move has been accepted in principle in TTML1, and now just needs to be implemented here. It's editorial only.

nigelmegitt:
Pull request #616 moves the media timing informative section to the appendix, but does not generalise or rename appendix I as proposed in this issue.

Issue #550, opened 2018-01-10
Media time relation assumption is incomplete.


nigelmegitt:
The TTML1 text (reintroduced at https://github.com/w3c/ttml2/pull/536/files#diff-5ac39ab39265717d36dfd789d254ce1aR24385) states an assumption about the begin time in media timebase relating to the begin of the related media object. However this is incomplete.

> The above formalisms assumes that the Root Temporal Extent corresponds with the beginning of a related media object. If this assumption doesn't hold, then an additional offset that accounts for the difference may be introduced when computing media time M.

The full assumption is that the origin coordinates match, for example:

_The above formalisms assume that the zero point of the time coordinates corresponds with the zero point of the time coordinates of the related media object. If this assumption doesn't hold, then an additional offset that accounts for the difference may be introduced by the Document Processing Context when computing media time M._

I have also clarified above that the difference is out of scope of the document instance.

skynavga:
Merger https://github.com/w3c/ttml2/pull/616 resolves closes this issue.

skynavga:
Marking as a substantive (in the weak sense) as it may be construed to resolve (or change) possibly ambiguous semantics around the notion of a document's temporal extent and its relation to some other possible external time lines.

PR #616, opened 2018-01-31
Make begin times relative to the document temporal coordinate space (#512)

### LINKED ISSUE #512: Begin times being relative to the root temporal extent can be wrong.
The term "root temporal extent" is used in several places in TTML2 as a reference point for the beginning of the document. However the root temporal extent itself is defined as:

> The temporal extent (interval) defined by the temporal beginning and ending of a document instance in relationship with some external application or presentation context.

This seems problematic to me. I think it would be clearer to relate unstated begin times, of e.g. `<body>` elements and `isd:isd` documents, to the applicable origin for the document's timebase. This frees up the root temporal extent also to begin at a time related to that same origin, and makes more sense especially in the context of §12.4.1 Media Timing Semantics, which goes into further detail about one consequence of the above definition, that the root temporal extent may be truncated at the beginning by the external application or presentation context. I would not expect the times in `isd:isd` documents for example to require offsetting to accommodate such a truncation - rather, they would remain the same relative to a fixed origin.

nigelmegitt:
Closes #512 and #550 and partly addresses #378. Resolves the ambiguity of whether the root temporal extent begins at the same time as the document's temporal coordinate space and moves the media timing informative section to appendix I.

Clarifies that body element and isd:isd begin and end times are relative to that coordinate space, rather than the root temporal extent, whose begin time is not necessarily the origin of the document's temporal coordinate space.


palemieux:
Built at https://rawgit.com/w3c/ttml2/issue-0512-begin-times-relative-to-doc-time-build/index.html

palemieux:
Couple of questions:

> The root temporal extent, i.e., the time interval over which a document instance is active, has an implicit duration that is equal to the implicit duration of the body element of the document, if the body element is present, or zero, if the body element is absent.

Does this mean that the implicit duration of a document with no `body` is 0 even if some regions have `tts:showBackground="always"`? That does not seem right.

 > then the begin point is interpreted as the later of the beginning point of the document temporal coordinate space and the beginning point of the root temporal extent.

How does one compute the _the beginning point of the root temporal extent_?

nigelmegitt:
> Does this mean that the implicit duration of a document with no `body` is 0 even if some regions have `tts:showBackground="always"`? That does not seem right.

@palemieux It may do, however that is not a change introduced in this pull request - you might want to open a separate issue about that. It looks like this text doesn't take into account the rehoming of `body` under `region`.

> How does one compute the _the beginning point of the root temporal extent_?

The only help the spec gives you (in the current ED or in this pull request) is "in relationship with some external application or presentation context" so I think that's defined out of band. The reason I added the "later of" construction is because a document with no `begin` attributes in practice begins whenever the presentation processor chooses to begin playing it, which is either at the document temporal origin or some time later, but by definition cannot be earlier.

nigelmegitt:
I realised this pull request also addresses #550, and updated the issue to match.

nigelmegitt:
Brought in the changes needed also to close #378.

PR #617, opened 2018-01-31
Enable inline ruby presentation based on document processing context (#255).

### LINKED ISSUE #255: How would someone display ruby inline?
10.2.39 tts:rubyPosition
https://www.w3.org/TR/2016/WD-ttml2-20161117/#style-attribute-rubyPosition

Is there a way to make the ruby display inline when desired?

Note that there are two possibilities here. Given a word such as 東京, the user may want to produce

a. 東(とう)京(きよ), or 
b. 東京(とうきよ)

from the same markup. The second approach tends to be used for compoun nouns in Japanese.

cconcolato:
closes #255


spec/ttml2.xml
```diff
@@ -11553,13 +11557,7 @@ computed value of <att>tts:ruby</att> of exactly one sibling is <code>delimiter<
 <item><p>if the computed value of <att>tts:ruby</att> is <code>delimiter</code>, then the
 computed value of <att>tts:ruby</att> of no descendant element is not <code>none</code>;</p></item>
 </ulist>
-<p>A validating processor must treat a violation of any of the above constraints as an error.
-For the purpose of presentation processing, the violation of any of these constraints should result in fallback (inline)
-presentation of ruby text annotations.</p>
```
nigelmegitt: @cconcolato Why remove this recommendation for fallback in the case of the constraints being validated? It seems like it would still be worth including that statement.

cconcolato: Thanks for pointing that out. I wasn't sure about it. I checked the rest of the spec and we don't tell implementation what to do in case of error. Given that we now have a clear way for implementation to do inline I did not see the need to specify that. It encourages the use of invalid content in general because it will work in implementations not supporting full presentation.

skynavga: @cconcolato actually we specify fallback behavior in many error cases, and I think it important we do it more (not less); I don't think it encourages the use of invalid content at all, but creates more consistency among implementations

cconcolato:
@nigelmegitt @skynavga Added the sentence back. Can you review again?

PR #619, opened 2018-01-31
Clarify use of tts:textCombine (#278).

### LINKED ISSUE #278: tts:textCombine and combining characters
10.2.43 tts:textCombine
http://w3c.github.io/ttml2/spec/ttml2.html#style-attribute-textCombine

I and other members of the i18n WG found it hard to tell that this is about tate chu yoko, rather than kumimoji, until we get to the information near the bottom about the digits and all values.  It would be good to make that clearer near the top.

> This attribute has no impact on or interaction with the nominal layout of glyph areas that constitute a Unicode combining character sequence.

I'm not sure what this means. Does it mean that in vertical text where `tts:textCombine:all` is applied to the non-Japanese letters in the sequence 

お  ‎304A  HIRAGANA LETTER O
茶  ‎8336  [CJK Unified Ideographs]
は  ‎306F  HIRAGANA LETTER HA
t  ‎0074  LATIN SMALL LETTER T
e  ‎0065  LATIN SMALL LETTER E
́  ‎0301  COMBINING ACUTE ACCENT
で  ‎3067  HIRAGANA LETTER DE
す  ‎3059  HIRAGANA LETTER SU

the acute accent would be part of the horizontal text or not?

cconcolato:
closes #278


spec/ttml2.xml
```diff
@@ -12445,8 +12445,8 @@ follows the semantics of the <code>text-align-last</code> property with a value
 <div3 id="style-attribute-textCombine">
 <head>tts:textCombine</head>
 <p>When a vertical writing mode applies, the <att>tts:textCombine</att> attribute is used to specify a style property that
-determines whether and how multiple nominally non-combining characters are combined so that their <loc href="#terms-glyph-area">glyph areas</loc>
-consume the nominal bounding box of a single em square of the surrounding vertical text.
+determines whether and how multiple characters, such as digits, are combined so that their 
+<loc href="#terms-glyph-area">glyph areas</loc> consume a smaller bounding box, potentially the nominal bounding box of a single em square of the surrounding vertical text.
```
cconcolato: In fact, looking at the matching [CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/text-combine-upright). That property also says "the combination of multiple characters into the space of a single character". I was confused by the example rendering below where 4 characters (AB34) are not collapsed to fit in one em box. I will revert that text and update the example.

PR #621, opened 2018-02-01
Remove usesForced named metadata item

### LINKED ISSUE #609: Loosen coupling between item name="usesForced" and forced.

palemieux:
Closes #609 

See https://github.com/w3c/ttml2/pull/611#issuecomment-362032101 and https://github.com/w3c/ttml2/pull/611#issuecomment-362031664

Moots PR #611

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -6128,9 +6128,7 @@ argument-list
 <label><code>forced</code></label>
 <def>
 <p>Evaluates to a boolean value that denotes whether the <loc href="#terms-content-processor">content processor</loc> is operating with
-<loc href="#terms-forced-subtitle">forced subtitles</loc> enabled.
-If used in a <loc href="#terms-document-instance">document instance</loc>, then a <loc href="#named-item-usesForced"><code>usesForced</code></loc>
-named metadata item should be specified as a child of the <loc href="#document-structure-vocabulary-head"><el>head</el></loc> element.</p>
+<loc href="#terms-forced-subtitle">forced subtitles</loc> enabled.</p>
 </def>
 </gitem>
 <gitem id="bound-parameter-mediaAspectRatio">
@@ -18893,7 +18891,6 @@ a <loc href="#metadata-vocabulary-item"><el>ttm:item</el></loc> element.</p>
 <eg xml:space="preserve">
 &lt;named-item&gt;
   : altText
-  | usesForced
 </eg>
 </td>
 </tr>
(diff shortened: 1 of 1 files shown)


cconcolato:
Built version: http://rawgit.com/w3c/ttml2/issue-0609-remove-usesForced-build/index.html

palemieux:
@skynavga Actually three members of the WG do not believe it serves a use case... so which use case does it serve? Moreover, as far as I can tell, `usesForced` was introduced in https://github.com/w3c/ttml2/commit/4f9fa5cab861c19cb5f1ec63374c85e86f9933e5 , which has no issue tied to it, so I doubt the group discussed it at all.

nigelmegitt:
#### Use case

@skynavga What real use case does it serve? 

Just in case this was mentioned in the requirements document for TTML at https://www.w3.org/TR/2006/NOTE-ttaf1-req-20060427/ I checked this, and it is absent there. (For clarity, I do not believe we are bound by that document in its entirety, and can potentially introduce use cases not considered in that document, but it's fairly comprehensive and has stood the test of time reasonably well; I checked it because if the need for a usesForced metadata item had somehow been represented in that document then the case for keeping it would have been clear.)

#### History of usesForced discussion in the TTWG

A search reveals that after it was introduced initially without discussion, we talked about it very briefly at the TPAC F2F in Lisbon ([minutes](https://lists.w3.org/Archives/Public/public-tt/2016Sep/0040.html)) where the group agreed to keep usesForced on the basis that it is referenced elsewhere in the specification. Even at that time there was evidently a request to remove it.

#### Impact of removal

However, on reviewing this dependency, it seems like it is somewhat arbitrary and the usesForced metadata item name can certainly be removed straightforwardly: there's a reference to it in the `forced` bound parameter that would need to be removed, but there's no further onward impact than that.

nigelmegitt:
Apologies for the mess here in terms of comments - the proper place for the discussion of removal of usesForced is in the issue, so I've copied my above comment to https://github.com/w3c/ttml2/issues/609#issuecomment-362541539; in the meantime the current count is that three members of the WG  are happy to remove it, and one is not.

nigelmegitt:
The issue has been closed, so closing this pull request.

PR #623, opened 2018-02-02
Mapping between ruby base and ruby text in complex cases (#247).

### LINKED ISSUE #247: How to handle mismatched sequences in complex ruby
I may have missed it, but i think the spec doesn't say anything about what should happen if, for example, an rbc contains 2 rb items and the rtc contains 4 rt items, or vice versa.

cconcolato:
closes #247 by indicating that mapping btw rb and rt in complex cases is implementation dependent


spec/ttml2.xml
```diff
@@ -11650,6 +11650,9 @@ base delimiter text delimiter =
 &lt;/span&gt;
 </eg>
 </note>
+<note role="clarification">
+<p>The syntax of TTML allows using multiple Ruby Base Content within a Ruby Base Container and multiple Ruby Text Content within a Ruby Text Container all within the same Ruby Container. This version of TTML leaves the mapping between Ruby Base Content and Ruby Text Content implementation defined.</p>
```
nigelmegitt: Two things:

1. (a nit) The line is very long - it's easier to review if split.
2. The second sentence needs to be scoped to situations where there is not a 1:1 match - this is not apparent from the context set by the first sentence.

cconcolato: Will do.

cconcolato: @nigelmegitt Done. Can you check?

spec/ttml2.xml
```diff
@@ -11317,7 +11317,10 @@ base delimiter text delimiter =
 </eg>
 </note>
 <note role="clarification">
-<p>The syntax of TTML allows using multiple Ruby Base Content within a Ruby Base Container and multiple Ruby Text Content within a Ruby Text Container all within the same Ruby Container. This version of TTML leaves the mapping between Ruby Base Content and Ruby Text Content implementation defined.</p> 
+<p>The syntax of TTML allows using multiple Ruby Base Content within a Ruby Base Container and multiple Ruby Text Content within 
+a Ruby Text Container all within the same Ruby Container. This version of TTML leaves the mapping between Ruby Base Content and 
+Ruby Text Content implementation defined when the number of Ruby Text Content does not match the number of Ruby Base Content in
```
nigelmegitt: s/defined/undefined

nigelmegitt: I would also edit "match" to "equal" for clarity

cconcolato: @nigelmegitt "implementation defined" is already used once in the spec. "implementation dependent" is used 10 times. "implementation specific" is used 8 times. But "implementation undefined" is not used and I don't think it makes sense to say "implementation undefined".

nigelmegitt: @cconcolato sorry, you're right, I meant change "implementation defined" to "undefined".

PR #625, opened 2018-02-02
Use glyph area wording for textOrientation (#281).

### LINKED ISSUE #281: Upright orientation involves more than just glyph orientation
10.2.46 tts:textOrientation
http://w3c.github.io/ttml2/spec/ttml2.html#style-attribute-textOrientation

> If the value of this attribute is upright, then, in vertical writing modes, glyphs from horizontal scripts are set upright, i.e., using their nominal orientation in horizontal text, while glyphs from vertical scripts are not affected. In addition, for purposes of bidirectional processing, this value causes all affected characters to be treated as strong left-to-right, i.e., to be treated as if a tts:direction of ltr and tts:unicodeOverride of override were applied.

There are additional things to bear in mind here. The treatment as strong left-to-right will put arabic script characters in the correct visual order down the vertical line, but it should also be said that the characters should use the isolated form.  Furthermore, the rotations should be applied to groups of glyphs that constitute a grapheme cluster, so that for example indic syllables remain together. (Although some consonant clusters are not fully encompassed by grapheme clusters even, in scripts like devanagari.)

I think there's some wording to this effect in the CSS spec that you could look at.

cconcolato:
closes #281


cconcolato:
@nigelmegitt Thanks, I saw them too but the resolution was explicit about the "quoted text". I will update the PR.

cconcolato:
@nigelmegitt can you check again?

I think it remains not very clear how the orientation of a glyph area is determined from its associated characters. In UTR50, it speaks about 'grapheme cluster' and says that the first character gives the orientation (with one exception), but IIUC during the call @skynavga did not want to refer to that complex notion. Should we leave the text as is and let implementation figure it out or should we repeat the UTR50 text replacing 'grapheme cluster' with 'glyph area'?

nigelmegitt:
> Should we leave the text as is and let implementation figure it out or should we repeat the UTR50 text replacing 'grapheme cluster' with 'glyph area'?

@cconcolato I think we should leave as is unless we need to introduce "grapheme cluster" for other reasons, in which case we should change this text too, depending on how that is done.

PR #627, opened 2018-02-02
In tts:fontSize replace applies and expresses with determines (#548).

### LINKED ISSUE #548: A single length value for tts:fontSize specifies both height and width.
One of two points arising from w3c/ttml1#301

The language in tts:fontSize:

> If a single `<length>` value is specified, then this length applies equally to the horizontal and vertical size of a glyph's EM square; if two `<length>` values are specified, then the first expresses the horizontal size and the second expresses vertical size.

is not clear enough. What it means is that the value _specifies_ or _defines_ both the horizontal and vertical size - the words "applies" and "expresses" are used instead in the same paragraph to mean the same thing, and neither is quite right.

Suggest rewording to:

_If a single `<length>` value is specified, then this length defines equally both the horizontal and vertical size of a glyph's EM square; if two `<length>` values are specified, then the first defines the horizontal size and the second defines the vertical size._

I would potentially be happy with other wording as long as the same is used in both and it is clear that the specified value is setting the em square size.

nigelmegitt:
Closes #548 as per the WG's resolution.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -10094,10 +10094,10 @@ then relative to nearest ancestor <loc href="#terms-styled-element">styled eleme
 </tr>
 </tbody>
 </table>
-<p>If a single <loc href="#style-value-length">&lt;length&gt;</loc> value is specified, then this length applies
-equally to the horizontal and vertical size of a glyph's EM square;
-if two <loc href="#style-value-length">&lt;length&gt;</loc> values are specified, then the first expresses the horizontal
-size and the second expresses vertical size.</p>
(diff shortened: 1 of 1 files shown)


nigelmegitt:
Well @cconcolato you're an Editor, you can merge it! We generally wait at least 3 days before merging an editorial fix unless it's a really obvious typo style issue, just to give fair chance for people to take a quick look, and more approvals = better generally.

cconcolato:
@nigelmegitt In that particular given that I was the first to review, my comment was a hint for the second reviewer @skynavga or for the author of the PR (you) to merge. Wait for 3 days is fine also.

PR #628, opened 2018-02-02
Clarify the note text around region association (#592).

### LINKED ISSUE #592: Possible misinterpretation of note in §11.3.1.2 Inline Regions.
The note:

> A content element can only be associated with a single region. Consequently, if a content element specifies a region attribute, then any explicit inline region specification is ignored.

is incorrect - we have established through discussion of other issues that some content elements are associated with multiple regions. I think this is about explicit association by specifying a `region` attribute. Suggest rewording to:

_A content element can only **specify** a single region. Consequently, if a content element specifies a region attribute, then any explicit inline region specification is ignored._

nigelmegitt:
Closes #592 by editing the note text about region association with a content element to clarify that it refers to direct association via a region attribute or child region element.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -16539,7 +16539,7 @@ which generates out-of-line regions that correspond with the
 explicit inline regions. This latter procedure additionally binds <loc href="#terms-content-element">content elements</loc> associated with inline regions
 to the corresponding generated out-of-line regions.</p>
 <note role="elaboration">
-<p>A <loc href="#terms-content-element">content element</loc> can be directly associated with only a single
+<p>A <loc href="#terms-content-element">content element</loc> can be explicitly associated with only a single
 <loc href="#terms-region">region</loc>, either by an attribute specification or via a child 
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -16539,8 +16539,9 @@ which generates out-of-line regions that correspond with the
 explicit inline regions. This latter procedure additionally binds <loc href="#terms-content-element">content elements</loc> associated with inline regions
 to the corresponding generated out-of-line regions.</p>
 <note role="elaboration">
-<p>A <loc href="#terms-content-element">content element</loc> can only be associated with a single
-<loc href="#terms-region">region</loc>. Consequently, if a <loc href="#terms-content-element">content element</loc>
+<p>A <loc href="#terms-content-element">content element</loc> can be directly associated with only a single
```
skynavga: Change "directly" to "explicitly", which is more consistent with usage elsewhere.

nigelmegitt: @skynavga Done.

PR #629, opened 2018-02-02
Quote value syntax keywords (#587).

### LINKED ISSUE #587: Quote all keywords in value syntax expression definitions.

nigelmegitt:
Closes #587 by quoting all the keywords in the vocabulary definitions.

Opening as a pull request now, however there is an open question about whether keywords need to be quoted in the element XML Representations too. I'm afraid the answer is going to be "yes"...

[updated] ... taking the answer to be "yes".

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -6105,10 +6105,10 @@ argument-list
 <td>
 <eg xml:space="preserve">
 &lt;bound-parameter&gt;
-  : forced
-  | mediaAspectRatio
-  | mediaLanguage
-  | userLanguage
+  : "forced"
+  | "mediaAspectRatio"
+  | "mediaLanguage"
+  | "userLanguage"
 </eg>
 </td>
 </tr>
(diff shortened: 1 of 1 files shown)


nigelmegitt:
All keywords now quoted and emphasised consistently in the XML Representations of elements as well as the attribute vocabulary definitions. 

This is good to review now please.

skynavga:
The use of unquoted keywords in the XML representations is based on similar conventions in other specifications, e.g, the XML schema specifications from which we borrowed conventions. However, for consistency, I suspect it is better to use one convention throughout, both for attribute values and XML representations.

spec/ttml2.xml
```diff
@@ -8190,7 +8190,7 @@ element in a block or inline context.</p>
 <tr>
 <td><emph>Values:</emph></td>
 <td>
-none | <loc href="#embedded-content-value-image">&lt;image&gt;</loc>
+"none" | <loc href="#embedded-content-value-image">&lt;image&gt;</loc>
```
skynavga: should be ``<code>"none"</code>``

nigelmegitt:
@skynavga I came to the same conclusion. The convention is actually described in the document conventions section. A particularly careful reader might spot that the convention of surrounding literals with quotes is defined for _value syntax expressions_ and might wonder if that includes an expression in an XML representation or not.

I think it is clearest if we put all literals in quotes, especially since it avoids the confusion over, say, `xml:id = ID`, which is emphatically _not_ supposed to mean `xml:id = "ID"`, for example.

nigelmegitt:
> Was there a particular reason you didn't add quotes to keywords serving as default (attribute) or initial (property) values? I'm just wondering.

@skynavga yes, I did think about that. My reasoning is that the quotes are there to define the syntactical options, but the actual values do not include the quote characters. Defaults are always actual values not type definitions.

spec/ttml2.xml
nigelmegitt: @skynavga Done

PR #575, opened 2018-01-18
Apply fix for ttml1/#201 (#358).

cconcolato:
partially addresses #358

PR #632, opened 2018-02-06
Fill in definition of tts:fontSelectionStrategy, add to schema, fix t…

### LINKED ISSUE #371: Enhance font selection semantics.
Define ``tts:fontSelectionStrategy`` style property based on                                                                                 
the XSL 1.1 font-selection-strategy property and [TTWG ML thread](http://lists.w3.org/Archives/Public/public-tt/2014Nov/0051.html).

Specify font selection semantics, including how multiple author defined font resources combine                 
with (local) platform defined font resources to obtain an ordered list of font resources for performing character to glyph mapping.

Editorial note from 2014-11-21.

### LINKED ISSUE #661: Specify that referenced fonts have priority in font selection strategy.

skynavga:
…ypo (#371,#575).

Closes #371.
Closes #661.


spec/ttml2.xml
```diff
@@ -9660,15 +9660,93 @@ example.</p>
...
+</td>
+</tr>
+<tr>
+<td><emph>Initial:</emph></td>
+<td><code>auto</code></td>
+</tr>
+<tr>
+<td><emph>Applies to:</emph></td>
+<td>
+<loc href="#content-vocabulary-p"><el>p</el></loc>
```
nigelmegitt: In what sense does this apply to `p`?

spec/ttml2.xml
```diff
@@ -9660,15 +9660,93 @@ example.</p>
...
+</tbody>
+</table>
+<p>For the purpose of determining applicability of this style property,
+each character child of a <el>p</el> element is considered to be enclosed in an anonymous
+span.</p>
+<p>In addition to the font selection critera specified by <bibref ref="xsl11"/>, &sect; 7.9.3, the
+following criteria also apply:</p>
+<ulist>
+  <item><p><loc href="#style-attribute-fontKerning"><att>tts:fontKerning</att></loc>; and</p></item>
+  <item><p><loc href="#style-attribute-fontShear"><att>tts:fontShear</att></loc>.</p></item>
```
nigelmegitt: Using font shear and kerning seems to be beyond anything even in CSS. Further, `fontShear` seems to be applied to glyphs _after_ selection, so I don't understand how it can affect selection.

spec/ttml2.xml
skynavga: In the same sense that font{Family,Kerning,Shear,Size,Style,Variant,Weight} apply.

spec/ttml2.xml
skynavga: Irrelevant. Many of our style properties go beyond CSS. In any case, shear can be mapped to specific faces or variation axes (in CSS4 Fonts), and kerning is supported by some fonts but not others (so if you ask for a kerning, it should favor fonts that support kerning).

spec/ttml2.xml
```diff
@@ -9660,15 +9660,93 @@ example.</p>
...
+</ulist>
+<p>If a computed value of the property associated with this attribute is not supported,
+then a <loc href="#terms-presentation-processor">presentation processor</loc> must use the value <code>normal</code>.</p>
```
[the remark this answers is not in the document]
skynavga: Good catch. Will change to ``auto``.

skynavga:
Note that font selection strategy has been included in the spec for a few years now, even though its details weren't fully spelled out.

spec/ttml2.xml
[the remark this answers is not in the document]

skynavga:
One more point worth noting: every implementation that actually treats ``tts:fontFamily`` as a list already does some form of font selection. Now, since the initial value of ``tts:fontSelectionStrategy`` is ``auto``, and the semantics of ``auto`` are "implementation defined", then all such implementations already support this initial value at least. It is also  (now) the fallback value in case ``character`` is not supported.

spec/ttml2.xml
nigelmegitt: In other words, this is so that the styles can be applied to character content in anonymous spans?

spec/ttml2.xml
nigelmegitt: [CSS4 Fonts](https://www.w3.org/TR/css-fonts-4/) is currently at FPWD so we cannot reference it normatively yet. I see what you mean though - it allows fonts to be selected based on in-font support for `oblique` property values. If we are going to adopt that model then presumably the `tts:fontShear` section would need to be modified to allow for _either_ application of the transformation using the feature support in the font, _or_ for transformation of the glyph areas after font and glyph selection.

spec/ttml2.xml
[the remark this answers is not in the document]
nigelmegitt: Thanks.

nigelmegitt:
> font selection strategy has been included in the spec for a few years now, even though its details weren't fully spelled out.

In my view an empty placeholder signals intent, but is not the same as "included in the spec".

spec/ttml2.xml
```diff
@@ -18060,7 +18152,7 @@ acting the part of a character.</p>
 <p>The use of <phrase role="deprecated">a reference from the
 <att>agent</att> attribute of a <el>ttm:actor</el> element to its
 parent <el>ttm:agent</el> element is deprecated</phrase>, and,
-furthermore, the semantics of such a circular chain of references is
+furthermore, the semantics of such a circular chain of references are
```
nigelmegitt: This change is not relevant to the issue.

skynavga:
Not sure how that happened (perhaps during a merge with master). I will fix.

spec/ttml2.xml
skynavga: Actually, on further thought, it should not be applied to p (but it it is inheritable from p). The reason that font{Family,Kerning,Shear,Size,Style,Variant,Weight} apply to p is because it is necessary to assign a default font to each paragraph in order to determine the nominal line height metrics (ascent and descent) that apply. Note that this (having these apply to p) is new in TTML2 (but only implied in TTML1); thus the need for new #fontSize-p etc designators proposed in #657. In contrast, contextual character control (by fontSelectionStrategy) does not come into play for this default paragraph font since only anonymous span descendants contain character content.

skynavga:
@nigelmegitt regarding your earlier comment (which I cannot now find above) regarding

> @@ -18060,7 +18152,7 @@ acting the part of a character.</p>
><p>The use of <phrase role="deprecated">a reference from the
><att>agent</att> attribute of a <el>ttm:actor</el> element to its
>parent <el>ttm:agent</el> element is deprecated</phrase>, and,
>-furthermore, the semantics of such a circular chain of references is
>+furthermore, the semantics of such a circular chain of references are
>
>This change is not relevant to the issue.

I see this is merely a fix to a grammatical typo. I routinely fold such fixes into (unrelated) PRs when I find them while preparing a PR.

spec/ttml2.xml

skynavga:
@nigelmegitt I believe all your (non-process) comments have been addressed

nigelmegitt:
> I routinely fold such fixes into (unrelated) PRs when I find them while preparing a PR.

@skynavga I can understand your motivation for doing so, but it muddies the traceability of fixes to issues and makes review more complex - also it means that if a pull request does _not_ get merged then the quick fix can easily get dropped too; I would prefer it if there were a simple no-issue pull request on its own to make the grammar fix. Or we could have a generic "grammar/typo fix" issue to link such pull requests to.

spec/ttml2.xml
nigelmegitt: Thanks, that makes sense.

spec/ttml2.xml
```diff
@@ -9470,15 +9470,106 @@ example.</p>
...
+span.</p>
+<p>The presentation semantics of this property are those defined by <bibref ref="xsl11"/>, &sect;7.9.3,
+where the set of selection criteria are extended to include the following properties:</p>
+<ulist>
+  <item><p><loc href="#style-attribute-fontKerning"><att>tts:fontKerning</att></loc></p></item>
+  <item><p><loc href="#style-attribute-fontShear"><att>tts:fontShear</att></loc></p></item>
+</ulist>
+<p>When selecting among font resources that satisfy the font selection criteria, font resources that are referenced
+by a <loc href="#embedded-content-vocabulary-font"><el>font</el></loc> element have priority over application or system supplied
+font resources.</p>
```
nigelmegitt: #661 is to specify that _loaded_ fonts have priority in font selection; this text specifies that _referenced_ fonts have priority. Those are not the same thing at all. I think #661 was supposed to deal with lazy loading of fonts, so this change does not address the issue in my view.

skynavga:
@nigelmegitt no. #661 doesn't care about lazy vs eager; it only cares about precedence; i.e., fonts referenced by a font element have precedence over other fonts

nigelmegitt:
@skynavga if you don't intend #661 to refer to lazy vs eager it probably shouldn't use the word "loading" in the issue name. Maybe include some description of issues to avoid misunderstandings like this too? The title is not enough here.

spec/ttml2.xml
nigelmegitt: @skynavga just changed the name of #661, so this now appears to be included in this pull request as being fixed.

PR #636, opened 2018-02-08
Revert <set> to TTML1 syntax and semantics

palemieux:
Remove fill and repeatCount attributes (closes #635)
Revert target style property value to a single value (closes #633)

PR #639, opened 2018-02-09
Add #set* feature designators, expand/refactor #animate* and #animati…

### LINKED ISSUE #635: Revert <set> to single attribute and remove @fill and @repeatCount attributes
The syntax of `set` was expanded to support setting multiple style properties and the `fill` and `repeatCount` attributes, without defining new feature designators.

Propose removing these new capabilities unless use cases can be provided.

skynavga:
…on* feature designators (#635).

Closes #635.
Partially addresses #657.

Intended to replace the part of #636 that deals with ``fill`` and ``repeatCount`` on ``set``. Additionally, expands and/or refactors #animate* and #animation feature designators for better coverage and flexibility.


spec/ttml2.xml
```diff
@@ -20978,6 +20999,41 @@ transforming the <loc href="#style-attribute-rubyReserve"><att>tts:rubyReserve</
...
+<item><p><loc href="#animation-vocabulary-set"><el>set</el></loc></p></item>
+</ulist>
+<p>A TTML <loc href="#terms-presentation-processor">presentation processor</loc> supports the <code>#set</code>
+feature if it implements presentation semantic support for the
+same vocabulary enumerated above.</p>
+<p>Notwithstanding the above, support for the <code>#set</code> feature does not require
+support for the <code>#set-fill</code> or <code>#set-repeat</code> feature.</p>
+<note role="elaboration">
+<p>In particular, support for the <code>#set</code> feature is intended to be semantically equivalent
+to support for the <code>#animation</code> feature as defined by <bibref ref="ttml1"/>.</p>
```
nigelmegitt: But the `#animation` feature is still in TTML2 - is that the same thing, in which case the text should include a link? Or should it be removed from TTML2? Or edited to qualify the definition such that it only relates to TTML1 semantics and syntax? At the very least the `#set` feature needs to be referenced from the definition of `#animation` with some descriptive text and it needs to be clear if they are true synonyms or if `#set` somehow adds something relative to `#animation`.

skynavga:
@palemieux added #set-multiple-styles

skynavga:
@nigelmegitt the alternative to having the (TTML1) ``#animation`` feature referencing the new ``#set`` feature is to define both with the same definition, which violates the DRY principle

skynavga:
@palemieux are there any more comments you need resolved on this PR? if not, then please approve

nigelmegitt:
> the alternative to having the (TTML1) #animation feature referencing the new #set feature is to define both with the same definition, which violates the DRY principle

@skynavga agreed - why not just keep `#animation` as before in that case?

skynavga:
@nigelmegitt firstly, can you agree that it doesn't matter from a technical perspective? i.e., the new formulation has exactly the same meaning (in terms of defining what support means) as the old formulation? if so, then there is no problem here; the new formulation doesn't introduce duplicate spec text, whereas retaining the old one would

nigelmegitt:
@skynavga yes, I think so. The situation as I see it is:

TTML1: 
*`#animation` => support for `set` element.

TTML2:
* `#animation` => support for `set` element.
* `#set` => support for `set` element.

That looks like duplication to me. Don't get me wrong though, I think `#set` is the better feature designator as a label, and we have to deal with the legacy `#animation` feature somehow. Why not say something like:

_This feature designator is retained for backwards compatibility with TTML1; it is equivalent to the `#set` feature defined in this specification; thus a TTML processor that supports `#set` also supports `#animation`. It is considered an error to specify different `@value` attribute values in `#animation` and `#set` `feature` elements within a timed text profile document instance or as children of a `profile` element._

In place of the existing text?

skynavga:
On Tue, Feb 13, 2018 at 10:29 AM, Nigel Megitt <notifications@github.com>
wrote:

> @skynavga <https://github.com/skynavga> yes, I think so. The situation as
> I see it is:
>
> TTML1:
> *#animation => support for set element.
>
> TTML2:
>
>    - #animation => support for set element.
>    - #set => support for set element.
>
> actually, that is not how it is defined, it is

   - #animation => support for #set feature
   - #set => support for set element

thus, no duplication (i.e., no two features are defined identically);
clearly, however, we have to retain #animation for TTML1 compatibility; and
we also need new #set-* designator, which, without a #set designator, would
appear strange indeed


>
>
> That looks like duplication to me. Don't get me wrong though, I think #set
> is the better feature designator as a label, and we have to deal with the
> legacy #animation feature somehow. Why not say something like:
>
> *This feature designator is retained for backwards compatibility with
> TTML1; it is equivalent to the #set feature defined in this specification;
> thus a TTML processor that supports #set also supports #animation. It is
> considered an error to specify different @value attribute values in
> #animation and #set feature elements within a timed text profile document
> instance or as children of a profile element.*
>
The current text is accurate and succinct, so I see no reason to change it;
I would, however, be willing to add a note (under the definition of
#animation) that says:

*The #animation feature designator is retained for backward compatibility
> with TTML1. Profiles designed for use with TTML2 and later versions are
> recommended to use the #set feature designator.*

In place of the existing text?

palemieux:
Built at https://rawgit.com/w3c/ttml2/issue-0635-add-set-fill-repeat-designators-build/index.html#feature-timing

skynavga:
@palemieux thanks

nigelmegitt:
> The current text is accurate and succinct, so I see no reason to change it;
I would, however, be willing to add a note (under the definition of
#animation) that says:

> *The #animation feature designator is retained for backward compatibility

@skynavga Yes please.

Do you not think we need to say something about what happens if, say, `#set` is `required` and `#animation` is `prohibited`? That seems like an error state to me.

nigelmegitt:
>>  TTML2:
>>
>>    - #animation => support for set element.
>>    - #set => support for set element.

> actually, that is not how it is defined, it is
>- #animation => support for #set feature
>  - #set => support for set element
>
>thus, no duplication (i.e., no two features are defined identically);

@skynavga This is bordering on sophistry: they both evaluate to the exactly the same thing, so it is duplication in practice.

skynavga:
@nigelmegitt not in the slightest; I presume you understand the notion of a pointers in the C language? so by analogy, we have

TTML1
elementType SET;
elementType *animation = &SET;

TTML2
elementType SET;
elementType *set = &SET;
elementType **animation = &set;

therefore

*animation [in TTML1] == **animation [in TTML2]

In any case, if #animation is permitted or required, but #set is prohibited, then, while this may be an error in logic on the part of the profile author, it is well defined semantically in terms of profile semantics.

However, in the interest of "informing the user not to do stupid things" I will add a note to that effect.

nigelmegitt:
> In any case, if #animation is permitted or required, but #set is prohibited, then, while this may be an error in logic on the part of the profile author, it is well defined semantically in terms of profile semantics.

Really? Surely it is impossible for a document or processor to conform in that case? That seems to be badly defined semantically, not well defined. To make it well defined there should be a clear error state, or alternatively a precedence order to derive which feature requirement applies.

skynavga:
@nigelmegitt addressed your change request, please approve

spec/ttml2.xml
```diff
@@ -18766,6 +18766,8 @@ feature if it implements presentation semantic support for the <att>repeatCount<
 <ulist>
 <item><p><loc href="#feature-set"><code>#set</code></loc></p></item>
 </ulist>
+<p>It is an error for a profile to require support for the <code>#animation</code> feature but prohibit support
+for the <loc href="#feature-set"><code>#set</code></loc> feature.</p>
```
nigelmegitt: Needs to say "permit" instead of "require", since if `#animation` is optional and `#set` is prohibited that is equally an error state.

spec/ttml2.xml
```diff
@@ -20884,6 +20888,8 @@ same vocabulary enumerated above.</p>
 <item><p><loc href="#feature-set-multiple-styles"><code>#set-multiple-styles</code></loc></p></item>
 <item><p><loc href="#feature-set-repeat"><code>#set-repeat</code></loc></p></item>
 </ulist>
+<p>It is an error for a profile to require support for the <code>#set</code> feature but prohibit support
+for the <loc href="#feature-animation"><code>#animation</code></loc> feature.</p>
```
nigelmegitt: Needs to say "permit" instead of "require", since if `#set` is optional and `#animation` is prohibited that is equally an error state.

spec/ttml2.xml

spec/ttml2.xml

PR #649, opened 2018-02-11
Specify semantics of percentage length used with tts:{bpd,ipd} (#369).

### LINKED ISSUE #369: Specify block and inline progression dimension percentages.
Specify resolution of percentage values.

Editorial notes from 2014-11-29.

skynavga:
Closes #369.


spec/ttml2.xml
```diff
@@ -10300,18 +10325,43 @@ where horizontal and vertical are always interpreted in an absolute sense.</p>
...
+<ednote>
+<name>Example Image - Inline Progression Dimension</name>
+<date>2018-02-11</date>
+<edtext>Add example image of above fragment.</edtext>
+</ednote>
+<note role="explanation">
+<p>In the above example, the <att>tts:ipd</att> attribute on the <loc href="#content-vocabulary-p"><el>p</el></loc> element
+assigns a fixed dimension to the width of the content rectangle of the block area generated by the paragraph, while
+the <att>tts:ipd</att> attribute on the <loc href="#content-vocabulary-span"><el>span</el></loc> element assigns a fixed
+dimension to the width of the content rectangle of the inline block area generated by the span.</p>
```
nigelmegitt: I have insufficient information to be able to deduce what the image should look like for the fragment. For example:

* if the region width is less than ipd, does wrapping occur at the ipd boundary or the region boundary?
* if the text "getting green" occupies a width less than 150px, what text alignment semantics apply?

PR #650, opened 2018-02-11
Specify font loading semantics (#368).

### LINKED ISSUE #368: Specify font loading semantics.
Specify font loading semantics, making as much use as possible (by reference) of material found at                                                         
[CSS Font Module Level 3, Font Loading Guidelines](http://www.w3.org/TR/css3-fonts/#font-face-loading).

Editorial note from 2014-11-21.

skynavga:
Closes #368.


spec/ttml2.xml
```diff
@@ -6910,12 +6910,9 @@ must be considered to be equal to the value(s) of the same named font characteri
...
 </note>
-<issue id="issue-ttml2-368">
-<head>Specify font loading semantics.</head>
-<source><loc href="https://github.com/w3c/ttml2/issues/368">https://github.com/w3c/ttml2/issues/368</loc></source>
-<p>Specify font loading semantics, making as much use as possible (by reference) of material found at
-<loc href="http://www.w3.org/TR/css3-fonts/#font-face-loading">CSS Font Module Level 3, Font Loading Guidelines</loc>.</p>
-</issue>
+<p>For the purpose of dereferencing (downloading) font resources, a <loc href="#terms-content-processor">content processor</loc>
+may defer the downloading process until such time that the content of the font resource is necessary in order to perform the necessary
+(transformation or presentation) processing.</p>
```
nigelmegitt: This appears to permit the processor to wait until presentation time, then load the font, but omits mention of fallback behaviour between requesting the font resource and it becoming available. It appears that either of at least the following options might be permitted:

a. Display the text in a different font until the required font is loaded.
b. Delay presenting the text until the required font is loaded.

It must be the first one, since getting the time right is essential in timed text and there can never be a guarantee that external resources can successfully be dereferenced. [CSS Fonts Level 3](https://www.w3.org/TR/css-fonts-3/#font-face-loading) has some good words about this in the last paragraph, which I suggest we adopt by copy or reference.

skynavga:
It is not necessary to specify fallback behavior. The default is that it is
implementation dependent, which is just fine. We don't define what fallback
presentation semantics are if any font is not available (whether it is a
downloaded font or not).

skynavga:
@nigelmegitt please approve now that #661 has been changed to remove reference to "loading"

nigelmegitt:
@skynavga #661 was never listed as being fixed by this pull request, so the dependency wasn't there (it was #632 that claimed to close #661, and was already approved and merged after #661 was changed).

nigelmegitt:
> It is not necessary to specify fallback behavior. The default is that it is implementation dependent, which is just fine.

@skynavga now that #661 has been changed to clarify that font prioritisation is not based on which font is loaded, we now need something. Your argument in last week's call was that we would have to apply the same fallback logic to all remote resources that need loading, but that is not the case - only for fonts (and in this pull request) do we make this explicit statement about implementations being permitted to do lazy loading.

Is it fine to be implementation dependent? I suppose that's acceptable. I've raised #675 for this, though I think it would have been straightforward and reasonable to include the relevant text from CSS Fonts Level 3 here.

skynavga:
Do not merge until CR1 is published.

PR #651, opened 2018-02-12
Update requirements for claim of document (content) compliance (#495).

### LINKED ISSUE #495: Claims section does not take into account ttp:contentProfiles and ttp:processorProfiles.
The claims section only references `ttp:profile` and ignores profiling vocabulary introduced in TTML2, such as `ttp:contentProfiles` and `ttp:processorProfiles`.

skynavga:
Closes #495.


spec/ttml2.xml
```diff
@@ -2142,17 +2142,16 @@ identify the utilized or supported TTML vocabulary profile(s) as
...
-a <el>ttp:profile</el> element as a child of the
-<el>head</el> element as defined by
-<specref ref="profile-vocabulary-profile"/>.</p>
+<p>A <loc href="#terms-document-instance">document instance</loc> for which a compliance claim is made must be associated with</p>
+<olist>
+<item><p>a non-null <loc href="#terms-effective-content-profile">effective content profile</loc>
+derived from an explicitly specified <loc href="#profile-attribute-contentProfiles"><att>ttp:contentProfiles</att></loc> attribute
+or one or more <loc href="#terms-top-level-profile">top-level</loc> <loc href="#terms-content-profile">content profiles</loc>; or</p></item>
+<item><p>a non-null <loc href="#terms-effective-processor-profile">effective processor profile</loc>
+derived from an explicitly specified <loc href="#profile-attribute-processorProfiles"><att>ttp:procesesorProfiles</att></loc> or
```
nigelmegitt: typo: procesesorProfiles -> processorProfiles

spec/ttml2.xml
```diff
@@ -2142,17 +2142,16 @@ identify the utilized or supported TTML vocabulary profile(s) as
...
+<loc href="#profile-attribute-profile"><att>ttp:profile</att></loc> attribute
```
nigelmegitt: this is saying the same thing twice: a top level profile _is_ specified by a `ttp:profile` attribute: no need to say it twice.

spec/ttml2.xml
```diff
@@ -2142,17 +2142,16 @@ identify the utilized or supported TTML vocabulary profile(s) as
...
-or (2)
-a <el>ttp:profile</el> element as a child of the
-<el>head</el> element as defined by
-<specref ref="profile-vocabulary-profile"/>.</p>
+<p>A <loc href="#terms-document-instance">document instance</loc> for which a compliance claim is made must be associated with</p>
+<olist>
+<item><p>a non-null <loc href="#terms-effective-content-profile">effective content profile</loc>
+derived from an explicitly specified <loc href="#profile-attribute-contentProfiles"><att>ttp:contentProfiles</att></loc> attribute
+or one or more <loc href="#terms-top-level-profile">top-level</loc> <loc href="#terms-content-profile">content profiles</loc>; or</p></item>
+<item><p>a non-null <loc href="#terms-effective-processor-profile">effective processor profile</loc>
```
nigelmegitt: Absent a process for inferring a content profile from a processor profile, this makes no sense. A document instance can only be validated against an effective content profile, not against a processor profile. I know this because of the definition of processor profile, which makes reference only to processors and not to documents. Also, we have discussed at huge length the logical difference between content and processor profiles, so I'm not now willing to elide the difference here - if this had been possible earlier we would only have had one kind of profile, and things would have been a lot simpler!

Suggest simply removing the second bullet about effective processor profile. That will mean that the following two comments will not need to be addressed either, though they're simple.

palemieux: @nigelmegitt What do you make of the current prose in the Claims section of TTML1, which equates compliance claims with `ttp:profile`, which, when used as an element, defines a processor profile?

palemieux:
Built at https://rawgit.com/w3c/ttml2/issue-0495-profile-claims-build/index.html#claims

spec/ttml2.xml
nigelmegitt: @palemieux I think the reason we have revamped the profiles language in TTML2 is that it was confused before. I don't think we should take the TTML1 claims section as a reference point for TTML2, having done that work.

palemieux: Understood. By the way, the definition of _effective content profile_ seems to support your position.

> When performing validation on a given document instance, then this validation is performed by making use of a document's effective content profile.

skynavga: Except this doesn't work when the effective content profile is null.

spec/ttml2.xml
skynavga: Nope. This is capturing the fact that there are three ways to derive the effective processor profile:

1. by specifying ttp:processorProfiles attribute, which may refer to internal (top-level) profiles or external profiles;
2. by specifying ttp:profile attribute, which may refer to an internal (top-level) profile or an external profile;
3. by specifying neither ttp:processorProfiles or ttp:profile attributes, but specifying one or more top-level profiles, i.e., ttp:profile elements

spec/ttml2.xml
skynavga: Sure we can validate against a processor profile. There is no other option for TTML1. We simply associate a set of schemas and validation rules with a processor profile. That is how TTV works with TTML1 (and IMSC1) documents.

skynavga: To be clear, for a TTML1 document, the only formal definition of profile is related to processor requirements and not content requirements; however, in the absence of a formalism for expressing the latter, what we do in practice (for TTML1) is to treat the processor profile as a moniker for a variety of rules, one subset of which are projected onto documents. It is, indeed, a muddled state of affairs in TTML1, but we now have the problem of dealing with both TTML2, where things are more clear, and TTML1 where they aren't. And even then, we don't force a TTML2 document to declare a content profile, but only a processor profile (or supply a default for the latter if the author is lazy). But we never supply a default for content profile, even though we may, for the purpose of validation, fall back onto a default association of schemas with processor profiles.

skynavga: So, I must reject your request to "remov[e] assessing document compliance using processor profile" since that is what is (and all that can be) done in practice for TTML1.

spec/ttml2.xml
nigelmegitt: Ah right, thanks, I confused profile attributes with elements.

spec/ttml2.xml
nigelmegitt: Thanks

spec/ttml2.xml
nigelmegitt: As per https://github.com/w3c/ttml2/pull/651/files/f4c9cb61ddb60522768f2e7fd8284720e98e0ee3..f867394372958098683939095e7d7e21c1f48638#r168159895 I don't accept reference to TTML1 as the semantic model here since we have changed it so much. We do not, in TTML2, have any text that deals with use of a TTML1 style processor profile for content conformance; we do have text that allows a processor profile to be inferred from a content profile, but not the other way around. Check the term definitions in TTML2 for content profile and processor profile. They're very clear.

The resolution I proposed is simple  - just remove the bit about processor profiles. It doesn't invalidate any conformant TTML1 documents to do so.

Including processor profiles here reverts the complex work we have done to clarify the situation around profiles and reintroduces the muddiness. I don't know why we would do that, or indeed why you are open to that @skynavga .

Here's an alternative proposal which I think is even better: before the second bullet about processor profiles, insert:

_A content processor for which a compliance claim is made must be associated with_

Would you be able to accept that?

skynavga: @nigelmegitt no, I cannot accept this last proposal, since the purpose of this text is only in regard to claims about document conformance; the underlying problem here is that we can't merely abandon the language that was in TTML1 regarding such a claim; the new language is carefully worded to use a content profile declaration at first order, and, only then, if no content profile declaration is present, fall back to the TTML1 claims language, somewhat upgraded to add processorProfiles into the mix

furthermore, this language is consistent with the definition of **[construct effective validation profile]**, which uses the __effective processor profile__ as a fall back as well

skynavga:
@nigelmegitt please re-review ASAP

spec/ttml2.xml
nigelmegitt: > the purpose of this text is only in regard to claims about document conformance

@skynavga the first sentence in the section says: "Any claim of compliance with respect to the conformance of a TTML document instance or content processor..." so the purpose of the text is also about processor conformance. If you do want to constrain it to document conformance, and you allow for association only with a processor profile, that begs the question how it is possible to claim that a document instance associated only with a processor profile is conformant to an unspecified content profile.

My proposal does not abandon the language in TTML1, it clarifies when content profiles are required vs when processor profiles are required. 

I'm going to label this for agenda, we seem to be going round in circles.

PR #652, opened 2018-02-12
Resolve confusion between default and implicit (#590).

### LINKED ISSUE #590: Does "implicit" begin value mean the "default" value of the begin attribute?
The following is incorrect: in the case of a child of a sequential container, the implicit begin time is equal to the desired end time of the previous sibling, or the desired begin time of the parent if no previous sibling exists.

>As defined by [SMIL 3.0], § 5.4.3, if no begin attribute is explicitly specified on a timed element, then a default (implicit) value is determined by the nearest time container ancestor; however, in TTML, that ancestor is either a parallel time container or a sequential time container, both of which apply a default value of 0.

Suggest referring to SMIL, and not elaborating.

skynavga:
Closes #590.


spec/ttml2.xml
```diff
@@ -16878,7 +16878,7 @@ by <bibref ref="smil3"/>, &sect; 5.4.3,
 while taking into account any overriding semantics defined by this specification.</p>
 <note role="clarification">
 <p>As defined by <bibref ref="smil3"/>, &sect; 5.4.3,
-if no <att>begin</att> attribute is explicitly specified on a timed element, then a default (implicit) value
+if no <att>begin</att> attribute is explicitly specified on a timed element, then the default value of the attribute
```
nigelmegitt: This statement is not actually true. SMIL 3 does not define any default begin attribute value for par or seq time containers. However it does define the semantics in the absence of such an attribute value. This distinction is important because it allows smpte discontinuous to be handled also. I think the sentence should read:

```
if no <att>begin</att> attribute is explicitly specified on a timed element, then the
beginning of the element's active period is resolved as the parent element's resolved
begin time if the parent is a parallel time container, or in relation to both the parent
element and the preceding siblings according to the semantics defined for a sequential
time container. In any case no additional begin time offset is defined in TTML.

**Comment by nigelmegitt** [comment]
> I think the prose should simply point to the appropriate TTML3 section (like TTML1 3ED https://w3c.github.io/ttml1/index.html#timing-attribute-begin)

@palemieux did you mean "point to the appropriate _SMIL3_ section"?

**Comment by nigelmegitt** [comment]
That would work for me too, though I think it is not a great service to the reader. If nobody thinks it incorrect, could we refer to SMIL3 _and_ add an informative note as per https://github.com/w3c/ttml2/pull/652#discussion_r167563559 ?

File: spec/ttml2.xml (hunk 1 above)
Comment (skynavga) [reply spec/ttml2.xml]: @nigelmegitt we are not talking about the beginning of the element's active period; we are talking about the default value of the `begin` attribute (in the XML sense that an attribute may have a default value); the note (as amended here), is entirely correct; this should be clear to any reader who reads _the default value of the **attribute**_ (my emphasis); furthermore, you are neglecting the language in SMIL 3:

for **par** [1]
>The default value of begin for children of a par is "0".

[1] https://www.w3.org/TR/2008/REC-SMIL3-20081201/smil-timing.html#Timing-ParSyntax

for **seq** [2]
>For children of a sequence, the only legal value for begin is a (single) non-negative offset value. The default begin value is "0".

[2] https://www.w3.org/TR/2008/REC-SMIL3-20081201/smil-timing.html#Timing-SeqSyntax

**Comment by skynavga** [comment]
@nigelmegitt awaiting your feedback on https://github.com/w3c/ttml2/pull/652#discussion_r192217402; please process ASAP

**Comment by skynavga** [comment]
@palemieux what SMIL defines as a default attribute value is dependent on SMIL DTDs; we don't use SMIL DTDs, so we need to express default values directly in our spec

File: spec/ttml2.xml (hunk 1 above)
Comment (nigelmegitt) [reply spec/ttml2.xml]: Right, I see that we might be talking about different things. My interpretation is that specifying a default begin value of `0` represents an offset time, and that is resolved relative to the syncbase. The resolution is defined for both par and seq containers by SMIL [offset values](https://www.w3.org/TR/2008/REC-SMIL3-20081201/smil-timing.html#Timing-OffsetValueSyntax). And this is independent of the `markerMode` too.

File: spec/ttml2.xml [hunk 2]
```diff
@@ -17356,7 +17356,7 @@ by <bibref ref="smil3"/>, &sect; 5.4.3,
...
 is determined by the nearest time container ancestor; however, in TTML, that ancestor is either
 a parallel time container or a sequential time container, both of which apply a default value of <code>0</code>.</p>
```
nigelmegitt: `0` is not actually a valid offset time expression in TTML - should this be `0s` instead?

Also, do we need to clarify that this is interpreted as an offset time relative to the element's syncbase as per [SMIL offset time](https://www.w3.org/TR/2008/REC-SMIL3-20081201/smil-timing.html#Timing-OffsetValueSyntax)?

skynavga:
@nigelmegitt addressed comment in https://github.com/w3c/ttml2/pull/652/commits/ad4ac88da2c3f173802b8495e778e1bb9734128b; we do not (and should not) paraphrase or redefine the semantics of begin; so in answer to your question of restating the meaning of "offset", the answer is NO.

skynavga:
@palemieux @cconcolato please (1) approve, (2) dismiss your review, or (3) propose specific text changes;

skynavga:
@nigelmegitt please re-review and approve

nigelmegitt:
@skynavga please could you resolve the merge conflicts?

skynavga:
@nigelmegitt trying, but something appears amiss with travis just now

skynavga:
@nigelmegitt merged

palemieux:
As stated multiple times [before](https://github.com/w3c/ttml2/pull/652#pullrequestreview-95834501), my preference is to simply point to SMIL and omit explanatory text, which is only marginally useful outside the full context of SMIL.

If the absence of informative text is a blocker, then the exact informative text used in SMIL should be used:

_Children of a par begin by default when the par begins (equivalent to begin="0s"). Children of a seq begin by default when the previous child ends its active duration (equivalent to begin="0s");_

skynavga:
@palemieux so, are you prepared to block progress on TTML2 if we don't adopt your preference? it is quite absurd to force the reader to chase down the relevant language in SMIL just to answer the simple question of what the default values are; and, unless there is a substantive error or mischaracterization of the language in SMIL, the language used here works just fine

palemieux:
> so, are you prepared to block progress on TTML2 if we don't adopt your preference

Yes. TTML and SMIL, are complex enough by themselves. I do not see any reason to complicate matters by introducing new terminology and language.

skynavga:
@palemieux then we are going to be blocked  ... because I am not prepared to incorporate language from SMIL that assigns an interpretation to the meaning of `0s`, which is what you are asking for

palemieux:
@skynavga My preference is to use the following text, adapted from TTML1, which addresses both our concerns since no new interpretation is assigned/inferred.

_See [SMIL 3], § 10.4.1, Begin value semantics when no begin attribute is specified._

skynavga:
@palemieux which was added in 3rd Ed draft without my review; further, this requires the reader to chase down the default value in SMIL, which says only:

>If no begin is specified, the default timing is dependent upon the time container.

then the reader has to apply a TTML specific constraint, namely, that only `par` and `seq` time containers apply,

then the reader has to figure out where a default value for begin is defined for these time containers, which leads the reader to Section 5.4.4, where they have to tease out

for `par`:
>The default value of begin for children of a par is "0".

for `seq`:
>For children of a sequence, the only legal value for begin is a (single) non-negative offset value. 
The default begin value is "0".

then the reader has to mentally translate `0` into `0s`, since the former is not a legal value in TTML

unlike your suggestion, the note we ALREADY added in TTML2, about which this most simple edit merely removes one word "(implicit)" and changes `0` to `0s`, does the reader a big favor by doing the leg work described above, and which has a definite clarifying affect, since a reader may fail to follow such necessary hermeneutic steps

nigelmegitt:
@palemieux @skynavga Can we do _both_ what is in this pull request _and_ add the statement under the `begin` attribute semantics referring to SMIL as per TTML1 3rd Ed? They're both useful and true as far as I can see.

skynavga:
@nigelmegitt the pre-existing note already says as much:

>As defined by <bibref ref="smil3"/>, &sect;5.4.3,
if no <att>begin</att> attribute is explicitly specified on a <loc href="#terms-timed-element">timed element</loc>,
then

referring to 5.4.3 twice is redundant and probably confusing

nigelmegitt:
@skynavga sorry, yes, I'm not sure why I missed that before. The notes don't actually say the same thing though. One says "see SMIL for semantics when begin is missing", the other says "as per SMIL, here's the default value for begin". The way I see this, it is important:

1. To refer the reader to SMIL for behaviour when there is no begin attribute
2. To be clear what the default value of the begin attribute is if one is needed/used.

They are separate points that both can usefully be made, possibly by a cleverly constructed edit to the existing note.

skynavga:
@nigelmegitt but yet again, the normative paragraph just before this note says:

>The semantics of the begin attribute are those defined by [SMIL 3.0], §5.4.3, while taking into account any overriding semantics defined by this specification.

the only point being made by the note is to clarify the default value, and not to define semantics

PR #665, opened 2018-02-13
Clarify that metadata has no presentation semantics (#656).

### LINKED ISSUE #656: Metadata has no presentation semantics.
Need to formally state that metadata has no presentation semantics.

skynavga:
Closes #656.


skynavga:
@palemieux I can't really say. At present, TTX does two things with itts:forcedDisplay
1. maps it to a conditional visibility style in order to effect the proper presentation semantics;
2. records the fact that itts:forcedDisplay was present, then later, when constructing an ISD document instance, adds a ttm:metadata element with a <ttm:item name="usesForced"> child, which is eventually serialized for the use of any downstream processor.

nigelmegitt:
I consider this to be editorial, since it merely describes the previous state of the spec and makes no normative changes. @skynavga is there a particular reason why you labelled it as substantive?

skynavga:
@nigelmegitt the following normative text was added; however, I am happy to treat this as editorial if you prefer

```
<p>This specification does not define any presentation semantics for metadata; therefore, a conformant
<loc href="#terms-presentation-processor">presentation processor</loc> may ignore all metadata matter.</p>
```

skynavga:
@palemieux are you requesting a specific change?

skynavga:
@palemieux your questions about ``usesForced`` appear to be orthogonal to the purpose of this issue; can you approve this as is, or, if not, then please suggest a specific change

palemieux:
@skynavga As explained at https://github.com/w3c/ttml2/issues/656#issuecomment-365504992, I do not understand what problem this PR is trying to solve, and it does not look urgent, so I do not believe it can and needs to be approved as-is.

nigelmegitt:
@skynavga I think that added text is editorial, yes.

cconcolato:
This PR was opened after our deadline. I wouldn't consider it editorial but I don't consider it changing anything in the spec. I think I can be approved.

skynavga:
@palemieux ping

PR #667, opened 2018-02-15
Add CR exit criteria

### LINKED ISSUE #663: Determine and specify CR exit criteria in SoTD.
I propose the following exit criteria:

>For every designated feature added in TTML2, the following conditions must be satisfied:
>1. at least one implementation of a transformation processor must support the feature, an example of which is a validating transformation processor that maps use of the feature to an output that indicates successful validation;
>2. at least two implementations of a presentation processors must support each mandatory feature that has defined presentation semantics;
>3. at least one implementation of a presentation processor must support each optional feature that has defined presentation semantics.

nigelmegitt:
Not everything needed to make it a CR, but the basics: switch the role of the spec to cr, add a cr version of `entitiesedcopy.dtd`, change the document status.

Closes #663.

Not intended to be merged into the ED for publication in gh-pages!


tmichel07:
Glenn,


It is up to the director to consider if there is enough 
implementation-experience during Transition.

The 2018 W3C Process mentions "independent interoperable"

https://www.w3.org/2018/Process-20180201/#implementation-experience



6.2.4 Implementation Experience

Implementation experience is required to show that a specification is 
sufficiently clear, complete, and relevant to market needs, to ensure 
that independent interoperable implementations of each feature of the 
specification will be realized. While no exhaustive list of requirements 
is provided here, when assessing that there is adequate implementation 
experience the Director will consider (though not be limited to):

     is each feature of the current specification implemented, and how 
is this demonstrated?
     are there independent interoperable implementations of the current 
specification?
     are there implementations created by people other than the authors 
of the specification?
     are implementations publicly deployed?
     is there implementation experience at all levels of the 
specification's ecosystem (authoring, consuming, publishing…)?
     are there reports of difficulties or problems with implementation?

skynavga:
If you can assure me that SKYNAV's distinct applications, TTV (a TTML
validator), TTX/TTXV (a TTML transformation engine), and TTPE (a TTML
presentation engine) can be presented to the director as distinct
implementations (whether or not they are judged to be independent), then I
can accept the proposed PR.

nigelmegitt:
The exit criteria do not say which implementations are permitted to be included in the implementation report; they merely state a requirement on those that are listed in order to transition beyond CR. I do not see any reason in principle why TTV and TTPE cannot both be listed as implementations against the same feature, within the implementation report. That is not the same thing as saying that they count as independent implementations of course.

skynavga:
@nigelmegitt in that case, I will remove my "will object..." label and approve; also, given your response, I will expect that the WG will NOT take a position on the nature of whether some implementation is independent or not, and leave that assessment to the director; however, I understand that the group (or chair) may be required to respond with an opinion if specifically asked to do so, in which case, if the group's opinion is not unanimous, I will expect that minority views be communicated to the director

PR #670, opened 2018-02-19
Comment out editor and issue notes for CR1 (#660).

### LINKED ISSUE #660: Remove (comment out in source) all remaining issue and editor's notes.
Remove (comment out in source) all remaining issue and editor's notes in preparation for CR1 pub.

skynavga:
Closes #660.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -1989,11 +1989,13 @@ with additional elements and/or attributes in order to satisfy TTML document con
 TTML documents and processors.</p>
 <div2 id="conformance-content">
 <head>Document Conformance</head>
+<!--
 <issue id="issue-ttml2-361">
 <head>Media Type Parameters</head>
 <source><loc href="https://github.com/w3c/ttml2/issues/361">https://github.com/w3c/ttml2/issues/361</loc></source>
 <p>Update step (1) below to account for to be defined <code>processorprofiles</code> media type parameter.</p>
 </issue>
+-->
 <p>A <loc href="#terms-timed-text-document-instance">timed text document instance</loc> conforms to this specification if the following criteria are
 satisfied:</p>
 <olist>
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -26438,7 +26486,9 @@ as defined by <bibref ref="xsl11"/>, &sect; 4.5, i.e., that a CSS-style line box
 </div4>
 <div4 id="derivation-luminanceGain">
 <head><loc href="#style-attribute-luminanceGain">tts:luminanceGain</loc></head>
-<ednote><edtext>Complete when/if a derivation is available, or specify "none"</edtext></ednote>
+<!--
+    <ednote><edtext>Complete when/if a derivation is available, or specify "none"</edtext></ednote>
+    -->
```
nigelmegitt: In this case the table also needs to be removed, and a sentence added, such as "`tts:luminanceGain` has no equivalent in XSL or CSS."

spec/ttml2.xml
```diff
@@ -26681,9 +26731,11 @@ The semantics of this style attribute are extended by this specification.</td>
 </div4>
 <div4 id="derivation-rubyPosition">
 <head><loc href="#style-attribute-rubyPosition">tts:rubyPosition</loc></head>
+<!--
 <ednote>
 <edtext>Complete when/if a derivation is available, or specify "none"</edtext>
-</ednote><p/>
+</ednote>
+--><p/>
```
nigelmegitt: In this example the table also needs to be completed, which I think I had asked you about @skynavga .

spec/ttml2.xml
```diff
@@ -26710,9 +26762,11 @@ defined in this specification.</td>
 </div4>
 <div4 id="derivation-rubyReserve">
 <head><loc href="#style-attribute-rubyReserve">tts:rubyReserve</loc></head>
+<!--
 <ednote>
 <edtext>Complete when/if a derivation is available, or specify "none"</edtext>
-</ednote><p/>
+</ednote>
+--><p/>
```
nigelmegitt: If there's no equivalent, the table also needs to be removed and a "no derivation" sentence adding, as per luminanceGain.

spec/ttml2.xml
skynavga: @nigelmegitt You are asking me to implement the ednote. The reason it was removed was so that CR would not have process related comments in place. My assumption is that we will address these ednotes and issues during leading up to CR2, and post CR1. Can you approve on that basis or are you insisting we address the substance of the commented out notes now?

spec/ttml2.xml
skynavga: @nigelmegitt See above comment response.

spec/ttml2.xml
skynavga: @nigelmegitt See above comment response.

spec/ttml2.xml
skynavga: @nigelmegitt p.s. this is why I didn't "remove" but only commented out the notes in the spec source code; after CR1 I will restore them

nigelmegitt: @skynavga Oh, okay, I didn't realise we have a requirement to remove ednotes from informative sections in CR, but if that's all it is, fine.

skynavga:
@nigelmegitt thanks

PR #696, opened 2018-03-09
Add CR1 upload package.

skynavga:


skynavga:
@nigelmegitt @cconcolato ping

nigelmegitt:
What is the difference between this package and https://github.com/w3c/ttml2/releases/tag/ttml2-CR-20180313-CR which already provides a zip and tar of the repo? It doesn't seem necessary to add it.

skynavga:
@nigelmegitt if you open that (release) ZIP your question will be answered: a release ZIP captures everything in the repo on the master branch

In any case, I need to update this upload package before merging since PLH needed to make a minor change to the HTML. I will update shortly.

skynavga:
@nigelmegitt updated to match published CR1 content; please approve so we can proceed with CR2 PR merges

nigelmegitt:
@skynavga When you create a release you choose which branch to use, so if the gh-pages branch is up to date with everything you need in it, you can create a release of that branch rather than master, instead of adding an additional tar file to the repo. That would be my preference since it adds less clutter.

skynavga:
@nigelmegitt except that in my mind, the release captures master, and not the artifacts one can build with master; a release of only the artifacts isn't a release IMO

skynavga:
@nigelmegitt are you going to approve this or are you asking me to do something?

skynavga:
@nigelmegitt I should add that (1) we have always created artifact package archives in both TTML1 and TTML2 for public TR releases; (2) capturing the artifacts in a GIT release in gh-pages is possible, but a bad idea, since gh-pages is transitory (you might change the build process tomorrow);

unless you wish to delay work on CR2, please approve this PR as I am not merging any CR2 PRs until this PR is merged

PR #698, opened 2018-03-15
Revert entities configuration to edcopy for preparation of CR2.

skynavga:


spec/entitiesedcopy.dtd
```diff
@@ -1,34 +1,31 @@
...
 
 <!ENTITY w3c.tr  "">	
 
 <!ENTITY document.role "editors-copy">
 
-<!ENTITY doctype "wd">
+<!-- The following is used for document .-->
+<!ENTITY document.status "W3C Candidate Recommendation">
+<!ENTITY w3c.status "CR">
+<!ENTITY doctype "cr">
```
nigelmegitt: Shouldn't this and the previous two lines be reverted to ED?

skynavga: Nope. Because this is an editor's draft of the (next) CR. The line

``<!ENTITY document.role "editors-copy">``

causes XSL to generate the appropriate Editor's Copy annotations and status.

nigelmegitt: The result is a bit inconsistent and unexpected: the top of the document says Candidate Recommendation but the SOTD section says

> This document is an editor's copy that has no official standing.

Shouldn't they match?

skynavga: No. It is not an editor's copy of an editor's copy. It is not an editor's copy of a WD. It is an editor's copy of a CR.

nigelmegitt: OK.

PR #706, opened 2018-03-28
Remove @condition from grouping element types (#704).

### LINKED ISSUE #704: Remove @condition from animation, head, layout, resources, styling, tt elements.
At present, the ``@condition`` attribute can apply conditionalization to the grouping elements ``animation``, ``head``, ``layout``, ``resources`` and ``styling``. However, only one of each of these grouping elements are permitted, and, consequently, a false condition on any of these would, respectively, remove all out-of-line animations, all out-of-line regions, all referential styling, etc., as a whole.

I am at a loss to find any concrete use case for such conditionalization scenarios to offset the complexity of supporting ``@condition`` in these contexts. Therefore, I propose we drop ``@condition`` from these element types.

skynavga:
Closes #704.

### DIFF
--- a/spec/rnc/ttml2-document.rnc
+++ b/spec/rnc/ttml2-document.rnc
@@ -11,6 +11,7 @@ TTAF.tt =
 
 TTAF.tt.attlist &=
   TTAF.Core.required.attrib.class,
+  TTAF.Core.Condition.attrib.class,
   TTAF.Parameter.attrib.class,
   TTAF.extent.attrib
 
--- a/spec/rnc/ttml2-document.rnc
+++ b/spec/rnc/ttml2-document.rnc
@@ -11,7 +11,6 @@ TTAF.tt =
 
 TTAF.tt.attlist &=
   TTAF.Core.required.attrib.class,
-  TTAF.Core.Condition.attrib.class,
   TTAF.Parameter.attrib.class,
   TTAF.extent.attrib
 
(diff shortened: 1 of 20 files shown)


spec/rnc/ttml2-document.rnc
```diff
@@ -11,6 +11,7 @@ TTAF.tt =
 
 TTAF.tt.attlist &=
   TTAF.Core.required.attrib.class,
+  TTAF.Core.Condition.attrib.class,
```
nigelmegitt: Is this supposed to be here? Unless I'm misreading, it allows the whole `tt` root element to be conditionalized...

spec/xsd/ttml2-document.xsd
```diff
@@ -18,6 +18,7 @@
   <xs:include schemaLocation="ttml2-embedded.xsd"/>
   <xs:attributeGroup name="tt.attlist">
     <xs:attributeGroup ref="tt:Core.required.attrib.class"/>
+    <xs:attributeGroup ref="tt:Core.Condition.attrib.class"/>
```
nigelmegitt: Is the root tt element sensibly conditionalizable?

skynavga: @nigelmegitt good question; I did not add ``tt`` to this list because I can imagine a use case where it might apply, e.g., having multiple ``tt`` documents that might apply in some usage context, where one is selected based on ``@condition``; OTOH, one could argue that the external usage context should make use of an external condition in such a case, and not one embedded in the document; in any case, I would be willing to remove from ``tt`` as well

nigelmegitt: @skynavga that is exactly what I would argue in this case.

skynavga: Removed.

PR #707, opened 2018-03-28
Remove @style from animate and set (#703).

### LINKED ISSUE #703: Remove @style from set and animate.
On further consideration, I now have doubts about the utility of adding ``@style`` on ``set`` and ``animate`` elements. In the case of ``set``, the ``@style`` attribute was not supported in TTML1, and, further, only a single style attribute was permitted on ``set`` in TTML1. In TTML2, we now allow multiple style attributes on a single ``set``, but adding ``@style`` seems to go to far, particularly given the disparity with multiple values on ``animate``. In the case of ``animate``, the values of targeted style attributes will consist of multiple values, i.e., as ``<animation-value-list>`` expressions. While it is straightforward to recognize the possibility of multiple values on a style attribute that appears directly on an ``animate`` element, it is not at all straightforward to support multiple values on an attribute that appears on a ``style`` element directly or indirectly referenced via ``@style`` on ``animate``. In this case, we would have some ``style`` elements with multiple value styles (for reference by ``animate`` elements) and others with single value styles (for reference by other content and ``set`` elements).

I propose that we back out the addition of ``@style`` on ``set`` and remove it from ``animate``. This will significantly reduce the complexity of implementations of animation elements without impacting functionality, since one can directly specify animated attributes on ``set`` and ``animate`` without use of ``@style``.

skynavga:
Closes #703.


skynavga:
@nigelmegitt fixed in https://github.com/w3c/ttml2/pull/707/commits/44ef7a2ed3df1a136dcff3e580d0644785eaf5ba; please re-review

PR #716, opened 2018-04-19
Change tts:luminanceGain to non-inheritable (#715).

### LINKED ISSUE #715: Inheritance of tts:luminanceGain
For the [`tts:luminanceGain` attribute](https://w3c.github.io/ttml2/#style-attribute-luminanceGain) the inheritance is set to "yes". This attribute however applies only to `tt:region`, to which it cannot be inherited from any other element (the `tt:initial` case is actually already handled separately by the [Specified Style Set Processing](https://w3c.github.io/ttml2/#semantics-style-resolution-processing-sss)).

skynavga:
Closes #715.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -10823,7 +10823,7 @@ luminance, such as 10,000 cd∙m<sup>-2</sup>.</p>
 </tr>
 <tr>
 <td><emph>Inherited:</emph></td>
-<td>yes</td>
+<td>no</td>
 </tr>
 <tr>
 <td><emph>Percentages:</emph></td>


nigelmegitt:
Built at https://rawgit.com/w3c/ttml2/issue-0715-luminance-gain-inheritance-build/index.html

palemieux:
> However, I wonder how it is supposed to work in practice - I might expect the value to vary in time on a subtitle by subtitle basis and therefore be applied to a p. That's a whole separate question though.

Using `<set>` to modulate the value?

nigelmegitt:
> Using `<set>` to modulate the value?

OK if you're willing to take the authorial pain of having the content times completely separated from the `luminanceGain` times. Maybe that's okay, if the variation of `luminanceGain` over time can be algorithmically calculated base only on some processing of the video without reference to the subtitle text.

PR #731, opened 2018-05-11
Refactor semantics of #region-inline-implicit (#727).

### LINKED ISSUE #727: Definition of implicit inline region
The [feature `#region-inline-implicit`](https://www.w3.org/TR/ttml2/#feature-region-inline-implicit) refers to the definition of the `tt:region` element ([11.1.2](https://w3c.github.io/ttml2/#layout-vocabulary-region)). However it is defined neither in 11.1.2 nor anywhere else in chapter 11. The only definition I can find is in [2.2](https://w3c.github.io/ttml2/#terms) as part of the "inline region" term definition.

skynavga:
Closes #727.


nigelmegitt:
Built at https://rawgit.com/w3c/ttml2/issue-0727-inline-region-build/index.html

spec/ttml2.xml
```diff
@@ -21149,13 +21149,13 @@ for explicit inline region element children of a <loc href="#terms-content-eleme
 <div3 id="feature-region-inline-implicit">
 <head>#region-inline-implicit</head>
 <p>A TTML <loc href="#terms-transformation-processor">transformation processor</loc> supports the
-<code>#region-inline-implicit</code> feature if it recognizes and is capable
-of transforming implicit inline region specifications on a <loc href="#terms-content-element">content element</loc>, as further defined by
-<loc href="#layout-vocabulary-region"><code>region</code></loc>.</p>
+<code>#region-inline-implicit</code> feature if it recognizes and is capable of transforming the style attributes on a
+<loc href="#content-vocabulary-div"><el>div</el></loc> or <loc href="#content-vocabulary-p"><el>p</el></loc> element that apply to an implicit inline region
```
nigelmegitt: This is an improvement but the wording does not wholly avoid the difficulty reported in the issue that `implicit inline region` is not defined. I would suggest changing 

`that apply to an implicit inline region in accordance with`

to:

`that are considered equivalent to an anonymous inline animation of the applicable region as defined by`

and "inline animation" can be wrapped in a reference to its term definition.

This language mirrors the language in 11.1.2.1 Special Inline Animation Semantics.

An alternative briefer form that might also work would be:

`that result in inline animation semantics of the region as defined by`

spec/ttml2.xml
```diff
@@ -21149,13 +21149,13 @@ for explicit inline region element children of a <loc href="#terms-content-eleme
...
+in accordance with <specref ref="layout-vocabulary-region-special-inline-animation-semantics"/>.</p>
 <p>A TTML <loc href="#terms-presentation-processor">presentation processor</loc> supports the
-<code>#region-inline-implicit</code> feature if it implements presentation semantic support
-for implicit inline region specifications on a <loc href="#terms-content-element">content element</loc>, as further defined by
-<loc href="#layout-vocabulary-region"><code>region</code></loc>.</p>
+<code>#region-inline-implicit</code> feature if it implements presentation semantic support for the style attributes on a
+<loc href="#content-vocabulary-div"><el>div</el></loc> or <loc href="#content-vocabulary-p"><el>p</el></loc> element that apply to an implicit inline region
```
nigelmegitt: Same comment as above on line 21153.

skynavga:
Actually, they (the properties) do apply to an implied region, namely, the region implied by the cited special inline animation semantics. I don't want to change the name of the feature designator, as that would make it overly asymmetrical with respect to `#region-inline-explicit`. I've marked #727 for the agenda so we can resolve this matter.

nigelmegitt:
@skynavga ok let's discuss on Thursday's call. 

My position at the moment is that I don't see how the implied region is any more significant in being implied than is the region that applies to any `p` element whose ancestor `div` or `body` specifies a `region` attribute. In particular the behaviour is so different to that of explicit region children of an element that I think it warrants a different name. Making them symmetrical appears to strongly suggest a similarity that does not exist: one _specifies_ a region, the other _modifies_ a region.

spec/ttml2.xml
```diff
@@ -21126,35 +21126,26 @@ for the same vocabulary specified above.</p>
...
-<div3 id="feature-region-inline-explicit">
-<head>#region-inline-explicit</head>
+<div3 id="feature-region-implied-animation">
+<head>#region-implied-animation</head>
 <p>A TTML <loc href="#terms-transformation-processor">transformation processor</loc> supports the
-<code>#region-inline-explicit</code> feature if it recognizes and is capable
-of transforming explicit inline region element children of a <loc href="#terms-content-element">content element</loc>, as further defined by
-<loc href="#layout-vocabulary-region"><code>region</code></loc>.</p>
+<code>#region-implied-animation</code> feature if it recognizes and is capable of transforming the style attributes on a
+<loc href="#content-vocabulary-div"><el>div</el></loc> or <loc href="#content-vocabulary-p"><el>p</el></loc> element that apply to an implied region
```
nigelmegitt: Rather than "implied region" it would be more consistent with the referenced inline animation semantics text to use "associated region".

spec/ttml2.xml
```diff
@@ -21126,35 +21126,26 @@ for the same vocabulary specified above.</p>
...
+in accordance with <specref ref="layout-vocabulary-region-special-inline-animation-semantics"/>.</p>
 <p>A TTML <loc href="#terms-presentation-processor">presentation processor</loc> supports the
-<code>#region-inline-explicit</code> feature if it implements presentation semantic support
-for explicit inline region element children of a <loc href="#terms-content-element">content element</loc>, as further defined by
-<loc href="#layout-vocabulary-region"><code>region</code></loc>.</p>
+<code>#region-implied-animation</code> feature if it implements presentation semantic support for the style attributes on a
+<loc href="#content-vocabulary-div"><el>div</el></loc> or <loc href="#content-vocabulary-p"><el>p</el></loc> element that apply to an implied region
```
nigelmegitt: Same comment - propose changing "implied" to "associated".

Issue #722, opened 2018-04-28
Derivation of tts:textShadow incorrect


palemieux:
The derivation of `tts:textShadow` states that _values map directly_ to CSS `text-shadow`, however the values of `tts:textShadow` are given relative to the block and inline progression directions whereas those of `text-shadow` are relative to the horizontal and vertical directions.

palemieux:
Mooted if #723 is resolved per https://github.com/w3c/ttml2/issues/723#issuecomment-386327663

PR #732, opened 2018-05-11
Account for use of auto keyword in derivation of tts:textShadow (#726).

### LINKED ISSUE #726: Derivation of tts:textEmphasis incorrect
[CSS Text Decoration], § 3 does not support the value `auto`.

skynavga:
Closes #726.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -27528,7 +27528,7 @@ determining the position of underlines applies.</td>
 </tr>
 <tr>
 <td><phrase role="strong">Values</phrase></td>
-<td>Values map directly</td>
+<td>Except for <code>auto</code>, values map directly. The semantics of <code>auto</code> are writing mode dependent.</td>
 </tr>
 <tr>
 <td><phrase role="strong">Notes</phrase></td>


palemieux:
@skynavga  Issue #726 is about tts:textEmphasis and not tts:textShadow.

spec/ttml2.xml
```diff
@@ -27528,7 +27528,7 @@ determining the position of underlines applies.</td>
 </tr>
 <tr>
 <td><phrase role="strong">Values</phrase></td>
-<td>Values map directly</td>
+<td>Except for <code>auto</code>, values map directly. The semantics of <code>auto</code> are writing mode dependent.</td>
 </tr>
```
palemieux: CSS3 provides the following defaults semantics, which seem to alleviate the need for 'auto':

_If a shape keyword is specified but neither of ‘filled’ nor ‘open’ is specified, ‘filled’ is assumed. If only ‘filled’ or ‘open’ is specified, the shape keyword computes to ‘circle’ in horizontal writing mode and ‘sesame’ in vertical writing mode._

https://www.w3.org/TR/css-text-decor-3/#text-emphasis-style-property

skynavga: This issue is about derivation, not whether `auto` is defined. The definition of `auto` is based on an (apparently) earlier CSS draft (I did not invent it). The language you indicate above still requires specifying filled or open.

nigelmegitt:
Build available at https://rawgit.com/w3c/ttml2/issue-0726-text-emphasis-derivation-build/index.html

nigelmegitt:
I think this is supposed to close #722 not #726.

nigelmegitt:
#726 is about `tts:textEmphasis` derivation and this pull request does modify that derivation - looks like the title of this pull request is wrong and should refer to textEmphasis.

spec/ttml2.xml
nigelmegitt: @palemieux that language suggests a different change, to make `(filled | open)` optional, with a default of `filled`, but the `auto` keyword effectively selects between `sesame` and `circle` based on writing direction, which is not present in the CSS3 specification at all.

I agree with @skynavga , #726 is about the derivation being incorrect, due to the absence of `auto` from the CSS 3 spec. I'm surprised that @skynavga refers to a previous CSS draft though because the CSS3 Text Decoration Module CR was published without `auto` in 2013, and our work here to introduce text emphasis post-dates that by some years.

PR #733, opened 2018-05-11
Specify fallback semantics for tts:textCombine (#725).

### LINKED ISSUE #725: Fallback for tts:textCombine="digits".
What should the fallback be if `tts:textCombine="all"` is supported, but `tts:textCombine="digits"` is not?

In particular CSS support for `text-combine-upright:digits` is not [universal](https://developer.mozilla.org/en-US/docs/Web/CSS/text-combine-upright).

Alternatively, should `tts:textCombine="digits"` be at-risk?

skynavga:
Closes #725.


nigelmegitt:
Build available at https://rawgit.com/w3c/ttml2/issue-0725-text-combine-fallback-build/index.html

PR #739, opened 2018-05-12
Add #length-root-container-relative feature (#718).

### LINKED ISSUE #718: Add TTML feature(s) for rw/rh <length> units
The `rw` and `rh` units can be used in [`<length>` expressions](https://w3c.github.io/ttml2/#style-value-length) and were added by TTML2. However they are currently not covered by a respective TTML feature - in contrast to the other allowed units of `<length>` (e.g. `#length-em` for the `em` unit).

skynavga:
Closes #718.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -14725,7 +14725,8 @@ applies to the element in which context the <code>em</code> unit is resolved.</p
 <p>The semantics of the unit of measure <code>c</code> (cell) are
 defined by the parameter <specref
 ref="parameter-attribute-cellResolution"/>.</p>
-<p>The units of measure <code>rw</code> and <code>rh</code> are defined as
+<p>The <termdef id="defs-root-container-relative" term="root container relative"><term>root container relative</term></termdef>
+units of measure <code>rw</code> and <code>rh</code> are defined as
 one percent (1%) of the width and height of the <loc href="#terms-root-container-region">root container region</loc>, respectively.
 For example, the length <code>25rw</code> is interpreted as 25% of the width of
 the <loc href="#terms-root-container-region">root container region</loc>.</p>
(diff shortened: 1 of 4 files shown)


spec/ttml2.xml
```diff
@@ -20448,8 +20449,10 @@ style value expression.</p>
 feature if it implements presentation semantic support for all defined values of
```
nigelmegitt: "all defined values" isn't right. I don't understand why it says all here and then the next paragraph contradicts that by stating the exclusions, with the informative note providing the inclusion list. Why not define the `#length` feature in terms of the inclusion list and remove the note? In particular, this formulation is sensitive to new length units being added in the future, which will require further exceptions to be specified, when for backwards compatibility reasons we will always want to freeze the definition of `#length` as per the TTML1 definition.

spec/ttml2.xml
```diff
@@ -20593,7 +20597,7 @@ href="#style-value-length">&lt;length&gt;</loc> style value
 expression.
 </p>
 <note role="clarification">
-<p>Support for <code>#length-positive</code> is intended to imply support for zero valued
+<p>Support for <code>#length-positive</code> implies support for zero valued
```
nigelmegitt: Just noting that, though this change is fine, it isn't needed to resolve #718.

spec/ttml2.xml
```diff
@@ -20618,7 +20622,7 @@ href="#style-value-length">&lt;length&gt;</loc> style value
 expression.
 </p>
 <note role="clarification">
-<p>Support for <code>#length-real</code> is intended to imply support for
+<p>Support for <code>#length-real</code> implies support for
```
nigelmegitt: Just noting that, though this change is fine, it isn't needed to resolve #718.

skynavga:
Need to update this PR based upon merging of #794 to master.

skynavga:
@nigelmegitt need your approval on #794 to proceed on this PR

skynavga:
@nigelmegitt please re-review post merges;  your prior comments were no longer relevant due to merge changes

PR #740, opened 2018-05-12
Clarify application of ruby alignment and positioning (#717).

### LINKED ISSUE #717: Applicability of tts:rubyPosition and tts:rubyAlign.
The `tts:rubyPosition` attribute applies to "span only if the computed value of tts:ruby is container". IMO this does not really make sense, as the ruby container is just a general container for text with ruby annotations (it may work though in the simple ruby case).

The `tt:span` that is actually affected corresponds to the ruby text content or instead (if present) the ruby text container. So the attribute should instead apply to a `tt:span` only if the value of `tts:ruby` is `text` (only in case `tt:span` is not a child of a ruby text container - as all ruby text content of the same ruby text container shall have the same position) or `textContainer`.

The current wording also does not work if double-sided ruby is used, as when relevant on the ruby container only, it would not be possible to specify the (different) positions in case _two_ text containers are present. The [double-sided ruby example next to the definition of `tts:ruby`](https://w3c.github.io/ttml2/#style-attribute-ruby) is a good example for this issue. It uses `tts:rubyPosition` on ruby text container level - where it cannot be applied according to the current wording.

skynavga:
Closes #717.


spec/ttml2.xml
```diff
@@ -11886,6 +11886,14 @@ base text is in any of the following character classes as defined by <bibref ref
 <p>See <bibref ref="jlreq"/>, &sect;3.3.8, <emph>Adjustments of Ruby with Length Longer than that of the Base Characters</emph> for
 further information.</p>
 </note>
+<p>When performing presentation processing,
+<termdef id="defs-isolated-ruby-content" term="isolated ruby content"><term>isolated ruby content</term></termdef>,
+namely, ruby base content or ruby text content that is specified without a ruby base container or ruby text container, respectively,
```
spoeschel: "without a ruby base container or ruby text container" - I think we should say "an enclosing" or "a parent" after "without" here

spoeschel:
Except for the commented text passage, this fix is fine for me.

spec/ttml2.xml
```diff
@@ -11886,6 +11886,14 @@ base text is in any of the following character classes as defined by <bibref ref
...
+is processed in such a manner as to imply the presence of an anonymous ruby base container and (or) anonymous ruby text container as required.
+An <termdef id="defs-implied-ruby-base-container" term=""><term>implied ruby base container</term></termdef> and (or)
```
palemieux: > is processed in such a manner as to imply the presence of an anonymous ruby base container and (or) anonymous ruby text container as required.

This significantly complicates processing. Why is it allowed?

skynavga: @palemieux I'm not sure what you are asking. Are you asking why specifying text and base content without specifying a text container and base container? If so, then this comes from the original Ruby content model [1] adopted in XHTML which served as a basis for the HTML5 ruby support as well. Technically speaking, we don't have to support this minimization feature (in which case we would always require base and text containers.

It is not overly complicated to implement. TTPE supports it.

[1] https://www.w3.org/TR/2001/REC-ruby-20010531/

palemieux:
Built at https://rawgit.com/w3c/ttml2/issue-0717-ruby-styling-applicability-build/index.html

spec/ttml2.xml
palemieux: @skynavga Ah right. Yes, text and base elements should be able to occur without parent text and base containers. I had misread the text.

However, for the application of which styles would an _implied ruby base container_ and (or) an _implied ruby text container_ be required?

skynavga: Well, all of the default or fallback values of all the styles that would apply have applied to the spans had the text container or base container been present. For example, rubyAlign says

>The tts:rubyAlign attribute is used to specify the position of ruby text within the inline area generated by the ruby text container annotation.

This implies there is an inline area that contains all of the glyph areas of a text container in which space distribution occurs in order to obtain alignment.

As for base container, in the context of CSS3 Ruby and JLREQ, the glyph areas of a base container may be distributed with extra whitespace in order to handle overhang constraints. For example, the default behavior regarding [overhang semantics](https://w3c.github.io/ttml2/index.html#d3e21895) may be accommodated by either (or both) distribution of space in inline areas generated by a text container and base container (whether explicitly specified or implied).

skynavga:
@palemieux I'd like to wrap this up; is there something you want me to change?

skynavga:
@palemieux thanks

PR #741, opened 2018-05-13
Clarify explanation of percentage edge offset in tts:position example…

### LINKED ISSUE #712: Incorrect description of tts:position example.
The bottom of the region seems to be positioned `10% * (root container region height)` from the bottom of the root container region.

This is inconsistent with the clause _percentage value component for a vertical position offset is relative to the height of the positioning rectangle minus the height of the associated region._

skynavga:
… (#712).

Closes #712.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -11443,7 +11443,8 @@ the computed position and the supported position is minimized. If there are mult
 the computed value, then the value least distant from [0,0], i.e., closest to the coordinate space position, is used.</p>
 </note>
 <p>The <att>tts:position</att> style is illustrated by the following example, which positions a region so that it is
-centered in the horizontal dimension and has a bottom edge 10% above the bottom of the positioning rectangle in the vertical dimension.</p>
+centered in the horizontal dimension and has a bottom edge offset at 10% above the bottom of the positioning rectangle in the vertical dimension,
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -11443,7 +11443,8 @@ the computed position and the supported position is minimized. If there are mult
 the computed value, then the value least distant from [0,0], i.e., closest to the coordinate space position, is used.</p>
 </note>
 <p>The <att>tts:position</att> style is illustrated by the following example, which positions a region so that it is
-centered in the horizontal dimension and has a bottom edge 10% above the bottom of the positioning rectangle in the vertical dimension.</p>
+centered in the horizontal dimension and has a bottom edge offset at 10% above the bottom of the positioning rectangle in the vertical dimension,
+where this offset is equal to 10% of the height of the positioning rectangle minus the height of the region.</p>
```
nigelmegitt: This is ambiguous: is it 1. `(10% of height of positioning rectangle) - height of region` or 2. `10% of (height of positioning rectangle - height of region)`?

palemieux: I believe the tests at https://github.com/w3c/imsc-tests/pull/49 reflect (2).

nigelmegitt: OK, so the text should reflect that - since the normal mathematical precedence that you'd understand from the text is (1) it needs to be edited to clarify.

PR #742, opened 2018-05-14
Exclude implied animation of region via tts:origin on div and p (#710).

### LINKED ISSUE #710: Update #origin for TTML1 compatibility.
`tts:origin` now applies to `p` and `div` but `#origin` feature has not been refactored such that:

- `#origin` means support for `tts:origin` on `region`
- `#origin-block` means support for `tts:origin` on `p` and `div`

skynavga:
Closes #710.

Note that ``tts:origin`` does not directly apply to ``div`` and ``p``; rather, it applies to ``region`` via an implicit redirection (of styles) on ``div`` and ``p``.


spec/ttml2.xml
```diff
@@ -20810,6 +20810,11 @@ attribute.</p>
 the <loc
 href="#style-attribute-origin"><att>tts:origin</att></loc>
 attribute.</p>
+<p>Notwithstanding the above, support for the <code>#origin</code> feature does not require support for the
```
nigelmegitt: I don't like this formulation - it seems confusing to contradict the definition immediately after it. I would prefer to qualify the definition in the first place, for example by changing:

> supports the `#origin` feature if it recognizes and is capable of transforming the `tts:origin` attribute.

to:

> supports the `#origin` feature if it recognizes and is capable of transforming the `tts:origin` attribute **as applied to the `region` element**.

skynavga: We use the same formulation throughout in the same manner, so your suggestion would create an inconsistency.

skynavga:
Closing without merge as the changes proposed here have been superseded by #794.

PR #744, opened 2018-05-14
Refactor extent features (#693, #730).

### LINKED ISSUE #693: Refine #extent-version-2 feature designator.
`#extent-version-2` feature is unusual as it both permits subordinate features and permit new attribute vocabulary.

This makes it difficult for profiles such as IMSC 1.1 to prohibit the new attribute vocabulary, without also prohibiting subordinate features.

Suggest factoring out the new attribute vocabulary into a `#extent-attribute-values-version-2` feature designator.

### LINKED ISSUE #730: Feature #extent-measure also affects <length>
When someone wants to set the `tts:extent` attribute to a `<length>` value, support of the [`#extent-measure` feature](https://w3c.github.io/ttml2/#feature-extent-measure) is required. The reason is that one can only get to `<length>` through `<measure>` here. 

As IMSC 1.1 currently prohibits the mentioned feature, thus also the usage of `<length>` is prohibited!

So it should be possible to allow the usage of `<length>` and prohibit the other values of `<measure>` at the same time. To address this e.g. the `#extent-measure` feature could be splitted somehow.

skynavga:
Closes #693.
Closes #730.


skynavga:
Additional refinements (from https://github.com/w3c/ttml2/pull/744/commits/036921b37a45371a98af52341bf43643d1657795) include:

- recast `#extent`, `#extent-region` and `#extent-root` in terms of TTML1 references
- improve definition of `#extent-image`
- rename `#extent-version-2-full` to `#extent-full-version-2`

Issue #697, opened 2018-03-13
Negative feature designators aren't future proof.


nigelmegitt:
The way that non-support inverse style feature designators like `#textEmphasis-no-color` and `#textEmphasis-no-quoted-string` are defined is open ended. Rather than requiring support for a specific subset of syntax+semantics, they require for support for everything except some syntax+semantic, in an open ended way.

This means that if in the future an additional syntax+semantic is added to the value set of `tts:textEmphasis`, then the meaning of those features would change by default.

It also presents a bit of a logic puzzle, to say the least, when interpreting `mostRestrictive` and `leastRestrictive` with respect to negative features.

To avoid this, I propose we refactor those features so that we only have positive definitions, i.e. support for a bounded set of syntax and semantics.

nigelmegitt:
OK let's discuss. IMO this is very badly broken in TTML1 and TTML2 and and we should fix it in TTML2, by:

1. deprecating negatively expressed feature designators inherited from TTML1.
2. adding new positively expressed feature designators into TTML2.
3. specifying that new feature or extension designators should be positively expressed.

skynavga:
I do not agree with your assessment (that there is a bug and that it needs a change now). Let's wait until we can discuss.

skynavga:
The only way I can see addressing this is to go through every feature in TTML1 that is marked as first defined by Version 1 (TTML1) and add a qualifier of the form "as defined or equivalent to the definition specified by [TTML1]". Doing so, we could still add "notwithstanding" notes warning the reader about exclusions, but those notes wouldn't be part of the normative definition of the feature.

The alternative you suggest (create a whitelist for every feature) would be very time consuming and prone to error, particularly since it would involve repeating the formal definition of the syntax or semantics found in the body of the specs, a task we certainly cannot under take.

nigelmegitt:
Meeting today: Nigel to come up with concrete example of difficulty doing feature combination.

nigelmegitt:
This issue was discussed by the working group today, and minuted at https://www.w3.org/2018/05/17-tt-minutes.html#item11

**RESOLUTION**: Mark TTML1 features as relating to the TTML1 definition

**SUMMARY**: This resolution is logged as #763

skynavga:
In order to reduce the likelihood of dealing with this, I will break out the semantics of profile combination into separate features (as an additional update to PR #747 (issue #687).

skynavga:
[...]

PR #745, opened 2018-05-14
Fix placement typo (wrong location of exclusion language) (#688).

### LINKED ISSUE #688: Exclude #rubyAlign-withBase from #rubyAlign.
`tts:rubyAlign-withBase` attribute does not exist

It is also not clear if `#rubyAlign` includes support for the value `withBase`

Suggest that `#rubyAlign` exclude support for value `withBase`

skynavga:
Closes #688.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -21223,6 +21223,8 @@ transforming the <loc href="#style-attribute-rubyAlign"><att>tts:rubyAlign</att>
 <p>A TTML <loc href="#terms-presentation-processor">presentation processor</loc> supports the
 <code>#rubyAlign</code> feature if it implements presentation semantic support for the
 <loc href="#style-attribute-rubyAlign"><att>tts:rubyAlign</att></loc> attribute.</p>
+<p>Notwithstanding the above, support for the <code>#rubyAlign</code> feature does not require support for the
+<loc href="#feature-rubyAlign-withBase"><code>#rubyAlign-withBase</code></loc> feature.</p>
 </div3>
 <div3 id="feature-rubyAlign-withBase">
 <head>#rubyAlign-withBase</head>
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -21223,6 +21223,8 @@ transforming the <loc href="#style-attribute-rubyAlign"><att>tts:rubyAlign</att>
 <p>A TTML <loc href="#terms-presentation-processor">presentation processor</loc> supports the
 <code>#rubyAlign</code> feature if it implements presentation semantic support for the
 <loc href="#style-attribute-rubyAlign"><att>tts:rubyAlign</att></loc> attribute.</p>
```
nigelmegitt: Rather than the `Notwithstanding the above` sentence that follows, I would prefer to explicitly state the supported syntax for this feature, i.e. `"start" | "center" | "end" | "spaceAround" | "spaceBetween"` values of `tts:rubyAlign`.

Then the "Notwithstanding the above," that follows should be changed to "Note that".

skynavga:
@nigelmegitt you are disrupting a well established norm (a formulaic usage) at the end of a long process; there is no justification for insisting on this change on this particular issue; furthermore, you are conflating this issue with your CRUSADE (#697) to refactor a large number of feature definitions; finally, you are ignoring that there are already 11 features that use this same language, and this change would need to apply to all of them; please approve this issue and then either open a new issue to cover all (11+1) uses or add this into your CRUSADE (#697)

nigelmegitt:
@skynavga this is a new addition, so the correct place to author it right is here in the pull request in which it is being introduced. I'm not insisting on all usages of "Notwithstanding the above..." being removed (I've found at least one that is entirely reasonable), but I _am_ insisting that we do not introduce new language that will reproduce for us in the future a problem we have now.

skynavga:
@nigelmegitt I am absolutely not going to positively define a feature by enumerating all of their syntactic and semantic components. Doing so would be complete folly.

nigelmegitt:
Given #697 discussion today I'll attempt to generate a concrete example where profile combination causes difficulties here.

skynavga:
Closing without merging in order to handle via new PR that addresses #793 et al.

PR #750, opened 2018-05-16
Qualify use of 'line area' (#749).

### LINKED ISSUE #749: Definition of "last affected line area" in <annotation-position>.
The definition of `<annotation-position>` should clarify what "last affected line area" means. Does it mean "last line area of the region", or something else?

skynavga:
Closes #749.


skynavga:
@palemieux see new clarification attempt; note also that only the last block area generated by a paragraph contains a _last line area_; it is not well-formed to say the last line of a `div` or `region`; also, I removed the qualifier _affected_ since the semantics here (whether equivalent to before or after) are not related to whether a line area is affected by (subjected to) an annontation

palemieux:
@skynavga Ok. This is much more precise. Why "the last block area" in "the last line area of the last block area"? Can a `<p>` contains multiple blocks in TTML2?

skynavga:
@palemieux regarding

>Can a `<p>` contains multiple blocks in TTML2?

In the general XSL formatting model yes, but, at present in TTML, no, since TTML doesn't have the concept of a page break (or region break). However, in the future, we could have a paragraph flow into a series of connected regions, in which last block area is pertinent.

Note also that the new language (here) matches the language already found in the note at the end of [**10.2.39 tts:textAlign**](https://w3c.github.io/ttml2/index.html#style-attribute-textAlign).

palemieux:
@nigelmegitt @skynavga This is not editorial, BTW

skynavga:
@palemieux there is definitely no impact on conformance, so I would have to disagree; making a change to normative text, for example, to do a better job of explaining something already implied, does not mean it is a substantive change

nigelmegitt:
Build available at https://rawgit.com/w3c/ttml2/issue-0749-last-line-area-build/index.html

spec/ttml2.xml
```diff
@@ -14108,8 +14108,8 @@ bottom of the <loc href="#terms-glyph-area">glyph areas</loc>. If a vertical wri
 <gitem>
 <label><code>outside</code></label>
 <def>
-<p>Equivalent to <code>before</code> for all but the last affected line area; otherwise, equivalent to <code>after</code>
-for the last affected line area.</p>
+<p>Equivalent to <code>before</code> for all but the last line area of the last block area generated by
+a <loc href="#content-vocabulary-p"><el>p</el></loc> element which contains annotated text; otherwise, equivalent to <code>after</code>.</p>
```
nigelmegitt: I see that `<annotation-position>` is only ever used as syntax for an attribute that applies to `span` elements. Should we be explicitly relating the `p` element mentioned here to the `span` on which the style attribute is applied?

skynavga: But it cannot be interpreted as related to any other `p`, so what is the point in saying it. That would be like saying _the impossible is impossible_.

nigelmegitt: I don't agree that it cannot be interpreted as related to any other `p` @skynavga, but I'm not motivated to argue this.

PR #755, opened 2018-05-16
Describe generically compliant processors that support only external …

### LINKED ISSUE #683: Profile related features should be optional
In the transformation and processor profiles, the following feature designators should be marked as optional rather than required:

* `#contentProfiles`
* `#inferProcessorProfile`
* `#permitFeatureNarrowing`
* `#permitFeatureWidening`
* `#profile-version-2`

Rationale: It is reasonable for processors to support TTML2 in processing contexts where profile processing is not required and to omit support for those features.

skynavga:
…profiles (#683).

Closes #683.


skynavga:
@nigelmegitt please re-check

PR #770, opened 2018-05-18
Add features for standard PNG images (#694).

### LINKED ISSUE #694: Use of PNG image format.
There are 16 uses of "PNG" in examples, yet it is not mentioned in the text and there is no citation to its definition.  Is PNG preferred?  We should probably at least add an informative reference.

### LINKED ISSUE #695: Use of HDR images.
Although we defined tts:lumnanceGain to assist composition of text onto absolute luminance HDR video, we have not added any support for HDR images, which seems like an omission (if we are going to include HDR).  In the case of PNG, we should consider recommending the HDR extensions defined in https://w3c.github.io/png-hdr-pq

skynavga:
Closes #694.


skynavga:
We can't make the PNG HDR PQ a non-normative reference and use it in a normative feature definition. Unless there is a policy prohibition on a normative reference to a WG Note, then there is no problem.

I recollect that 10 years ago or so that we did not use normative references to WG Notes, but I'm not sure that was a written policy, or whether it applies today (I suspect not). However, you might want to check with @plehegar.

nigelmegitt:
We could change the feature definition to be one that requires generic support for PQ HDR images in PNG, and informatively reference the WG Note as a "for example".

palemieux:
> We could change the feature definition to be one that requires generic support for PQ HDR images in PNG, 

No "generic" support exists as far as I know.

I would simply remove the `#image-png-hdr-pq` feature and add an informative note under `#image-png` pointing to the Note.

palemieux:
> I don't think that matters. There may only be one mechanism we know of for creating PQ HDR PNG images,
> but in principle there could be others.

I do not know how one can test for "generic" conformance.

nigelmegitt:
> I do not know how one can test for "generic" conformance.

The conformance claim would be "I have an implementation that can process PQ HDR PNG images, using scheme `my_great_pq_scheme`, and when I provide an image using this scheme, it processes correctly. Since this is an example of a PQ HDR PNG image, this implementation therefore supports the `#image-png-hdr-pq` feature.

I think it would be reasonable to argue that this isn't very useful, but it is logically _possible_.

Seems like we have agreement (at least me and @palemieux ) to remove `#image-png-hdr-pq`.

skynavga:
@nigelmegitt @palemieux if we remove this feature, then there is no normative way to test luminanceGain; and we can't define this feature without a normative reference to a concrete format; I think we should keep the feature and reference PAL's document

palemieux:
> and we can't define this feature without a normative reference to a concrete format;

@skynavga Can you clarify what you have in mind? `tts:luminanceGain` applies to both text and image content, so I do not see why removing a specific HDR image format would make testing `tts:luminanceGain` impossible?

skynavga:
@nigelmegitt ok, i'll take your word for it; but I still don't see why we shouldn't go with the current feature as defined; that is, unless you are confident nobody will implement support for it before PR; if that is the case, then I will remove forthwith

nigelmegitt:
@skynavga I've lost track of what you're taking my word for. I think the main points right now are:

1. We don't need a feature in TTML2 for PQ HDR PNG images.
2. If we did have such a feature then we might have a problem with a normative reference to a WG Note (which I'm pretty sure we are Not Supposed To Do).
3. `tts:luminanceGain` is orthogonal to HDR images.

> `tts:luminanceGain` applies to both text and image content

@palemieux I find this to be a surprising statement, but I can see that it's true. It concerns me that a PQ image in a region might have some kind of luminance modification treatment applied to it twice, once by the data in the image and another time by `tts:luminanceGain`. That seems likely to be a Bad Thing. Perhaps it is useful in the case of SDR images though.

skynavga:
@nigelmegitt sorry, was reading PAL's comment as yours (about luminance gain applying to text); but, since we have it applying to region, then it should apply to the region as a whole, which I suspect does make it possible to apply twice for HDR images

I think you should not assume that we can't normatively reference a WG Note. Please check with PLH. Things have gotten significantly looser with the publishing of HTML5.

palemieux:
> I think you should not assume that we can't normatively reference a WG Note. Please check with PLH. Things have gotten significantly looser with the publishing of HTML5.

It would be good to know before making a final decision on this.

The Note could always be moved to the Rec Track after making sure there is intent to implement out there.

nigelmegitt:
@plehegar would appreciate your view on https://github.com/w3c/ttml2/pull/770#issuecomment-390710256 please - can we normatively reference a WG Note from a Rec?

nigelmegitt:
We should revisit this in the light of #810.

skynavga:
@palemieux @nigelmegitt removed ``#image-png-hdr-pq`` feature from this PR; please approve

PR #771, opened 2018-05-18
Clarify use of multiple specifications of initial value for the same …

### LINKED ISSUE #764: Property specified more than once on tt:initial
The same property could be specified more than once on different `tt:initial` elements at the moment, e.g.:

```xml
...
<head>
  <styling>
    <initial tts:color="yellow"/>
    <initial tts:color="lime"/>
  <styling>
<head>
...
```

The [Specified Style Set Processing in 10.5.4.2](https://w3c.github.io/ttml2/#semantics-style-resolution-processing-sss) in step 6.c....i currently says:

"if an initial element defines the initial value for P, then use that value"

So the expected behaviour in such a case should probably be deterministic. Or instead it should be forbidden that at the same time two `tt:initial` elements that set the same property can apply (e.g. with appropriate conditions added, which exclude each other, the above example would be fine).

skynavga:
…property (#764).

Closes #764.


spec/ttml2.xml
```diff
@@ -7783,6 +7783,8 @@ the specification defined initial value(s).</p>
 </tr>
 </tbody>
 </table>
+<p>If more than one <el>initial</el> element specifies an initial value for the same style property, then
```
spoeschel: I would add the word "applicable" i.e. "If more than one applicable ...", because due to the specified `condition`, a present `tt:initial` element may not apply at all.

skynavga: No, as that will lead to confusion. We don't everywhere qualify the use of an element in terms of whether its semantics are conditionally excluded. Furthermore, we use the word "applied" with style properties for an entirely different purpose (unrelated to conditionalization).

spoeschel: OK, that makes sense. In that case I'm fine with the current wording.

PR #772, opened 2018-05-18
Add explicit exceptions to ignoring element semantics with condition …

### LINKED ISSUE #583: Prune elements whose condition evaluates to false?
Where are presentation elements pruned if their condition evaluates to false? I propose adding a pruning condition to [construct intermediate document]:

_they are a _presentation related element_ associated with a _condition_ that evaluates to false_

This needs to avoid pruning, say, `style` elements whose condition evaluates to false, because that would break referential styling (a style whose condition is false may reference a style whose condition is true).

To be sure that is the case, we need to modify the term definition of [presentation related element] to explicitly exclude styling elements. This appears to be a harmless change right now.

skynavga:
…processing (#583).

Closes #583.


spec/ttml2.xml
```diff
@@ -5971,6 +5971,16 @@ attribute that evaluates to <code>false</code>, then the content of that element
 <p>In certain cases, ignoring (excluding) the semantics of an element by means of a <att>condition</att> attribute consists
 of partially rather than wholly ignoring the semantics of the element, about which see specific element definitions.</p>
 </note>
+
+<p>Unless a <loc href="#terms-higher-level-protocol">higher level protocol</loc> applies, the following exceptions apply for the
+purpose of determining which semantics are to be ignored when a <loc href="#content-value-condition">&lt;condition&gt;</loc> expression
+value evaluates to <code>false</code>:</p>
+<ulist>
+<item><p>the <loc href="#semantics-timing">timing semantics</loc> of a <loc href="#terms-content-element">content element</loc> are not ignored;</p></item>
```
nigelmegitt: Content elements exclude `region` but `@condition` can be applied to `region`. I think this needs to be generalised to all elements that have timing semantics.

skynavga: Fixed in https://github.com/w3c/ttml2/pull/772/commits/cbd9ce8730c140a56c8aa786935f81a4612a236b.

skynavga:
@nigelmegitt implemented actions from SUMMARY above; should be good to go

PR #774, opened 2018-05-18
Make tts:ruby non-animatable (#773).

### LINKED ISSUE #773: Remove animation from tts:ruby.
The `tts:ruby` attribute is used to reflect the hierarchy of the different ruby elements (`ruby`, `rb` etc.) from the [Ruby Annotation spec](https://www.w3.org/TR/2001/REC-ruby-20010531/) while keeping the present TTML content element hierarchy, as the attribute is always applied to `tt:span`. The attribute is used to express a fixed relation between different parent/children/neighbour nodes. So in my opinion it is not feasible to flag this attribute as animatable, as the position within the ruby hierarchy is always fixed.

The only decent meaningful case I could think of is a discrete animation whereby the `tts:ruby` value of a ruby container and all of its children is changed from the respective original value to `none` at the same time. However I can't see why anyone would want to do that.

So I propose to make `tts:ruby` non-animatable.

skynavga:
Closes #773.


skynavga:
@nigelmegitt request approval for early merger at upcoming meeting

PR #795, opened 2018-05-24
Editorial improvement and typo fix for ruby constraints (#786, #787, …

### LINKED ISSUE #786: Fix typos and improve readability of ruby nesting constraints.
The ruby nesting constraints as part of [10.2.34](https://w3c.github.io/ttml2/#style-attribute-ruby) currently are a bit hard to read, as it is a quite long list of about 30 constraints without any separation in between.

As it seems that adjacent constraints have been grouped by the ruby category that applies to the element on which it is actually specified, the list should somehow be (visually) divided when that ruby category changes. Or for example something like that (of course with correct text emphasis):

- if the computed value of tts:ruby is container
  - then the computed value of tts:ruby of all ancestor elements is none
  - then the computed value of tts:ruby of each of its child elements is not none
  - ...
- ...

Furthermore the order of the constraints needs to be fixed, as after the first constraint applying to `delimiter`, two further `text` constraints follow. So the `delimiter` constraints should be moved afterwards the two.

(and there is a small typo; search for "descendats")

### LINKED ISSUE #787: Clarify no preceding sibling case on ruby text content.
The ruby nesting constraints contain the following constraint:

"If the computed value of tts:ruby is text, then the computed value of tts:ruby of its preceding sibling is base, text, or delimiter"

This however will not work when complex ruby is used, as the first child of a ruby text container is ruby text content, but has no preceding sibling. See e.g. "Example Fragment – Double-sided Ruby",

So in line with the respective constraint related to ruby base content, the mentioned constraint should be changed to:

"If the computed value of tts:ruby is base, then its preceding sibling is either null (i.e., no preceding sibling) or the computed value of tts:ruby of its preceding sibling is base, text, or delimiter."

### LINKED ISSUE #788: Clarify no preceding sibling case on ruby delimiter.
The issue of #787 also affects the "delimiter" case i.e. the following constraint (as TTML2 also allows delimiters with complex ruby, while the W3C ruby recommendation only allows delimiters when simple ruby is used):

"If the computed value of tts:ruby is delimiter, then the computed value of tts:ruby of its preceding sibling is base or text."

In addition ruby text content must not occur before a ruby delimiter, so the value "text" has to be removed from the constraint, too.

Proposal to change to:

"If the computed value of tts:ruby is delimiter, then its preceding sibling is either null (i.e., no preceding sibling) or the computed value of tts:ruby of its preceding sibling is base."

skynavga:
…#788).

Closes #786.
Closes #787.
Closes #788.


cconcolato:
Built version https://rawgit.com/w3c/ttml2/7a823a1dde0223104070248ec2ffbd2827c0ad6f/index.html#style-attribute-ruby

spoeschel:
@skynavga
Thank you, this really improved the readability.

PR #798, opened 2018-05-25
Clarify 'outside' for annotation position (#782)

### LINKED ISSUE #782: Clarify "outside" semantics in <annotation-position>
Shouldn't "outside" mean "before" if there is only one line?

The current prose states _[outside] is equivalent to before for all but the last affected line area; otherwise, equivalent to after for the last affected line area_, which translates to "after" if there is only one line (which is also the last).

skynavga:
- when only one line is present.
- when more than two lines are present

Closes #782.


skynavga:
@nigelmegitt request approval for early merger at upcoming meeting

cconcolato:
Built spec here : https://rawgit.com/w3c/ttml2/f015e2fa26107ad16c6fd67f08532de04a7a7e70/index.html

skynavga:
Ensure that `outside` is defined in a consistent manner with that used by `tts:rubyReserve` (about which see recent update in #812). In particular, `outside` is `before` for first line only and `after` for subsequent line(s).

skynavga:

PR #805, opened 2018-05-29
Add and update element derivations (#380).

### LINKED ISSUE #380: Update vocabulary derivation tables in Appendix N.
Add or update element and attribute vocabulary derivation information.

skynavga:
Closes #380.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -25198,6 +25198,12 @@ CTA-708-D,
 <titleref href="http://www.cta.org/">Digital Television (DTV) Closed Captioning</titleref>, CTA Standard,
 Consumer Technology Association (CTA).
 </bibl>
+<bibl id="data-scheme" key="Data Scheme">L. Masinter,
+<titleref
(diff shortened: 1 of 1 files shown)


nigelmegitt:
In case you hadn't noticed it @skynavga , if you are going to add `xml:base` then note 3 under the element derivations table should also have `xml:base` added.

spec/ttml2.xml
```diff
@@ -25866,6 +25878,20 @@ content vocabulary</td>
...
 <tr>
+<td><loc href="#embedded-content-vocabulary-chunk"><code>tt:chunk</code></loc></td>
+<td><emph>none</emph></td>
+<td></td>
+<td></td>
+<td></td>
+</tr>
+<tr>
+<td><loc href="#embedded-content-vocabulary-data"><code>tt:data</code></loc></td>
+<td><emph>none</emph></td>
```
nigelmegitt: Not [rfc2397](https://tools.ietf.org/html/rfc2397)? Seems to me like that's a pretty good match as a conceptual basis.

PR #807, opened 2018-05-30
Add tts:lineShear and tts:shear style attributes (#784).

### LINKED ISSUE #784: Add support for block level shear.
`tts:fontShear` currently specifies that "for the purpose of determining applicability of this style property, each character child of a `p` element is considered to be enclosed in an anonymous span." Combined with the fact that the style property is inheritable, this means that the shear transformation is applied independently to each character child of a `p` element.

Based on D-Cinema practice, a `p` should instead be sheared as a block (instead of each of its character elements independently) in order to preserve character alignment across lines esp. in the case of ruby text (see https://github.com/w3c/imsc-tests/blob/imsc-1.1/imsc1_1/png/fontShear001/3.000000.png).

Furthermore, shearing of inline elements is not supported by CSS.

I recommend making `tts:fontShear` applicable to `p` only, and specify that it yields a transformation of the entire block.

Alternatively, `#fontShear` could be split into `#fontShear-block` and `#fontShear-inline`, and the semantics of `tts:fontShear` clarified to specify that how it applies differently to inline and block areas.

skynavga:
Closes #784.


skynavga:
Need to add example code and renderings for both new properties.

skynavga:
@palemieux would it be possible for you to provide sample code and renderings for both these new properties; if so, then I would suggest two examples, both having more than one line, where one has ruby in a horizontal writing mode and the other has tate-chu-yoko in a vertical writing mode

cconcolato:
Built version here:
https://rawgit.com/w3c/ttml2/d5da9f5363dc9e6a0498d2f7ae858653e22b072d/index.html

spec/ttml2.xml
```diff
@@ -10814,6 +10848,91 @@ the computed value, then the value most distant from 0, i.e., the largest line h
...
+</tr>
+<tr>
+<td><emph>Semantic basis:</emph></td>
+<td><loc href="#derivation-lineShear">lineShear derivation</loc></td>
+</tr>
+</tbody>
+</table>
+<p>If the value of this attribute is <code>0%</code>, then no shear transformation is applied; if the value is <code>100%</code>, then a 2D shear transformation
+of 90 degrees is applied in the axis associated with the inline progression direction; if the value is <code>-100%</code>, then a shear transformation
+of -90 degrees is applied. If the absolute value of the specified percentage is greater than 100%, then it must be interpreted as if 100% were specified with
```
cconcolato: Can we reduce the repetition by having this text in a single section and have the other ones refer to it?

spec/ttml2.xml
```diff
@@ -10814,6 +10848,91 @@ the computed value, then the value most distant from 0, i.e., the largest line h
...
+</eg>
+<p>where <emph>a</emph> is equal to the tangent of the shear angle.</p>
+<p>If the inline progression direction corresponds to the Y axis, then the 2D shear transformation is described by the following matrix:</p>
+<eg xml:space="preserve">
+| 1 0 0 |
+| <emph>a</emph> 1 0 |
+| 0 0 1 |
+</eg>
+<p>For the purpose of determining applicability of this style property, a shear (skew) transformation is applied individually
+to each line area generated by a <el>p</el> element.</p>
```
cconcolato: Why is "for the purpose of determining applicability"? 
Also that sentence seems to be covered already by the above sentence:
>The tts:lineShear attribute is used to specify a style property that determines whether and how a shear transformation is applied to line areas.

If you want to have sentence that is not an overview, I would suggest replacing the entire text after the table by:
> When applying this property, a shear transformation is applied to line areas. The transformation is derived from the tts:lineShear attribute value as defined in tts:fontShear.

spec/ttml2.xml
```diff
@@ -12346,6 +12470,93 @@ with a value equal to one half of the <emph>used value</emph> of the <code>tts:l
...
+<td><emph>Animatable:</emph></td>
+<td>discrete, continuous</td>
+</tr>
+<tr>
+<td><emph>Semantic basis:</emph></td>
+<td><loc href="#derivation-shear">shear derivation</loc></td>
+</tr>
+</tbody>
+</table>
+<p>If the value of this attribute is <code>0%</code>, then no shear transformation is applied; if the value is <code>100%</code>, then a 2D shear transformation
```
cconcolato: Same remark for this repeated text

spec/ttml2.xml
```diff
@@ -12346,6 +12470,93 @@ with a value equal to one half of the <emph>used value</emph> of the <code>tts:l
...
+<td><emph>Initial:</emph></td>
+<td><code>0%</code></td>
+</tr>
+<tr>
+<td><emph>Applies to:</emph></td>
+<td>
+<loc href="#document-structure-vocabulary-body"><el>body</el></loc>,
+<loc href="#content-vocabulary-div"><el>div</el></loc>,
+<loc href="#content-vocabulary-p"><el>p</el></loc>,
+<loc href="#layout-vocabulary-region"><el>region</el></loc>
```
cconcolato: What does it mean to be applicable to body ? There is a single shear (single origin) for all p in the document? Same thing for region and div. I think we should only make it applicable to p.

spec/ttml2.xml
```diff
@@ -12346,6 +12470,93 @@ with a value equal to one half of the <emph>used value</emph> of the <code>tts:l
...
+</eg>
+<p>where <emph>a</emph> is equal to the tangent of the shear angle.</p>
+<p>If the inline progression direction corresponds to the Y axis, then the 2D shear transformation is described by the following matrix:</p>
+<eg xml:space="preserve">
+| 1 0 0 |
+| <emph>a</emph> 1 0 |
+| 0 0 1 |
+</eg>
+<p>For the purpose of determining applicability of this style property, a shear (skew) transformation is applied individually
+to each line area generated by a <el>p</el> element.</p>
```
cconcolato: The term "individually" could be confusing here. The important part is that there is one transformation (one origin) for all the lines. 

Also it talks about the p element but the table says it's applicable to region, body, div which may have multiple p elements. I am assuming that it is applicable only to p (otherwise the proposed text below would need to be changed). 

I would rephrase as follows:
> When applying this property, a single shear transformation, i.e. with the same origin, is used to transform all the line area generated by the p element. The transformation is derived from the tts:shear attribute value as defined in tts:fontShear.

spec/ttml2.xml
skynavga: Addressed in https://github.com/w3c/ttml2/pull/807/commits/27194c7843e0428fc04f4684f7f68276c9b24f52. Moved text to a new commonly referenced section 10.4.5.3.

spec/ttml2.xml
skynavga: Addressed in https://github.com/w3c/ttml2/pull/807/commits/27194c7843e0428fc04f4684f7f68276c9b24f52. Moved text to a new commonly referenced section 10.4.5.3.

spec/ttml2.xml
skynavga: Addressed in https://github.com/w3c/ttml2/pull/807/commits/27194c7843e0428fc04f4684f7f68276c9b24f52 by  removing (redundant) paragraph.

spec/ttml2.xml
skynavga: Addressed in https://github.com/w3c/ttml2/pull/807/commits/27194c7843e0428fc04f4684f7f68276c9b24f52 by  removing (redundant) paragraph.

spec/ttml2.xml
skynavga: In other cases where we have a property that applies to a block area, such as padding, we allow that property to apply to all content elements that produce a block area, namely, `p`, `div`, and `body`. I'm not sure what you mean by "there is a single shear (single origin) for all `p`". In any case, I will go ahead and scale back applicability to `p` for the time being, and ask the WG tomorrow whether anyone thinks we should also allow application to `div` or higher level generated block areas (from `body` and `region`).

skynavga: See https://github.com/w3c/ttml2/pull/807/commits/bfcc71b950e68592d926513f5d0ac89b3613c109.

skynavga:
@cconcolato please re-review

cconcolato:
lineShear or shear enable having ruby or tate-chu-yoko slanted and aligned properly as expected by native japanese readers (at least from our experience), however, proper tate-chu-yoko slanting requires aligning slanted glyphs on a slanted baseline. This might be hard to implement in some cases. If such implementation cannot be done correctly, I'm told that it is preferable to not apply shear at all rather than applying a wrong shear transform (e.g. using a horizontal shear when tts:lineShear is used). I would therefore suggest adding a sentence such as the following:
> If a presentation processor does not support presenting characters affected by tts:textCombine, and any of the tts:shear or tts:lineShear properties that apply, then it must perform rendering as if the tts:shear and tts:lineShear had been set to 0% on those characters.

skynavga:
@cconcolato if line shear is applied, then it applies to the entire line area as a whole, regardless of its contents; if an unsheared vertical line area sets tate-chu-yoko on a horizontal baseline, then the vertically sheared line area will result in a slanted baseline for the tate-chu-yoko as desired; the presentation engine need not do anything internally to adjust the baseline; furthermore, the same applies if block shear is used; given these facts, why is any additional text needed?

nigelmegitt:
Built version actually here: https://rawgit.com/w3c/ttml2/issue-0784-shear-build/index.html

spec/ttml2.xml
```diff
@@ -1331,6 +1354,13 @@ that is outside the scope of this specification.</p>
 <p>An axis parallel to the plane of the horizon.</p>
 </def>
 </gitem>
+<gitem id="terms-line-area">
+<label>[line area]</label>
+<def>
+<p>A type of <loc href="#terms-block-area">block area</loc>, as defined by <bibref ref="xsl11"/>, &sect;4.5, which children
+consist of one or more <loc href="#terms-inline-area">inline areas</loc>.</p>
```
nigelmegitt: Doesn't this need to say that all those inline areas are positioned within the same line?

skynavga: No. It is self-evident. A line area is a single line, so how could a child inline area be in another line?

skynavga:
@palemieux [here](https://www.w3.org/International/datespace/2010/02/jlreq-examples/) are some somewhat longer examples of ruby that can be broken into a couple of short lines; perhaps https://github.com/akiseino can be some assistance here; I will look for longer tate-chu-yoko examples also

skynavga:
@palemieux [here](https://helpx.adobe.com/in/indesign/using/formatting-cjk-characters.html) are some interesting examples depicting both a form of font (per-glyph) shear (_shatai_) and also tate-chu-yoko

skynavga:
@palemieux more [here](https://lists.w3.org/Archives/Public/www-archive/2013Sep/0012.html)

skynavga:
@palemieux here's an excerpt from the example in https://github.com/w3c/ttml2/pull/807#issuecomment-394117937 that I coded and ran through TTPE to produce an SVG output which I manually tweaked to add skew transforms depicting line and block shear; the borders around region areas and line areas (and line labels) are an optional decoration TTPE can produce, which, in this case, shows how the tate-chu-yoko is causing the line height (bpd) to be increased on the second (left) line containing "1988";

screen shot 2018-06-02 at 4 34 47 pm

TTML2 source and tweaked SVG output is here:
[sheared-text-combine.zip](https://github.com/w3c/ttml2/files/2065449/sheared-text-combine.zip)

skynavga:
@palemieux @nigelmegitt BTW, here is an example of font shear from TTPE for the similar content (with slightly wider regions and smaller font) which depicts how the line heights (bpds) of the lines containing TCY (tate-chu-yoko) have been slightly increased when shear is applied to the TCY glyphs; you can see this by observing the width of the remaining (available) space on either side of the two lines for the middle region vs the left/right regions

screen shot 2018-06-02 at 5 02 20 pm

akiseino:
@palemieux Yes, I can make slant Tate-Chu-Yoko sample which has two lines with one TCY line tomorrow.
Is that OK by Lambda Cap or Digital cinema ?

skynavga:
@akiseino it would be nice to see the same example rendered by both a LambdaCAP system, e.g., CanvaSS, and also a D-Cinema system; if you can, do one set with shear and another set without shear; a multi-line ruby example using the same configurations would be welcome as well

akiseino:
@skynavga 

> @akiseino it would be nice to see the same example rendered by both a LambdaCAP system, e.g., CanvaSS, and also a D-Cinema system; if you can, do one set with shear and another set without shear; a multi-line ruby example using the same configurations would be welcome as well

Yes, here is sheared Tate-Chu-Yoko (two lines with ruby) sample of Lambda cap. 
I'll share D-Cinema one later after clearing current work...

I made below four set.
A) First one (tate_sample1.JPG); no shearing both two lines.
B) Second one (tate_sample2.JPG); shearing both two lines.
C) Third one (tate_sample3.JPG); shearing left side line only.
D) Fourth one (tate_sample4.JPG); shearing right side line only.
The test_TateChuYoko.cap writes above order. 
I attached the image file, test_TateChuYoko.cap_image.jpg just in case.

It's just made two lines but I may move 2nd line (including TCY 1998)
to left side on image in normal work for natural look.

Please let me know any additional requests. Thank you.

tate_sample1
tate_sample2
tate_sample3
tate_sample4
test_tatechuyoko cap_image
[test_TateChuYoko.zip](https://github.com/w3c/ttml2/files/2066989/test_TateChuYoko.zip)

skynavga:
@akiseino excellent; exactly what I was looking for; i'm looking forward to seeing D-Cinema version as well

cconcolato:
> @cconcolato if line shear is applied, then it applies to the entire line area as a whole, regardless of its contents; if an unsheared vertical line area sets tate-chu-yoko on a horizontal baseline, then the vertically sheared line area will result in a slanted baseline for the tate-chu-yoko as desired; the presentation engine need not do anything internally to adjust the baseline; 

It depends how you implement it. You could still implement it on a glyph-by-glyph basis, in which case you have to track the baseline.

> furthermore, the same applies if block shear is used; given these facts, why is any additional text needed?

The one I suggested:
> If a presentation processor does not support presenting characters affected by tts:textCombine, and any of the tts:shear or tts:lineShear properties that apply, then it must perform rendering as if the tts:shear and tts:lineShear had been set to 0% on those characters.

skynavga:
@cconcolato neither of your comments make any sense to me; nobody would implement line shear by using font (glyph) shear; doing font shear is much more complex than doing line shear; if you are going to choose to implement one or the other, you are going to choose line shear;

as for your second comment, why would you specify a behavior for shear as a fallback position for lack of support for text combine? that doesn't make any sense either; we have **never** defined such a _combination property fallback_ and I don't intend to do so now; if you don't support text combine, that has nothing to do with your support for shear and should not have any impact on shear in any case

akiseino:
@skynavga 
> @akiseino excellent; exactly what I was looking for; i'm looking forward to seeing D-Cinema version as well

Thank you for your waiting.
I made below four set for D-Cinema same as Rambda Cap.
A) First one (DCP_sample1.JPG); no shearing both two lines.
B) Second one (DCP_sample2.JPG); shearing both two lines.
C) Third one (DCP_sample3.JPG); shearing left side line only.
D) Fourth one (DCP_sample4.JPG); shearing right side line only.
The sample20180605.xml writes above order.
I attached the image file, sample20180605.xml.jpg just in case.

In the D-Cinema, I will not move 2nd line (including TCY 1998) normally.
It means both two lines locate right side only.
It's different from Lambda Cap because D-cinema is lager screen
and audience is hard to see both side subtitle.
And it's Interop DCP not SMPTE DCP. In Japan, Interop DCP is major.

Please let me know any additional requests. Thank you.
dcp_sample1
dcp_sample2
dcp_sample3
dcp_sample4
sample20180605 xml
[sample20180605.zip](https://github.com/w3c/ttml2/files/2071466/sample20180605.zip)

skynavga:
thanks very much @akiseino, you've been a great help

spec/ttml2.xml
nigelmegitt: It doesn't actually say that the line area is a single line though.

skynavga: if you read the referenced XSL-FO material, it is quite obvious that is the case; we don't need to say so here.

palemieux:
2-line 
`tts:shear` example:

[shear001-example.zip](https://github.com/w3c/ttml2/files/2073995/shear001-example.zip)

palemieux:
`tts:textCombine` + `tts:shear example`:

[shear002-example.zip](https://github.com/w3c/ttml2/files/2074069/shear002-example.zip)

cconcolato:
@palemieux is there a way to increase the font weight, specially for the ruby because the images look quite pixelated.

spec/ttml2.xml
nigelmegitt: Oh, I see, the order of the text led me to a different interpretation of the reference. It looks like the ", as defined by [XSL ..." applies to _a type of block area_ not to _line area_. Suggest rewording to:

```
<p>As defined by <bibref ref="xsl11"/>, &sect;4.5; a type of <loc href="#terms-block-area">block area</loc> which children
consist of one or more <loc href="#terms-inline-area">inline areas</loc>.</p>
```

Note the use of a semicolon after the definition reference.

skynavga: @nigelmegitt every definition in the terminology section has an implied subject, namely, the term being defined, and it is clear from the language that it is the subject **[line area]** that is being referred to; the same applies to all of the other definitions of **[XXX area]** and every other term for that matter; you can't possibly expect that we rewrite all the terms because they assumed another subject applies;

you will note that the terminology section uses a formula for presentation of the form:

[SUBJECT] [[_is defined to be_]] [DEFINITION]

if you read it this way, it will help you avoid misreadings

nigelmegitt: I've reviewed all the term definitions and this problem arises for the newly introduced terms block area, glyph area, inline area and line area only, and is not duplicated for any of the other term definitions. They all have the form:

 "A type of [x], as defined by [ref]" 

when the thing being defined in the reference is the term not the [x]. 

They are all misleading in this respect, and yes, I'm expecting that we rewrite those in the form:

"As defined by [ref], a type of [x]".

`<snarky>`If you write it that way, it will help you avoid having to explain how to read English the way you write it. `</snarky>`

spec/ttml2.xml
```diff
@@ -10844,6 +10864,106 @@ the computed value, then the value most distant from 0, i.e., the largest line h
...
+to aid in debugging layout issues. Such lines are not part of a normative TTML presentation as defined by this specification.
+Here, the decorations depict (1) bounderies of outer <loc href="#terms-block-area">block areas</loc>
+generated by <loc href="#layout-vocabulary-region"><el>region</el></loc> elements,
+(2) boundaries of <loc href="#terms-line-area">line area</loc> children of the outer <loc href="#terms-block-area">block areas</loc>
+generated by <loc href="#content-vocabulary-p"><el>p</el></loc> elements, and
+(3) boundaries of <loc href="#terms-inline-area">inline area</loc> children of <loc href="#terms-line-area">line areas</loc> generated
+by implied ruby text containers.</p>
+<p>This example depicts how a shear (skew) transform is applied to each affected line area as a whole; however,
+when line shear is to be applied to a line, then the available measure that can be allocated to the line, i.e., its maximum inline progression dimension
+length, is reduced by <code>2*B*sin(&theta;)</code>, where <code>B</code> is the block progression dimension of the line area and
```
nigelmegitt: suggest changing "i.e. its maximum inline length" to "i.e. its maximum inline length prior to shear transformation" for clarity.

spec/ttml2.xml
```diff
@@ -12395,6 +12520,100 @@ with a value equal to one half of the <emph>used value</emph> of the <code>tts:l
...
+<p>In the above example, the yellow lines are decorations created by a particular presentation processor
+to aid in debugging layout issues. Such lines are not part of a normative TTML presentation as defined by this specification.
+Here, the decorations depict (1) bounderies of outer <loc href="#terms-block-area">block areas</loc>
+generated by <loc href="#layout-vocabulary-region"><el>region</el></loc> elements and
+(2) boundaries of <loc href="#terms-line-area">line area</loc> children of the outer <loc href="#terms-block-area">block areas</loc>
+generated by <loc href="#content-vocabulary-p"><el>p</el></loc> elements.</p>
+<p>This example depicts how a shear (skew) transform is applied to the affected outer <loc href="#terms-block-area">block areas</loc>
+generated by <loc href="#content-vocabulary-p"><el>p</el></loc> elements; however,
+when shear is to be applied to an outer <loc href="#terms-block-area">block area</loc>, then the available measure that can be allocated to the block area,
+i.e., its maximum inline progression dimension length, is reduced by <code>2*B*sin(&theta;)</code>, where <code>B</code> is the block progression dimension
```
nigelmegitt: Suggest changing "i.e., its maximum inline progression dimension length" to "i.e. its maximum inline progression dimension length prior to shear transformation" for clarity (and to remove the unnecessary comma).

skynavga:
@nigelmegitt it is impossible to tell from the examples in tts:fontShear which was applied, because they will look identical unless ruby, text emphasis, or text combine are present, which is not the case for the examples

spec/ttml2.xml
skynavga: Done.

spec/ttml2.xml
skynavga: Done (but the commas are correct, so no change there).

nigelmegitt:
> it is impossible to tell from the examples in tts:fontShear which was applied

@skynavga I was expecting the inline progression extent of each glyph area to be increased as a result of the transformation, so that they are effectively spaced apart further. As it is, the sheared glyph areas overlap in the inline direction.

I just tried to create an example of this in CSS and discovered that according to [CSS Transforms Module Level 1](https://drafts.csswg.org/css-transforms-1/#transformable-element) non-replaced inline elements are not transformable, and indeed applying a transform to a `<span>` has no effect.

Given this and the previous discussion, I wonder if we should remove `tts:fontShear` altogether.

skynavga:
@nigelmegitt is there any reason you haven't approved this PR yet? removing fontShear would have to be another issue, and as I answered @cconcolato who asked this question, LambdaCAP requires being able to specify shear for individual characters, meaning they may be different;

when all of the glyph areas on a line are assigned the same shear, there is no extra space added; however, at boundaries between different shears, it is added

spec/ttml2.xml
```diff
@@ -10844,6 +10864,107 @@ the computed value, then the value most distant from 0, i.e., the largest line h
...
+<tbody>
+<tr>
+<td><graphic source="images/lineShear_1.png" alt="TTML lineShear style property"/></td>
+</tr>
+</tbody>
+</table>
+<note role="explanation">
+<p>In the above example, the yellow lines are decorations created by a particular presentation processor
+to aid in debugging layout issues. Such lines are not part of a normative TTML presentation as defined by this specification.
+Here, the decorations depict (1) bounderies of outer <loc href="#terms-block-area">block areas</loc>
```
cconcolato: typo: bounderies should be boundaries

spec/ttml2.xml
```diff
@@ -12395,6 +12521,100 @@ with a value equal to one half of the <emph>used value</emph> of the <code>tts:l
...
+<tbody>
+<tr>
+<td><graphic source="images/shear_1.png" alt="TTML shear style property"/></td>
+</tr>
+</tbody>
+</table>
+<note role="explanation">
+<p>In the above example, the yellow lines are decorations created by a particular presentation processor
+to aid in debugging layout issues. Such lines are not part of a normative TTML presentation as defined by this specification.
+Here, the decorations depict (1) bounderies of outer <loc href="#terms-block-area">block areas</loc>
```
cconcolato: same typo

skynavga:

PR #808, opened 2018-05-30
Add xml:base to core vocabulary (#804).

### LINKED ISSUE #804: Add missing xml:base attribute.
The ``xml:base`` attribute is missing from the following elements that employ a possibly relative ``<uri>`` valued attribute, e.g., ``src`` and ``xlink:href``:
- ``audio``
- ``data``
- ``font``
- ``image``
- ``span``
- ``source``

In addition, at least the ``tt`` element should be able to specify a document-wide default base as well.

skynavga:
Closes #804.


cconcolato:
built version here:  https://rawgit.com/w3c/ttml2/2e148d4cb8a96c353d1b310f8ea58e1a58f15adb/index.html

spec/ttml2.xml
```diff
@@ -6011,6 +6020,20 @@ of partially rather than wholly ignoring the semantics of the element, about whi
...
+<p>The <att>xml:base</att> attribute is used as defined by
+<bibref ref="xmlbase"/>.</p>
+<p>The <att>xml:base</att> attribute may be used with any element in the core vocabulary catalog
+that admits a <loc href="#content-attribute-condition"><att>condition</att></loc> attribute, a descendant
+element in the <loc href="#element-vocab-group-metadata"><code>Metadata.class</code></loc> element group, or
+a descendant element that admits an <att>xml:base</att> attribute.</p>
+<note role="elaboration">
+<p>In this version of this specification, the <att>xml:base</att> attribute is permitted on any element
+type defined by the <specref ref="core-vocabulary-overview"/> except for the
+<loc href="#embedded-content-vocabulary-chunk"><el>chunk</el></loc> element.</p>
```
cconcolato: This seems contradictory with the definition of the chunk element. Suggest to delete the note.

spec/ttml2.xml
```diff
@@ -6011,6 +6020,20 @@ of partially rather than wholly ignoring the semantics of the element, about whi
...
 <loc href="#embedded-content-vocabulary-image"><el>image</el></loc> element.</p>
 </div3>
+<div3 id="content-attribute-xml-base">
+<head>xml:base</head>
+<p>The <att>xml:base</att> attribute is used as defined by
+<bibref ref="xmlbase"/>.</p>
+<p>The <att>xml:base</att> attribute may be used with any element in the core vocabulary catalog
+that admits a <loc href="#content-attribute-condition"><att>condition</att></loc> attribute, a descendant
+element in the <loc href="#element-vocab-group-metadata"><code>Metadata.class</code></loc> element group, or
+a descendant element that admits an <att>xml:base</att> attribute.</p>
```
cconcolato: A quick check shows that xml:base and xml:id can be used in the same places except for isd:css and isd:region. Is that correct? Why is this sentence different from the one in xml:id that simply says: 
>may be used with any element in the core vocabulary catalog.

spec/ttml2.xml
skynavga: Removed comment.

spec/ttml2.xml
skynavga: Added `xml:base` to ISD vocabulary. In addition, added missing `xml:id` and `xml:lang` that were missing on a few ISD vocabulary. Synchronize schemas.

skynavga:
@cconcolato please re-check

nigelmegitt:
Built version really at https://rawgit.com/w3c/ttml2/issue-0804-xml-base-build/index.html

nigelmegitt:
The element derivations need updating to add `xml:base` too.

spec/ttml2.xml
```diff
@@ -6011,6 +6020,15 @@ of partially rather than wholly ignoring the semantics of the element, about whi
...
```
nigelmegitt: This list excludes the `tt` and `head` elements, but they do actually allow `xml:base`. It seems like an unhelpful dependency to use `condition` as a predicate for defining what `xml:base` is permitted on, since there's no obvious relationship between the two concepts. Or is the idea that when resolving condition expressions the applicable value of `xml:base` is used as a "namespace" prefix used to qualify named parameters or media queries? Otherwise perhaps define it relative to any element that can be an ancestor of a metadata class element?

skynavga:
@cconcolato ping

spec/ttml2.xml
nigelmegitt: By my reading neither `tt` nor `head` are a descendant element of one of the other categories so the last clause does not include them.

skynavga: I will add a (1), (2), and (3) inline after the first _admits_.

skynavga: @nigelmegitt done... currently building

nigelmegitt: Thanks, that addresses the comment.

skynavga:

PR #810, opened 2018-06-01
Clarify luminanceGain applicability.

### LINKED ISSUE #796: Clarify that tts:luminanceGain shall not apply unless absolute HDR scheme used.
It is not entirely clear that `tts:luminanceGain` must be ignored if the HDR scheme being used does not require absolute luminance values. In practice it would be actively dangerous to apply `tts:luminanceGain` for scene relative HDR schemes, but there is a use case where content providers want to specify `tts:luminanceGain` but have it only apply when, for example, PQ is used, and not apply for HLG, say.

I propose that we add a clarifying normative statement to back up the introductory clause: _, and when that presentation requires an absolute luminance value._ to make this a shall/shall not.

By the way, the test for if it should be applied is based on the qualities of the related media object, not on the content. So for example I expect `tts:luminanceGain` to be applied even to the images in a region and even if those images are already, say, PQ PNG images. We should probably advise not putting PQ images in `region`s that set `tts:luminanceGain` to anything other than 1.

nigelmegitt:
Closes #796 by clarifying that `tts:luminanceGain` only applies when the document processing context requires absolute luminance pixel value outputs and only to sRGB pixel values.


skynavga:

PR #812, opened 2018-06-01
Remove around and between in rubyReserve (#811).

### LINKED ISSUE #811: Eliminate references to around and between in tts:rubyReserve.
https://github.com/w3c/ttml2/issues/783 was addressed by https://github.com/w3c/ttml2/pull/800 and the `around` and `between` values where removed for the syntax, but the semantics for `around` and `between` remained, and `around` is referenced by `auto`.

Also, the semantics are hard to read because of all the if conditions.

cconcolato:
Closes #811


skynavga:
@palemieux addressed  https://github.com/w3c/ttml2/pull/812#pullrequestreview-125413191 in https://github.com/w3c/ttml2/pull/812/commits/6ef7ada152f1f69498904cdf77637920bbe4df9d; I had also been pondering the removal of `auto`, which did have a subtle difference with `outside` since it was originally defined in terms of `around` in the case of more than one line, and `around` applied `before` to all but the last line while `outside` applies `after` to all but the first line; my conclusion is that we should remove `auto` since it will be confusing to have both `auto` and `outside` due to this subtle difference, and I'm doubtful there is a strong use case for the former `around` semantics anyway; note that this change upgrades this PR to substantive

skynavga:

PR #813, opened 2018-06-04
Revert deprecation of 'seq' time container in smpte discontinuous mod…

### LINKED ISSUE #377: Time containment semantics for smpte time base.
Indicate that parallel time container semantics always apply in smpte discontinuous mode, and, as such, sequential containment must not be specified, i.e., parallel containment always applies.

Editorial notes from 2015-01-14.

skynavga:
…e (#377).

Closes #377.


skynavga:

PR #815, opened 2018-06-05
Further refinement of features (#814).

### LINKED ISSUE #814: Further fine tuning of features.
Now that #794 has been merged into master, there are some additional tweaks needed, about which see the PR associated with this issue.

### LINKED ISSUE #819: Fix typos.
- "the the" in https://w3c.github.io/ttml2/#feature-extent-version-2
- use of `#opacity` instead of `#origin` in https://w3c.github.io/ttml2/#feature-origin

### LINKED ISSUE #821: Missing hook from conformance to profile abort processing.
When the new profile related spec language was incorporated, a hook from processing conformance to profile processing semantics was unintentionally omitted. That hook needs to be added at this time.

### LINKED ISSUE #822: Inconsistent application of profile combination methods.
The procedures for constructing effective content and processor profiles applies an inconsistent rule to the combination of profiles when those profiles derive from a list of designators versus a list of top-level profiles. In both cases, they should be using the same methods, namely, `ttp:contentProfileCombination` and `ttp:processorProfileCombination`.

skynavga:
Closes #814.
Closes #819.
Closes #821.
Closes #822.

Addresses lingering comment from #718, #739, i.e., adding ``#length-version-2``.
Addresses https://github.com/w3c/ttml2/issues/697#issuecomment-390088403 from #697.


spec/ttml2.xml
```diff
@@ -20032,10 +20039,15 @@ it does not imply support for the <loc href="#style-attribute-extent-special-usa
 <p>A TTML <loc href="#terms-transformation-processor">transformation processor</loc> supports the
 <code>#extent-image</code> feature if it supports the attribute values
 of the <loc href="#style-attribute-extent"><att>tts:extent</att></loc> attribute supported by the
-<loc href="#feature-extent-auto"><code>#extent-auto</code></loc> and <loc href="#feature-extent-length"><code>#extent-length</code></loc> features
-but applied to an <loc href="#embedded-content-vocabulary-image"><el>image</el></loc> element.</p>
+<loc href="#feature-extent-auto-version-2"><code>#extent-auto-version-2</code></loc> and
```
nigelmegitt: `#extent-auto-version-2` doesn't state that it only applies to `region`, so it seems that it already implies application to `image`.

spec/ttml2.xml
```diff
@@ -19987,6 +19987,8 @@ attribute as defined by this specification.</p>
 feature if it implements presentation semantic support for the semantic extensions to the <code>auto</code> value of the
 <loc href="#style-attribute-extent"><att>tts:extent</att></loc> attribute as described above.</p>
 <p>The <code>#extent-auto-version-2</code> feature is a semantic extension of the <code>#extent-auto</code> feature.</p>
```
nigelmegitt: Does `#extent-auto-version-2` imply support for `#region-implied-animation`? Do we not need to scope it to some particular elements?

skynavga: No and No. It applies to whatever elements that `auto` is assigned semantics to in the context of those elements. Namely, to `tt`, `region`, and `image`.

spec/ttml2.xml
skynavga: The intent here is to define `#extent-image` as a subset of the full set of values that may apply to a `region`; in particular, the subset that excludes `contain`, `cover`, and `<measure>`. What is remains is the union of the set of values supported by `#extent-auto-version-2` that apply to `image` and the set  of values supported by `#extent-length` that apply to `image`.

spec/ttml2.xml
nigelmegitt: I don't think it is invalid for an author to create:

```
<p tts:extent="auto" begin="..." end="..." xml:id="p123">
[some contents]
</p>
...

so it also applies to `div` and `p`, doesn't it?

File: spec/ttml2.xml (hunk 1 above)
Comment (nigelmegitt) [reply spec/ttml2.xml]: I'm not quite following you. Here's my reasoning:

`#extent-auto-version-2` and `#extent-length` imply support for some syntax and semantics that apply both to `region` and to `image`.

`#extent-image` implies support for `#extent-auto-version-2` and `#extent-length` as applied to `image`. Since they already do apply to `image`, that's a no-op. So by this argument `#extent-image` is unnecessary.

I don't think that's the right answer though because it seems like a good idea to add features for the application to `image`. What about creating `#extent-region-auto-version-2` and `#extent-image-auto-version-2` being similar but have the former relate to application on `region` and the latter on `image`?

File: spec/ttml2.xml (hunk 2 above)
Comment (nigelmegitt) [reply spec/ttml2.xml]: OK, I see what you mean @skynavga , thanks for explaining. Would you object to a note under `#extent-auto-version-2` (or, depending on your response to https://github.com/w3c/ttml2/pull/815#discussion_r193401615, `#extent-region-auto-version-2`) saying that support for `#region-implied-animation` is not implied?

**Comment by skynavga** [comment]
@nigelmegitt I have attempted to address your (private) comments

**Comment by skynavga** [comment]
In order to address https://github.com/w3c/ttml2/issues/697#issuecomment-390088403, refactored profile features as follows:
1. separate profile combination semantics by introducing
    * ``#contentProfiles-combine`` and
    * ``#processorProfiles-combine``;
1. incorporate processor profile inference feature semantics into ``#contentProfiles``, allowing removal of ``#inferProcessorProfile``;
1. separate profile nesting semantics by introducing ``#profile-nesting``;
1. reduce ``#profile-version-2`` complexity by excluding support for profile combination, profile nesting, and use of feature narrowing and widening;
1. define full blown support for all TTML2 profile features by introducing ``#profile-full-version-2``;
1. fixed various missing or problematic language in profiles related sections;

These changes significantly reduce the complexity of the ``#profile-version-2`` feature which was made mandatory in TTML2 standard profiles. In particular, it now supports only simple,  non-combining uses of ``ttp:contentProfiles`` and ``ttp:processorProfiles``, i.e., with only one designator, use of at most one top-level profile, no nested profiles, and no feature narrowing or widening are now covered, which I expect to be the sweet spot in terms of the majority of documents.

These changes will also reduce the possible need to more seriously deal with the original issue reported in #697.

**Comment by skynavga** [comment]
@nigelmegitt ping

**Comment by cconcolato** [comment]
Built version here https://rawgit.com/w3c/ttml2/issue-0814-feature-tuning-build/index.html

File: spec/ttml2.xml [hunk 3]
```diff
@@ -785,6 +785,13 @@ of an <loc href="#terms-embedded-data-resource">embedded data resource</loc> by
 child <loc href="#embedded-content-vocabulary-chunk"><el>chunk</el></loc> element.</p>
 </def>
 </gitem>
+<gitem id="terms-combined-profile">
+<label>[combined profile]</label>
+<def>
+<p>A <loc href="#terms-profile">profile</loc> for which the constituents component of the combined specification set component of the
```
cconcolato: I don't understand this sentence. Can you rephrase?

skynavga: @cconcolato in order to understand this, you need to understand the profile object state concepts (5.2.4.1); click on the profile instance link in the definition to go there

skynavga: @cconcolato I believe the definition is correct and complete; if after you review the profile instance state object you still have questions, let me know

spec/ttml2.xml
```diff
@@ -1556,6 +1583,15 @@ a content processor.</p>
 internal refers to being internal to a <loc href="#terms-document-instance">document instance</loc>.</p>
 </def>
 </gitem>
+<gitem id="terms-profile-processing">
+<label>[profile processing]</label>
+<def>
+<p>The process of performing the <loc href="#semantics-procedure-abort-if-unsupported-processor-profile"><phrase role="strong">[abort if
+unsupported processor profile]</phrase></loc> procedure, which, as a side effect, determines the
+<loc href="#terms-effective-content-profile">effective content profile</loc> and
```
nigelmegitt: It looks as though the [abort if unsupported processor profile] process only determines the effective content profile if step 5 of [construct effective processor profile] is reached; in all other cases the effective content profile remains undetermined.

spec/ttml2.xml
```diff
@@ -4446,7 +4536,7 @@ ttp:contentProfileCombination
 </tr>
 </tbody>
 </table>
-<p>If not specified, the value of this parameter must be considered to be <code>leastRestrictive</code>.</p>
+<p>If not specified, the value of this parameter must be considered to be <code>replace</code>.</p>
```
nigelmegitt: Looks like the motivation for this change is to set a default that works for the case that processors do not support profile combination. However in that case I would expect most processors to use the unstated `ignore` value, where only _arg1_ is used and all subsequent arguments are ignored.

spec/ttml2.xml
nigelmegitt: I agree it makes sense, though it is certainly a difficult sentence to read.

spec/ttml2.xml
skynavga: The content profile is used in two places, one is by performing [validate document] during the construction of effective validation schemas, the other is by performing profile processing, which will perform [construct inferred processor profile]. Consequently, it (ECP) gets set exactly when it is needed; otherwise it is null (because it doesn't matter).

spec/ttml2.xml
skynavga: Exactly.

skynavga:
@nigelmegitt as i've pointed out above, I don't believe there is an error with respect to ECP

spec/ttml2.xml
nigelmegitt: Right, but why not define `ignore` then?

spec/ttml2.xml
nigelmegitt: OK, the text says that the process determines the ECP but we're both agreed that sometimes it doesn't. So the text needs to be changed, for example by qualifying "determines the effective content profile" with "if needed" or something like that.

spec/ttml2.xml
skynavga: Implemented. Next time make your hint sound more like a request.

skynavga:
@nigelmegitt are we done?

spec/ttml2.xml
skynavga: Actually it does always determine a value, which just may be null (if it isn't required). In any case, I added "(if required)" to satisfy your sensibilities.

skynavga:
@nigelmegitt thanks

skynavga:
Merging early per WG Resolution.

PR #818, opened 2018-06-07
Remove digits keyword from tts:textCombine (#817).

### LINKED ISSUE #817: Remove digits value from tts:textCombine values.
Per prior discussion, see https://github.com/w3c/ttml2/issues/797#issuecomment-392143715.

skynavga:
Closes #817.


skynavga:

PR #824, opened 2018-06-10
Remove word 'measure' that does not refer to <measure> value syntax (…

### LINKED ISSUE #823: Avoid use of 'measure' that does not refer to <measure> value syntax.
Eliminate use of the word "measure" that does not refer to the `<measure>` value syntax.

skynavga:
…#823).

Closes #823.


spec/ttml2.xml
```diff
@@ -15058,7 +15058,7 @@ interpretation of the specified measure as specified below.</p>
 </tr>
 </tbody>
 </table>
-<p>The semantics of measure values are defined as follows:</p>
+<p>The semantics of these values are defined as follows:</p>
```
nigelmegitt: This one should not have been changed I think - it actually does apply to `measure`!

skynavga: Sure, but it is not wrong.

PR #825, opened 2018-06-11
Add informative text clarifying line height and line stacking semanti…

### LINKED ISSUE #806: Use of maximum descendant font size in normal line height computation.
It appears that when we reworked the algorithm for resolving the value of ``normal`` line height that the language regarding use of maximum descendant font size was inadvertently left out.

TTML1 (8.2.12) says

>If the value of this attribute is ``normal``, then the computed value of the style property must be considered to be no less than the size of the largest font size that applies to the element and its descendant elements in the intermediate synchronic document as determined by 9.3.3 Intermediate Synchronic Document Construction.

As it turns out, this language, which was intended to be compatible with XSL-FO and CSS2, does not in fact yield the correct result because it takes the maximum descendant font size for the paragraph and assigns that value as a single value to the every line area generated by the paragraph. In contrast, the definition of **per-inline-height-rectangle**, which XSL-FO states is compatible with CSS2, makes use of the maximum font size of the inline areas on a per-line-area basis, not for all line areas in the paragraph, and, further, factors in the ascent, descent, and leading for individual inline area descendants of a line area.

XSL-FO 1.1 (4.5) says

>The extent of the per-inline-height-rectangle in the block-progression-direction is then defined to be the minimum required to enclose both the expanded-nominal-requested-line-rectangle and the expanded-rectangles of all the inline-areas stacked within the line-area; this may vary depending on the descendants of the line-area.

The defining language in CSS2 can be found in section 10.8, which has evolved (mainly via clarification and additional prose) from CSS2 (1998) to the current draft of [CSS2.2](https://www.w3.org/TR/CSS22/visudet.html#line-height).

Taken together, I conclude that, as presently written:
- TTML1 semantics for `normal` do not match either CSS2 or XSL-FO;
- TTML2 semantics for `normal` do not say anything (at present) about taking the maximum of descendants, and, therefore, match none of TTML1, CSS, nor XSL-FO;

To resolve this, we need to clarify/refine the algorithmic language found in [TTML2 10.2.27](file:///Users/glenn/work/ttml2/spec/ttml2.html#style-attribute-lineHeight). We also need to back port the essential clarifications to TTML1.

I am going to mark this as a a bug, and requiring substantive changes, even though one might (marginally) argue that it is editorial since our intentions were to match CSS2/XSL-FO in the first place. However, to resolve this, we will need to introduce new normative algorithmic language which can affect conformance.

skynavga:
…cs (#806).

Closes #806.


spec/ttml2.xml
```diff
@@ -24933,10 +25007,10 @@ W3C Recommendation, 28 October 2004. (See
 W3C Recommendation, 28 October 2004. (See
 <xspecref href="http://www.w3.org/TR/xmlschema-2/">http://www.w3.org/TR/xmlschema-2/</xspecref>.)
 </bibl>
-<bibl id="xsl11" key="XSL 1.1">Anders Berglund,
-<titleref href="http://www.w3.org/TR/2006/REC-xsl11-20061205/">Extensible Stylesheet Language (XSL) Version 1.1</titleref>,
+<bibl id="xslfo11" key="XSL-FO 1.1">Anders Berglund,
+<titleref href="http://www.w3.org/TR/2006/REC-xslfo11-20061205/">Extensible Stylesheet Language (XSL) Version 1.1</titleref>,
 W3C Recommendation, 05 December 2006. (See
-<xspecref href="http://www.w3.org/TR/2006/REC-xsl11-20061205/">http://www.w3.org/TR/2006/REC-xsl11-20061205/</xspecref>.)
+<xspecref href="http://www.w3.org/TR/2006/REC-xslfo11-20061205/">http://www.w3.org/TR/2006/REC-xslfo11-20061205/</xspecref>.)
```
nigelmegitt: This is a 404.

spec/ttml2.xml
```diff
@@ -17550,14 +17564,74 @@ the recommendations defined by <loc
...
+<p><emph>This sub-section is non-normative.</emph></p>
+<p>The calculation of the <loc href="#terms-block-progression-dimension">block progression dimension</loc> of a
+<loc href="#terms-line-area">line area</loc>, conventionally called <emph>line height</emph>,
+may be particularly complex, especially when multiple font families, font sizes, font styles,
+<loc href="#terms-font-resource">font resources</loc>, and advanced text formatting features, e.g., ruby, text emphasis, and
+text combination (tate-chu-yoko), are applied.</p>
+<p>Each <loc href="#terms-line-area">line area</loc> produced by this process is <emph>stacked</emph>, either
+vertically below the preceding <loc href="#terms-line-area">line area</loc> (in the context of horizontal lines) or
+horizontally to the side of the preceding line (in the context of vertical lines).
+As specified by step (5) of the <loc href=""><phrase role="strong">[flow transformation]</phrase></loc> procedure,
```
nigelmegitt: missing href

nigelmegitt:
I'm also unclear that the reference changes can be considered editorial, and cannot verify that for the time being since the new XSL-FO reference gives a 404.

spec/ttml2.xml

spec/ttml2.xml

skynavga:
@nigelmegitt addressed comments

skynavga:
@nigelmegitt the link changes were inadvertent and have been reverted so your last comment above no longer applies:

>I'm also unclear that the reference changes can be considered editorial, and cannot verify that for the time being since the new XSL-FO reference gives a 404.

spec/ttml2.xml
```diff
@@ -17550,14 +17564,74 @@ the recommendations defined by <loc
...
+<item><p>when the computed value of <loc href="#style-attribute-lineHeight">tts:lineHeight</loc> of a paragraph is
+a specified <loc href="#style-value-length">&lt;length&gt;</loc>, then the
+<loc href="#terms-line-area">line areas</loc> of a paragraph exactly abut one another but their content may (1) intersect with the preceding
+<loc href="#terms-line-area">line area</loc> of that paragraph (when negative <emph>half leading</emph> applies) or (2) consume more space
+in the <loc href="#terms-block-progression-dimension">block progression dimension</loc> than required (when positive <emph>half leading</emph>
+applies).</p></item>
+</ulist>
+<note role="clarification">
+<p>When any of the following advanced text formatting features is applied to the content of a line, then additional adjustments to <emph>line height</emph>
+may occur:</p>
```
cconcolato: Why is "may" used here?

skynavga: Ruby may fit into the available line height or it may require the height to be increased, text combine may fit into the available space with adjustment, or it may require more space, in which adjustment is necessary, etc.

cconcolato: Is there a normative requirement that if the ruby text does not fit, then the line height must be increased?

skynavga: If `tts:lineHeight` is `normal` or a positive length, then child inline areas must fit within a line area (or inline area). Ruby text containers effectively map to inline area children. These are basic XSL-FO and CSS semantics for line stacking.

palemieux:
Built at https://rawgit.com/w3c/ttml2/issue-0806-line-height-build/index.html

spec/ttml2.xml
```diff
@@ -17550,14 +17564,74 @@ the recommendations defined by <loc
...
+<item><p>when the computed value of <loc href="#style-attribute-lineHeight">tts:lineHeight</loc> of a paragraph is <code>normal</code>, then (1) each
+<loc href="#terms-line-area">line area</loc> of a paragraph exactly abuts and does not intersect with the preceding
+<loc href="#terms-line-area">line area</loc> of that paragraph, and (2) the <emph>line height</emph> of the
+<loc href="#terms-line-area">line areas</loc> of a paragraph vary in accordance with the styled content of each
+<loc href="#terms-line-area">line area</loc>;</p></item>
+<item><p>when the computed value of <loc href="#style-attribute-lineHeight">tts:lineHeight</loc> of a paragraph is
+a specified <loc href="#style-value-length">&lt;length&gt;</loc>, then the
+<loc href="#terms-line-area">line areas</loc> of a paragraph exactly abut one another but their content may (1) intersect with the preceding
+<loc href="#terms-line-area">line area</loc> of that paragraph (when negative <emph>half leading</emph> applies) or (2) consume more space
+in the <loc href="#terms-block-progression-dimension">block progression dimension</loc> than required (when positive <emph>half leading</emph>
```
palemieux: It would be good to explain why, as illustrated [here](https://codepen.io/palemieux/pen/aKvZBg), line boxes can expand even when `tts:lineHeight="100%"`, but not completely so as to fit the contents of the inline areas.

skynavga: @palemieux why should we explain this case? this is an informative section, why hold up the PR now when we are trying to get a CfC out the door? we can update this at any time, all the way up to the time we do the Proposed Recommendation request to W3M; in any case, we have already pointed to the relevant language in XSL-FO 4.5:

image

what exact change to the current text would you like?

palemieux: > why hold up the PR now when we are trying to get a CfC out the door?

This is a pretty extensive last minute change, and I am merely trying to figure out (a) what the objective is , and (b) does it achieve the objective.

skynavga:
On Mon, Jun 11, 2018 at 3:29 PM, Pierre-Anthony Lemieux <

>
> In spec/ttml2.xml
> <https://github.com/w3c/ttml2/pull/825#discussion_r194554021>:
>
> > +<bibref ref="xslfo11"/>, &sect;6.5.2 (when applied to a <loc href="#terms-line-area">line area</loc>). For a given
> +<loc href="#terms-line-area">inline area</loc>, analogous treatment applies, although this is not explcitly indicated by
> +<bibref ref="xslfo11"/>, but can be inferred from language found in <bibref ref="css2"/>, &sect;10.8.1.</p>
> +<p>In summary, the behavior defined by the <code>line-height</code> line stacking strategy, as defined by <bibref ref="xslfo11"/>, &sect;4.5,
> +and adopted by TTML, is intended to be compatible with the line stacking strategy of <bibref ref="css2"/>, which means that:</p>
> +<ulist>
> +<item><p>when the computed value of <loc href="#style-attribute-lineHeight">tts:lineHeight</loc> of a paragraph is <code>normal</code>, then (1) each
> +<loc href="#terms-line-area">line area</loc> of a paragraph exactly abuts and does not intersect with the preceding
> +<loc href="#terms-line-area">line area</loc> of that paragraph, and (2) the <emph>line height</emph> of the
> +<loc href="#terms-line-area">line areas</loc> of a paragraph vary in accordance with the styled content of each
> +<loc href="#terms-line-area">line area</loc>;</p></item>
> +<item><p>when the computed value of <loc href="#style-attribute-lineHeight">tts:lineHeight</loc> of a paragraph is
> +a specified <loc href="#style-value-length">&lt;length&gt;</loc>, then the
> +<loc href="#terms-line-area">line areas</loc> of a paragraph exactly abut one another but their content may (1) intersect with the preceding
> +<loc href="#terms-line-area">line area</loc> of that paragraph (when negative <emph>half leading</emph> applies) or (2) consume more space
> +in the <loc href="#terms-block-progression-dimension">block progression dimension</loc> than required (when positive <emph>half leading</emph>
>
> why hold up the PR now when we are trying to get a CfC out the door?
>
> This is a pretty extensive last minute change, and I am merely trying to
> figure out (a) what the objective is , and (b) does it achieve the
> objective.
>

This is not a change at all! How could it be since the entire section is
marked as non-normative?

spec/ttml2.xml
skynavga: Did you see the sentence at the beginning of this new section that says the section is _non-normative_? Therefore, there is no change here.

skynavga: @palemieux in any case, if these changes were normative (which they are not), then what behavioral change do you think this text implies?

skynavga:
@palemieux regarding
>Since tts:lineHeight="normal" is intended to function like CSS2, is the following prose still needed:

that is an independent question, unrelated to the intent of the current issue and PR; furthermore, removing the text you refer to would be a substantive change

skynavga:
@palemieux regarding any discussion of CSS and the CSS line-height property, note well ISSUE 1 [1] (emphasis added):

>This section is being rewritten. Refer to section 10.8 of [CSS2] for the normative CSS definition or the 2002 Working Draft if you want pretty pictures. (*But ignore the old text, half of it’s wrong. We’re not specifying which half, that’s to be determined.*) The CSS2 specification should be used as the guideline for implementation.

[1] https://drafts.csswg.org/css-inline/#line-height

This makes any discussion of what CSS does (in spec and implementations) rather speculative in nature, which is to say, something we cannot rely upon for the foreseeable future.

skynavga:
@nigelmegitt if you don't have any outstanding unresolved comments, then please approve

skynavga:
@nigelmegitt thanks; i'm having a call with @palemieux later this morning to resolve any changes he wants to see in place at this time;

nigelmegitt:
@skynavga thanks for the info, I hope you and @palemieux can get this to an approved state and merge so I can issue the CfC.

tairt:
After review I found a couple of issues that may need corrections or enhancements. I am not quite sure about the process. This pull request has been merged early on the base of a Call of Consensus. As I was absent I could not participate in the corresponding review. I could re-open the PR. But as it is merged already it possibly makes no sense. I can also open a ticket for each of the required changes.

Some issues I found/text that needs discussion (all [Section 11.3.2.2 - Line Stacking Strategy]( https://w3c.github.io/ttml2/index.html#semantics-line-stacking)):

* The term `expanded-area of each inline area` needs to be replaced with `expanded-**rectangle** of each inline-area`. It should match terminology used in XSL-FO.

* > Note well that the half leading applied to a line area and the half leading applied to an inline area child of that line area are often distinct values, 

I doubt that there is half leading applied to the line area at all. Regarding line area construction it is only applied to inline areas.

* > when the computed value of tts:lineHeight of a paragraph is a specified <length>, then the line areas of a paragraph exactly abut one another but their content may (1) consume less space in the block progression dimension than required to fit its content, in which case that content may intersect with the preceding line area of that paragraph or (2) consume more space in the block progression dimension than required to fit its content.

a) The text is difficult to read and seems to be circular:  When content consumes less space in a line area then needed to fit into that  line-area one consequence could be that it consumes more space than needed to fit in that line area?

b) I can't really see a case where content exceeds the line area. Can someone give an example?

skynavga:
Once closed, a PR cannot be re-opened as far as I know. So the best process is to file a new issue labeled with the PR milestone. Since Section 11.3.2.2 is non-normative, we can make any changes we like up to the time of going to PR, since they would be editorial changes by definition.

skynavga:
@tairt regarding your  last question, here is an example on Chrome; since XSL-FO `line-height` stacking strategy matches (or at least is supposed to match) CSS, then such results are indeed possible:

screen shot 2018-06-24 at 5 14 39 am

[inline-area-taller-than-line-area.html.zip](https://github.com/w3c/ttml2/files/2130706/inline-area-taller-than-line-area.html.zip)

nigelmegitt:
@tairt Thanks for reviewing this - I agree with @skynavga that the best course of action is to raise one or more new issues (and possibly also open a pull request) please; the content should be marked as editorial.

skynavga:
@tairt just open a new issue for now; I will start a PR, at least to fix the typo you encountered; as for the rest of your comments, we have to come to a common understanding before we can tune the other language

tairt:
Ok. I will open new issues for my comments.

PR #829, opened 2018-06-12
Switch to CR entities, update dates in status.

skynavga:
For CfC review.

### DIFF
--- a/spec/entitiescr.dtd
+++ b/spec/entitiescr.dtd
@@ -2,10 +2,10 @@
 
 <!-- Date parameters -->
 <!ENTITY draft.year "2018">
-<!ENTITY draft.month "March">
-<!ENTITY draft.mm "03">
-<!ENTITY draft.day "13">
-<!ENTITY draft.dd "13">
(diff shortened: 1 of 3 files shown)


spec/entitiescr.dtd
```diff
@@ -2,10 +2,10 @@
...
 <!-- Date parameters -->
 <!ENTITY draft.year "2018">
-<!ENTITY draft.month "March">
-<!ENTITY draft.mm "03">
-<!ENTITY draft.day "13">
-<!ENTITY draft.dd "13">
+<!ENTITY draft.month "June">
+<!ENTITY draft.mm "06">
+<!ENTITY draft.day "25">
+<!ENTITY draft.dd "25">
```
nigelmegitt: This is one day before the end of the Call for Consensus ends, so it will certainly need to be edited again, to match the expected publication date.

skynavga:
This was based on your starting the CfC this past Monday, which didn't happen. In any case, it has been updated to 06/28 in #831.

PR #833, opened 2018-06-14
Improve constraints on 'tts:rubyPosition' usage (#832).

### LINKED ISSUE #832: Applicability of tts:rubyPosition when tts:ruby="text".
If a ruby text is not a child of a ruby textContainer, then `tts:rubyPosition` specified on the ruby text applies to the implicit parent ruby textContainer.

If a ruby text is a child of a ruby textContainer that specifies `tts:rubyPosition`, then `tts:rubyPosition` cannot be specified on the ruby text.

What happens if `tts:rubyPosition` is specified on ruby text is the child of a ruby textContainer for which `tts:rubyPosition` is not specified? Does it override the computed value of `tts:rubyPosition` on the parent ruby textContainer? What happens if different children of a ruby textContainer specify different `tts:rubyPosition` values -- which one applies the parent ruby textContainer?

Suggest expanding the rule to state that `tts:rubyPosition` cannot be specified on a ruby text that has an explicit parent ruby textContainer.

skynavga:
Closes #832.


skynavga:

PR #843, opened 2018-06-19
Refactor #textEmphasis* features to avoid negative definitions (#838,…

### LINKED ISSUE #838: Typo at #textEmphasis-no-quoted-string.
`#textEmphasis-no-quoted-string` should be `#textEmphasis-no-color`

### LINKED ISSUE #841: Rework #textEmphasis* features.
Rework to avoid negative feature definitions, i.e., `-no-color` and `-no-quoted-string`.

skynavga:
…#841).

Closes #838.
Closes #841.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -19608,6 +19608,17 @@ if it satisfies the requirements of this appendix with respect to the definition
 feature designation <emph>F</emph> as pertains to transformation or
 presentation processing, respectively.</p>
 </note>
+<note role="interoperability">
+<p>If an implementation of a TTML processor does not support the semantics of some feature and the
+syntactic expression of that feature is known to the implementation, then it is recommended that the
+implementation be able to parse the expression but ignore the semantics of the unsupported components of the expression.</p>
+<p>For example, if <loc href="#feature-textEmphasis-minimal"><code>#textEmphasis-minimal</code></loc> is
+supported, but <loc href="#feature-textEmphasis-color"><code>#textEmphasis-color</code></loc> is not
+supported, then values of <loc href="#style-attribute-textEmphasis"><att>tts:textEmphasis</att></loc>
+that contain a <loc href="#style-value-emphasis-color">&lt;emphasis-color&gt;</loc> component should
+be parsed while ignoring the semantics of that unsupported component (but not ignoring the semantics of other
+supported components of the attribute value).</p>
+</note>
 <div2 id="feature-designations">
 <head>Feature Designations</head>
 <p>A feature designation is expressed as a string that adheres to the following
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -22037,7 +22037,7 @@ href="#style-attribute-textEmphasis"><att>tts:textEmphasis</att></loc>
 attribute which (1) do not contain an <loc href="#style-value-emphasis-color">&lt;emphasis-color&gt;</loc> component and
 (2) do not contain an <loc href="#style-value-emphasis-style">&lt;emphasis-style&gt;</loc> component that
 takes the form of a <loc href="#content-value-quoted-string">&lt;quoted-string&gt;</loc> value.</p>
-<p>A TTML <loc href="#terms-presentation-processor">presentation processor</loc> supports the <code>#textEmphasis-color</code>
+<p>A TTML <loc href="#terms-presentation-processor">presentation processor</loc> supports the <code>#textEmphasis-minimal</code>
 feature if it implements presentation semantic support for the same vocabulary enumerated above.</p>
 </div3>
 <div3 id="feature-textEmphasis-quoted-string">
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -22003,64 +22003,53 @@ feature if it supports the <code>#textDecoration-under</code> feature as defined
 </div3>
 <div3 id="feature-textEmphasis">
 <head>#textEmphasis</head>
+<p>A TTML processor supports the <code>#textEmphasis</code> feature if it
+supports the following features:</p>
+<ulist>
+<item><p><loc href="#feature-textEmphasis-color"><code>#textEmphasis-color</code></loc></p></item>
+<item><p><loc href="#feature-textEmphasis-minimal"><code>#textEmphasis-minimal</code></loc></p></item>
+<item><p><loc href="#feature-textEmphasis-quoted-string"><code>#textEmphasis-quoted-string</code></loc></p></item>
+</ulist>
+<p>The <code>#textEmphasis</code> feature is a syntactic and semantic extension of the
+<loc href="#feature-textEmphasis-minimal"><code>#textEmphasis-minimal</code></loc> feature.</p>
+<p>It is an error for a content profile to require or permit use of the <code>#textEmphasis</code> feature
+but prohibit use of any of the following features:
+<loc href="#feature-textEmphasis-color"><code>#textEmphasis-color</code></loc>,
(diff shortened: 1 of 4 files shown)


spec/ttml2.xml
```diff
@@ -22003,64 +22003,53 @@ feature if it supports the <code>#textDecoration-under</code> feature as defined
...
+of transforming values of the <loc
 href="#style-attribute-textEmphasis"><att>tts:textEmphasis</att></loc>
-attribute, the value of which does not contain a <loc href="#style-value-emphasis-style">&lt;emphasis-style&gt;</loc> component that
-takes the form of a <loc href="#content-value-quoted-string">&lt;quoted-string&gt;</loc> value, or, if
-it contains said component value, then that component value is treated as if <code>auto</code> were specified.</p>
-<p>A TTML <loc href="#terms-presentation-processor">presentation processor</loc> supports the
-<code>#textEmphasis-no-color</code> feature if it implements presentation semantic support
-for all defined values of the <loc
+attribute which (1) do not contain an <loc href="#style-value-emphasis-color">&lt;emphasis-color&gt;</loc> component and
+(2) do not contain an <loc href="#style-value-emphasis-style">&lt;emphasis-style&gt;</loc> component that
```
nigelmegitt: This formulation apparently prohibits `tts:textEmphasis` values from including either a color or quoted string emphasis-style component if this feature is supported, but I don't believe that is intended or desirable: the result we want to end up with is that those components may be included if the processor supports the related features.

Proposal: specify this in terms of which parts of the syntax the processor must minimally support rather than which parts of the syntax may be present. For example:

> A TTML transformation processor supports the `#textEmphasis-minimal` feature if it recognizes and is capable of transforming the following components of values of the `tts:textEmphasis` attribute:
> * `<emphasis-position>` and
> * `<emphasis-style>` excluding `<quoted-string>` values.

nigelmegitt:
Additional review comment: you could also add an `extends="#textEmphasis-minimal"` attribute value on the profile entry for `#textEmphasis`.

spec/ttml2.xml
skynavga: @nigelmegitt you seem to have neglected the language "recognizes and is capable of transforming values ..."; therefore, it is already formulated in terms of what is supported; if color or quoted string values are present, but their transformation is not supported, then that usage is not supported, but it doesn't read on whether they may be present or not in content; so,  IMO, there is no problem here (except a typo I am just now fixing)

skynavga:
@nigelmegitt fixed typo and missing extends

nigelmegitt:
@skynavga ok they're good changes, what about my proposal at https://github.com/w3c/ttml2/pull/843#discussion_r196716082 ?

spec/ttml2.xml
skynavga: @nigelmegitt your reading is not supported by the language, which is consistent with other feature definitions of this form, e.g., `#condition-primary`, `#fontSize-isomorphic`, `#textOutline-unblurred`.

nigelmegitt: @skynavga I disagree, or at least I'm seeing another potential reading that we should avoid by editing. The wording:

> ... it recognizes and is capable of transforming values of the tts:textEmphasis attribute which (1) do not contain an `<emphasis-color>` component and (2) do not contain an `<emphasis-style>` component that takes the form of a `<quoted-string>` value. 

suggests that the entire values which *do* contain either `<emphasis-color>` or (`<emphasis-style>` that is a `<quoted-string>`) or both do not need to be recognised or transformed at all by a processor that supports the feature.

I think it needs to say that the included components would be recognised or transformed, but say nothing about what happens to components that are not included in the feature.

In other words, how should a processor that supports the feature but not any of the other `#textEmphasis-x`features deal with, say, `tts:textEmphasis="auto blue before"`? I'm expecting it to recognise the `auto` and the `before` but do something unspecified with the `blue`, such as ignore it. But the current wording reads to me that the processor will fail to recognise or transform the *entire* value, even the bits it might be able to process.

skynavga: >suggests that the entire values which do contain either `<emphasis-color>` or (`<emphasis-style>` that is a `<quoted-string>`) or both do not need to be recognised or transformed at all by a processor that supports the feature.

that's correct, and the desired semantics; we do not require (elsewhere) a processor to parse a construct containing unsupported syntax, though we could do so, but we have not done so yet, and would need to do the same in other similar contexts;

of course, a processor that knows about some possible construct but that only (semantically) supports part of it can be constructed, and that is a good approach for implementers; however, this doesn't protect one against new constructs not known at the time of implementation

>I think it needs to say that the included components would be recognised or transformed, but say nothing about what happens to components that are not included in the feature.

we do not specify this elsewhere in similar contexts; e.g., we don't say that the blur radius component must be parsed (or transformed) by a processor that supports `#textOutline-unblurred` but not `#textOutline-blurred`

>In other words, how should a processor that supports the feature but not any of the other `#textEmphasis-x` features deal with, say, `tts:textEmphasis="auto blue before"`? I'm expecting it to recognise the auto and the before but do something unspecified with the blue, such as ignore it. But the current wording reads to me that the processor will fail to recognise or transform the entire value, even the bits it might be able to process.

right now, we don't say either way (in general); robust implementations will generally parse known syntax the semantics of which are unsupported; I would not agree that the current language implies it will or must fail to recognize or transform; rather, it is silent on this matter, which is consistent with the other like features I enumerated

nigelmegitt: Hmm, okay, I might be persuadable. Any other views? @palemieux perhaps?

skynavga:
@nigelmegitt I think your proposal could be worded so as to be equivalent, but I think it does not add anything or improve the precision of the specification, so I am inclined to stay with the current (PR) text which is more consistent with the other uses I have cited;

on the general question of whether we should say more than nothing (about parsing known syntax that is not semantically supported), I wouldn't object to adding a note in the front matter (of E.1) that recommends parsing but ignoring unsupported value components

nigelmegitt:
> I wouldn't object to adding a note in the front matter (of E.1) that recommends parsing but ignoring unsupported value components

I'd certainly support that.

skynavga:
@nigelmegitt added such a note in https://github.com/w3c/ttml2/pull/843/commits/8426f59fa41b1524ca80efc9e97f1ec829f66e47

nigelmegitt:
> added such a note in 8426f59

@skynavga 👍

spec/ttml2.xml
nigelmegitt: > we do not require (elsewhere) a processor to parse a construct containing unsupported syntax

Actually we don't say anything about it at all, so in that sense the text that triggered this thread is unusual. There are only two other `#xxx-minimal` features currently:

* `#rubyAlign-minimal` talks about the recognised values that the processor is capable of transforming.
* `#animate-minimal` uses the phrasing "if it recognizes and is capable of transforming the semantics of the `animate` element when used with the discrete and linear values of the `calcMode` attribute."

One difference here is that the syntax they reference does not have multiple components, whereas `tts:textEmphasis` does. It would be more consistent to relate this `#textEmphasis-minimal` feature to the supported values of specific components of the `tts:textEmphasis` attribute. Looked at this way, I'm back to preferring my proposal, since it does exactly that.

skynavga:
@nigelmegitt since the current text is not broken and since it is consistent with the type of language expressed elsewhere, e.g., see `#condition-primary`, then I insist on following the editor's preference

palemieux:
Built at https://rawgit.com/w3c/ttml2/issue-0841-text-emphasis-feature-build/index.html#feature-textEmphasis

skynavga:

skynavga:

PR #848, opened 2018-06-21
Add `#region-implied-animation` to the at risk list.

### LINKED ISSUE #847: Add `#region-implied-animation` to the at risk list
From the discussion in today's TTWG meeting in relation to w3c/imsc#366, it seems that the usage of `tts:origin` and `tts:extent` on content elements is becoming less prevalent, and that is a direction worth encouraging. In addition, some ambiguities have been noted in #820.

I would therefore like to add this feature to the "at risk" list for CR2.

nigelmegitt:
Closes #847


skynavga:

PR #851, opened 2018-06-21
Clarify that height term can mean horizontal axis in exceptional case…

### LINKED ISSUE #830: Add clarifying note to definition of [height] term.
Add note under definition of [height] pointing out exceptional interpretation of _height_ as horizontal in the context of certain usage scenarios with ``tts:lineHeight`` and the phrase _line height_.

skynavga:
…s (#830).

Closes #830.


skynavga:

PR #852, opened 2018-06-21
Clarify semantics of implied inline animation of the same style on re…

### LINKED ISSUE #820: Implied region animation ambiguity.
According to https://w3c.github.io/ttml2/#layout-vocabulary-region-special-inline-animation-semantics, if you specify tts:extent on a `p` the behavior is as if an animation had been created. The text also says:
>if multiple animation elements (Ai, ..., Aj) already apply to R, then A applies subsequent to these the last of these existing animations, namely, immediately after Aj.

Aside from the typo ("to these the last" - > "to the last"), it is not clear what happens if two `p` have the same begin/end and refer to the same region but also use two different extent attributes. 

One ambiguous point is the order following which the animations are added. Is it document order? I would think so but it is not so clear.

The second ambiguous point is: how to apply the two animations? The text says "subsequent" which would not make sense because the paragraphs will not be active once the first animation is done, so the subsequent animation will never have any effect. SMIL defines the `additive` attribute to resolve this case and the default value is `replace`, meaning that in this case the last paragraph in document order sets the extent of the single region used by both paragraph. 

Of course, this is assuming that there is only one region being processed, that the animations don't create temporary regions.

skynavga:
…gions (#820).

Closes #820.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -16932,13 +16932,22 @@ the following style attributes apply exceptionally to the <el>region</el> <emph>
   <item><p><loc href="#style-attribute-origin"><att>tts:origin</att></loc></p></item>
   <item><p><loc href="#style-attribute-position"><att>tts:position</att></loc></p></item>
 </ulist>
-<p>In such cases, the application of these style attributes are considered to be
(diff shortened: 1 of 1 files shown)


skynavga:

PR #867, opened 2018-06-28
Clarify constancy of ECP and EPP (#860).

### LINKED ISSUE #860: Clarify that construct effective {content,processor} profile is performed only once.
Add notes in §5.2.4.2 and §5.2.4.3, under the definitions of **[construct effective content profile]** and **[construct effective processor profile]**, respectively, that these procedures are performed only once during the life cycle of a document, i.e., that if they are invoked more than once, then subsequent invocations obtain the results produced by the first invocation (the values are memoized).

skynavga:
Closes #860.


skynavga:
@nigelmegitt please propose a change that would permit adding a note that addresses the original issue, e.g., you may propose adding language that carves out some space where the behavior you imagine is possible

nigelmegitt:
> please propose a change that would permit adding a note that addresses the original issue, e.g., you may propose adding language that carves out some space where the behavior you imagine is possible

@skynavga I would much prefer to simply _do nothing_ here. I think this is a case where less is more and we should just stay silent on the matter.

skynavga:
@nigelmegitt doing nothing does not address the issue; so, please be so kind as to propose qualifying language language that addresses your generalized fear of saying something wrong.

skynavga:
@nigelmegitt regarding defining a "processing session", I believe we already do that (implicitly) in the sense that all processing defined by TTML (until now) is scoped by a single processing session; in other words, any implementation that chooses to map this single processing session onto multiple possibly distinct sub-processing sessions is going outside the scope of TTML (as such) to do so

nigelmegitt:
> we already do that (implicitly)

@skynavga That's one possible reading, sure. Without stating it explicitly though, I think it's asking too much of the reader to deduce with any certainty that is the intent. Most readers will consider a "processor" to be synonymous with an "implementation".

In particular the definition of "content processor" refers to the entire processing system. 

The word "implementation" appears 61 times in the document and is not defined, though it does have conformance language associated with it in some places. That word is not used in this pull request, but it seems relevant.

skynavga:
@nigelmegitt we give an operational definition of content processor in 3.2.1:

>A TTML2 content processor conforms to this specification if the following generic processor criteria are satisfied ...

this establishes an explicit scope on what aspects of content processing are relevant, and also (implicitly) delimits the scope of a processing "session" as such

since there is no procedure under the scope of 3.2 that may sub-divide a document instance into segments each of which are processed as a distinct processing session and there is, likewise, no procedure that would, for example, attempt to perform multiple invocations of [validate document] each based on a distinct EVP, then any such processing is clearly out of scope of the operational definition of content processor

as such, I fail to conceive how one may reasonably suppose that such form of processing [multiple invocations of validation] is in scope, and, thereby, demand that EVP (or ECP/EPP) be interpretable as non-constant **in the context of what we DO explicitly define as content processing**.

skynavga:
@nigelmegitt giving this some more thought, it occurs to me that the behavioral possibility you mention (attempting to use multiple profiles to validate a document before picking one as the EVP) is already possible with the current language combined with the changes proposed here; in particular, there is nothing to prevent an implementation of a content processor from attempting to use multiple profiles and performing multiple validation attempts in steps (1) of **[construct effective content profile]** and/or **[construct effective processor profile]**, then using those results (setting ECP and EPP) in the determination of EVP in **[construct effective validation profile]**; in particular, it doesn't matter what internal, provisional values of a content/processor/validation profile are used, or whether they stay constant, as long as by the time that ECP/EPP/EVP are set (for the processing of a given document instance), the are no longer changed while performing subsequent processing of that document instance; note that "that document instance" may be of whatever granularity an author or processing system wishes; e.g., a processing system might have previously segment another, larger document instance into smaller segments, and each of these segment document instances might be processed multiple times in the steps (1) noted above in order to determine a final ECP/EPP/EVP for each segment document instance; such a segmentation process as well as performing multiple passes to determine a final ECP/EPP/EVP for each segment is entirely outside the scope of TTML as defined, but is not prevented either by the current spec text or the changes proposed here; the only intent of these changes is _for the purpose of processing a particular document instance (in the defined scope), then the values of ECP/EPP/EVP **in the in-scope processing model** are constant once set by the defined processing procedures_; this semantic does not prevent the out-of-scope processing you have described

skynavga:

PR #871, opened 2018-06-30
Clarify inheritance of 'normal' line height with further resolved com…

### LINKED ISSUE #870: Clarify inheritance of line height 'normal'.
It looks like https://github.com/w3c/ttml1/issues/220 never got listed in https://github.com/w3c/ttml2/issues/358, and we didn't catch this in review.

Note that CSS2.1 (2011) [1] already says as much, and we normatively reference CSS2.1 in TTML2 (emphasis added):

**normal**
Tells user agents to set the used value to a "reasonable" value based on the font of the element. The value has the same meaning as `<number>`. We recommend a used value for 'normal' between 1.0 to 1.2. _The computed value is 'normal'._

[1] https://www.w3.org/TR/2011/REC-CSS2-20110607/visudet.html#leading

skynavga:
…puted value processing (#870).

Closes #870.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -10945,7 +10945,7 @@ descendants of a <loc href="#content-vocabulary-p"><el>p</el></loc> element is u
 <loc href="#terms-inline-area">inline areas</loc>, about which see <bibref ref="xslfo11"/>, &sect;4.5, and <bibref ref="css2"/>, &sect;10.8.1.</p>
 </note>
 <p>If the value of this attribute is <code>normal</code>, then the
-computed value of this style property is determined as follows:</p>
+resolved computed value of this style property is determined as follows:</p>
 <olist>
 <item><p>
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -10974,9 +10974,16 @@ where scaled(<emph>X</emph>) denotes font metric <emph>X</emph> scaled according
...
 </p></item>
 <item><p>
-Set the computed value of this style property to <emph>LH</emph>.
+Set the resolved computed value of this style property to <emph>LH</emph>.
 </p></item>
 </olist>
+<note role="clarification">
+<p>When the specified value of this attribute is <code>normal</code>, the computed value is the specified value; therefore,
+when applying inheritance semantics, the value <code>normal</code> is the inherited value.
+However, when this compute value is to be used to perform paragraph layout, it is further processed to obtain a resolved computed value (also called a
```
cconcolato: Typo. Should be "computed"

skynavga:
@cconcolato fixed

PR #875, opened 2018-07-01
Add tts:display 'inlineBlock' value to schemas; update rnc schemas to…

### LINKED ISSUE #874: Schemas are missing tts:display 'inlineBlock' value.
This is an editorial change since isolate is already defined in spec, which is takes precedence.

In addition, update EOL convention in RNC schemas to use unix EOLs.

skynavga:
… use unix end of line convention (#874).

Closes #874.


skynavga:
@nigelmegitt see https://github.com/w3c/ttml2/pull/875#pullrequestreview-133775213

PR #886, opened 2018-07-12
Note potential for false negatives with [validate document] (#876).

### LINKED ISSUE #876: Clarify that [validate document] may produce false positive results.
Add note under §5.3.1.2 [validate document] that, since the effective validation schemas are implementation dependent, false positive results may occur. In particular, describe false positive example where the use of `prohibited` in a content profile does not necessarily lead to a validation error in the case that a prohibited feature is used in a document but the effective validation schemas are overly general in their definition.

skynavga:
Closes #876.


skynavga:
@nigelmegitt addressed comments, please re-review

spec/ttml2.xml
```diff
@@ -3338,6 +3338,24 @@ by the <loc href="#terms-effective-validation-schemas">effective validation sche
 <termref def="defs-validation-error">validation error</termref> condition applies. If any of the above verification steps detect a state of
 non-conformance with respect to a recommendation (should) or disrecommendation (should not), then a
 <termref def="defs-validation-warning">validation warning</termref> condition applies.</p>
+<note role="elaboration">
+<p>Since the determination of <loc href="#terms-effective-validation-schemas">effective validation schemas</loc> is implementation
+dependent, and since there is no requirement that the <loc href="#terms-effective-validation-schemas">effective validation schemas</loc> detect
+every possible <termref def="defs-validation-error">validation error</termref>, then performing the
+<loc href="#semantics-procedure-validate-document">[validate document]</loc> procedure need not produce a
+<loc href="#semantics-validation-state-validation-exception">validation exception</loc>, in which case,
+the results of <loc href="#terms-validation-processing">validation processing</loc> may be considered a <emph>false positive</emph>.</p>
```
nigelmegitt: By the way, using the current TTML schemas typically generates false negatives when using semicolon-delimited lists of values in style attributes in an animation context, except where the schema just defines the syntax to be a string.

skynavga: Thanks for pointing this out. Perhaps I will swap it for the current discussion that uses an RNC example.

skynavga:

PR #890, opened 2018-07-13
Clarify use of terms altitude, depth, and line gap; add non-normative…

### LINKED ISSUE #889: Improve consistency of terminology w.r.t. line height computation.
In line height text, improve terminological consistency; also add note referring to OpenType realization.

skynavga:
… reference to open font format (#889).

Closes #889.


skynavga:
See https://rawgit.com/w3c/ttml2/issue-0889-line-height-build/index.html.

skynavga:
@tairt regarding https://github.com/w3c/ttml2/pull/890#pullrequestreview-137681500, while I appreciate the intent of your comments, I cannot accept any of them, for the following reasons:

I am not willing to change the reference from an international standard to a company specific document that does not follow international standards practices; furthermore, the other ISO references we have in TTML also point to the same generic www.iso.org URL; readers are expected to know how to find the relevant document at ISO

as for the sign of sTypeDescender, it is not necessarily negative, but it follows the standard convention in coordinates/dimensions in TrueType/OpenType that Y values pointing up are positive and Y values pointing down are negative (here the point of reference is the baseline - for up and down); any reader who bothers reading this comment is expected to be sufficiently knowledgable about OT to understand this, so there is no need to further elaborate

as for referencing XSL-FO text-{altitude,depth}, this is important because we normatively refer to the line area model which uses these terms, and since we are here using altitude (A) and depth (D); furthermore, the mapping of altitude and depth here is consistent with the 3rd note of CSS 2.2 §10.8.1; the fact that XSL-FO does not provide similar guidance (as CSS and as we do here) does not contradict XSL-FO and corresponds to actual implementations, e.g., see [1]

[1] https://github.com/apache/fop/blob/trunk/fop-core/src/main/java/org/apache/fop/fonts/truetype/OpenFont.java#L1569

[2] https://github.com/apache/fop/blob/trunk/fop-core/src/main/java/org/apache/fop/layoutmgr/inline/AlignmentContext.java#L455

[3] https://github.com/apache/fop/blob/trunk/fop-core/src/main/java/org/apache/fop/layoutmgr/inline/AlignmentContext.java#L185

skynavga:
@tairt btw, you missed a significant COMMA in the text (before _and `sTypoLineGap`_), since your suggested change:

>Improve readability of the sentence e.g. `sTypoAscender`, minus one times `sTypoDescender` and `sTypoLineGap` fields

would be wrong, i.e., the negation of `sTypoDescender` does not apply to `sTypoLineGap`

skynavga:
@tairt earlier you had asked for references to browser code on this subject, here is an excerpt from Firefox code I came up with quickly just now in a brief search:

From [1]:

```
311         mMetrics.emAscent = os2->sTypoAscender * yScale;
312         mMetrics.emDescent = -os2->sTypoDescender * yScale;
313         FT_Short typoHeight =
314             os2->sTypoAscender - os2->sTypoDescender + os2->sTypoLineGap;
315         lineHeight = typoHeight * yScale;
```

[1] https://dxr.mozilla.org/mozilla-central/source/gfx/thebes/gfxFT2FontBase.cpp#311-315

Note that `mMetrics.emAscent` is our _scaled(A)_ and `mMetrics.emDescent` is our _scaled(D)_. Note also that `lineHeight` corresponds exactly with our _set LH to the sum of scaled(A), scaled(D), and scaled(G)_.

tairt:
> I am not willing to change the reference from an international standard to a company specific document that does not follow international standards practices; furthermore, the other ISO references we have in TTML also point to the same generic www.iso.org URL; readers are expected to know how to find the relevant document at ISO.

It is important to guide implementers. Information to implement TTML needs to be as accessible as possible. Therefore I am in favor to point to the Microsoft publication of the specification. 

Please see also this reference:

https://www.w3.org/TR/2018/CR-css-fonts-3-20180626/#ref-OPENTYPE

It make sense to point to the same resource as CSS.

tairt:
> as for referencing XSL-FO ...

To spead up the process I will not object your current text.

tairt:
> as for the sign of sTypoDescender

To spead up the process I will not object your current text.

skynavga:
@tairt it turns out there is a free of charge PAS version of OFF available from ISO; I have added a link to it in 0ce8c510d1571f92b9ff183d988992091696e8c6; which resolves your point about free of charge version of the reference

skynavga:
@tairt updated further in 1df13044ed92eb6bfde8da2fc74391fe542df223. see https://rawgit.com/w3c/ttml2/issue-0889-line-height-build/index.html for build.

skynavga:
@tairt ready for your approval for this to make the CR3 cut

skynavga:
@tairt ping

skynavga:
@tairt please re-review, otherwise we may be forced to dismiss your review to progress this issue

tairt:
Thanks

nigelmegitt:

PR #898, opened 2018-07-15
Ensure audio and image (in presentation contexts) are treated as cont…

### LINKED ISSUE #891: Add audio, image in presentation contexts to content elements.
I reviewed all 99 instances of references to the term _content element_, and all of them should include and not exclude audio and image descendants of body. Furthermore, these elements need to be added to 10.4.4.4 step (1) and 10.2.1 (the list of elements on which `@style` applies - image was already present, but missing audio).

This is an editorial change to correct an inadvertent omission which is substantially backed up by normative text elsewhere, e.g., both audio and image accept style properties which are applied to them, and, therefore, style resolution must apply (in contradiction to the [filter] step of 10.4.4.4 and the [associate region] procedure (defined in terms of content elements). In addition, presentation processing explicitly applies, e.g., see [presentation related element].

skynavga:
…ent elements (#891).

Closes #891.


skynavga:
Taking a conservative position, I am relabeling this as substantive.

skynavga:

PR #900, opened 2018-07-15
Improve definition of conditionally excluded (included) (#893).

### LINKED ISSUE #893: Refine conditional {ex,in}clusion terms.
+ section 2.2

   * [conditionally excluded (or included)] clarify
   that the descendants of a conditionally excluded element are
   all conditionally excluded;

   * [conditionally excluded (or included)] clarify that an element
   is conditionally included if it is not conditionally excluded;

   * introduce sub-term _unconditionally included_

skynavga:
Closes #893.


skynavga:
per @nigelmegitt, see https://github.com/w3c/ttml2/issues/893#issuecomment-405932247, which should have been made here

skynavga:
@nigelmegitt comments addressed, please re-review

spec/ttml2.xml
```diff
@@ -6218,6 +6218,10 @@ value evaluates to <code>false</code>:</p>
 <item><p>the <loc href="#semantics-style-association-chained-referential">chained referential semantics</loc> of a
 <loc href="#styling-vocabulary-style"><el>style</el></loc> element are not ignored;</p></item>
 </ulist>
+<note role="clarification">
+<p>The exceptions noted above also apply to <loc href="#terms-conditionally-excluded">conditionally excluded</loc> descendants of a
```
nigelmegitt: The inclusion of "conditionally excluded" before "descendants" here suggests that it might be possible to conditionally _include_ descendants of a conditionally excluded element, but this is not the case. Suggest removing the qualifier so it just reads "The exceptions noted above also apply to descendants of a conditionally excluded element".

nigelmegitt: Raised #926.

nigelmegitt:

PR #908, opened 2018-07-15
Fix typos and styling (#894).

### LINKED ISSUE #894: Fix typos or styling.
+ section 2.2

   * [speech data resource] typo "that consist" should be "that
   consists";

 + section 4

   * item (3) grammar is not quite right w.r.t. plurality: change
   "namespace URI" to "namespace URIs"
   * example fragment in note has poor whitespace formatting

 + section 10.4.1.1

   * change "TT Style Namespace" to "TT Style Namespaces" (also
   review other uses of the former that may need the same change;

 + section 10.4.4.2

   * step (6)(b), change "not the region element" to "not a region
   element";

   * step (6)(c), change "is the region element" to "is a region
   element";

 + section 11.3.1.3

   * change "for the purposes of performing" to "for the purpose of
   performing"; check for need to make this change elsewhere as well

 + section 11.3.1.4

   * in step (2), embolden "[construct intermediate document]" in
   note;

 + section 11.3.2.2

   * in para (4), change "which is, respectively" to "which are,
   respectively"; change "a parameter" to "(1) a parameter" and
   "the dimensions" to "(2) the dimensions";

   * in para (5), 2nd bullet, change "their content" to "their
   content rectangles" and change "to fit its content" to "to
   fit their content" (two instances, in both (1) and (2));

skynavga:
Closes #894.


spec/ttml2.xml
```diff
@@ -16557,7 +16541,7 @@ next style property;</p>
 <!-- (b) -->
 <item>
 <p>if <emph>P</emph> is inheritable and <emph>E</emph> is
-not the <loc href="#layout-vocabulary-region"><el>region</el></loc> element, then perform the following:</p>
+not a <loc href="#layout-vocabulary-region"><el>region</el></loc> element, then perform the following:</p>
```
palemieux: There is only ever one `region` element in the style resolution tree, right?

nigelmegitt: Just guessing that this is a hangover from root inheritance where this would have read "the `tt` element". 

I can't see that "the region element" refers to any specific region element from the text, so I agree it should say "a region element", or to be consistent with the text in step 6, we could say "and the element type of E is `region`".

spec/ttml2.xml
```diff
@@ -17225,7 +17225,7 @@ and the resulting generated out-of-line region is shown by the following example
 &lt;tt xml:lang="" xmlns="http://www.w3.org/ns/ttml"&gt;
   &lt;head/&gt;
   &lt;body&gt;
-    &lt;div begin="5s" end="15s"&gt;
+    &lt;div begin="5s" dur="10s"&gt;
```
nigelmegitt: Why change from `end` to `dur`?

skynavga: Read step (3)(a) of [process inline regions]. `dur` didn't conform to that step.

skynavga: @nigelmegitt please re-review these changes  under #911, to which I have move the inline regions related changes

skynavga:

PR #909, opened 2018-07-15
Clarify computed cell size; improve language pertaining to pixels (#8…

### LINKED ISSUE #896: Clarify computed cell size, improve pixel related term definitions and usage.
Add note under definition of [computed cell size] indicating that the computed cell size is expressed in (logical) pixels. This follows from the definition, i.e., computed cell size width is root container width divided by column count, etc. [See also Appendix H.2 _Resolution_.]

Add `term{def,ref}` wrappers for pixel related terms. Improve language around use of column and row terms, which we use both for the cell grid and the (logical) pixel grid.

skynavga:
…96).

Closes #896.


skynavga:
@palemieux btw, I suggest checking the "Hide whitespace changes" box under the Diff Settings, as there were a few paragraphs in the source rewrapped.

nigelmegitt:
Just following up as Chair on https://github.com/w3c/ttml2/pull/909#pullrequestreview-137282063 our agreement when such a request is made is not to merge early, i.e. the normal decision policy period applies.

skynavga:
I am more interested in discussing the substance of the changes than editing process policy. Please try to review the substance ASAP.

skynavga:
@palemieux I'd like to wrap this up by COB on Monday to make it into CR3

nigelmegitt:
@palemieux My intention is to allow this to be merged so that it can be reviewed as part of the CfC we agreed that I'd issue today; I'll assume that's okay unless you say otherwise since you were on the call when this was implied, and the overall review period will effectively be extended rather than reduced.

nigelmegitt:
Correction, I need to issue the CfC tomorrow (preferably) or on Wednesday, otherwise I'll hit my vacation before managing to request the transition and publishing.

skynavga:

PR #911, opened 2018-07-16
Editorial improvements to 11.3.1.2 Inline Regions (#903).

### LINKED ISSUE #903: Editorial improvements to Inline Regions §11.3.1.2.
+ section 11.3.1.2

   * change "explicit inline" to "inline"; we no longer have implicit
   inline regions;

   * change formatting of Generated Out-of-line Region example to
   prevent column overflow;

   * change "computed" to "active" in step (3) bullet (5)(a)

   * add note under step (3)(a) that clarifies use of "dur"; change
   example to use end instead of dur with the newly generated region element;

   * additional minor language improvements;

skynavga:
Closes #903.


spec/ttml2.xml
```diff
@@ -17120,21 +17120,21 @@ specification and the content that references that region. Semantically, each
...
 <p>Inline regions are processed in accordance with the procedure
 <phrase role="strong"><loc href="#procedure-process-inline-regions">[process inline regions]</loc></phrase>
-which generates out-of-line regions that correspond with the
-explicit inline regions. This latter procedure additionally binds <loc href="#terms-content-element">content elements</loc> associated with inline regions
+which generates out-of-line regions that correspond with
+inline regions. This procedure additionally binds <loc href="#terms-content-element">content elements</loc> associated with inline regions
 to the corresponding generated out-of-line regions.</p>
 <note role="elaboration">
-<p>A <loc href="#terms-content-element">content element</loc> can be explicitly associated with only a single
+<p>A <loc href="#terms-content-element">content element</loc> can be associated with only a single
```
nigelmegitt: This is not true in the general case within a TTML document instance, but is true in the specific case of an ISD. For example the `body` element is generally associated with multiple regions assuming that it does not explicitly specify a region.

This change needs to be reverted I think.

spec/ttml2.xml
```diff
@@ -17180,24 +17180,12 @@ otherwise, let <emph>ID</emph> be the value of the <att>xml:id</att> attribute o
   as follows:</p>
   <olist>
     <item><p>if not operating in smpte time base and discontinuous marker mode, then set the values of these <att>begin</att> and <att>end</att>
-    attributes to values equivalent to the respective computed begin and end times of <emph>B</emph> within the root temporal extent;</p></item>
+    attributes to values equivalent to the respective active begin and end times of <emph>B</emph> within the root temporal extent;</p></item>
```
nigelmegitt: I think this was better as "computed". The closer term for this in SMIL 3 seems to be _global_, but SMIL's [Descriptive terms for times](https://www.w3.org/TR/2008/REC-SMIL3-20081201/smil-timing.html#Timing-DescriptiveTermsforTimes) section specifies that "active time" is:

> Time as measured relative to the element's active duration. A time is measured as an offset from the active begin of the element.

So by that definition this step is requiring that B's `begin` is transferred to R's `begin`. Except in the case that B has no ancestors that specify a `begin` this is incorrect. What is intended here is that the active begin of B, expressed in _document time_, i.e. a global time relative to the document syncbase, is transferred to R's `begin`. We have the term "document coordinate space" defined already which could be useful here.

For example this would be better:

```xml
attributes equivalent to the active begin and end times of <emph>B</emph> expressed as values in the
<loc href="#terms-document-temporal-coordinate-space">document coordinate space</loc>;</p></item>
```

spec/ttml2.xml
```diff
@@ -17276,6 +17265,17 @@ and the resulting generated out-of-line region is shown by the following example
 </tr>
 </tbody>
 </table>
+<note role="elaboration">
+<p>Take note that, in the above pair of examples, the timing attributes on the generated out-of-line region
+make use of an <code>end</code> attribute rather than the original <code>dur</code> attribute of the
+<loc href="#content-vocabulary-div"><el>div</el></loc> parent element of the inline <loc href="#layout-vocabulary-region"><el>region</el></loc>
+element. This translation from <code>dur</code> to <code>end</code> is required by the above procedure in order to specify the
+active time interval of the generated out-of-line region within the root temporal extent.</p>
```
nigelmegitt: I'd prefer to change "within the root temporal extent" to "in the document temporal coordinate space" as per the suggestion above for line 17183.

spec/ttml2.xml
skynavga: Incorrect. The result of the [associate region] procedure is a single region.

spec/ttml2.xml
skynavga: No. The phrase active begin/end times within the root temporal extent make it clear what syncbase is referenced; i.e., the syncbase of the root temporal extent.

spec/ttml2.xml
skynavga: As I suggested above, it would be better to avoid both root temporal extent and document temporal coordinate space, and, instead, write this as relative to the syncbase of the body element.

spec/ttml2.xml
nigelmegitt: This text is not scoped to the [associate region] procedure so that is not relevant. Even if you consider that it _is_ relevant, running [associate region] for different descendant elements of, say, `body`, can result in that `body` effectively being associated with different regions each time. 

Further, after the comma the two methods of explicit association are listed. So this sentence is about explicit association, and the word "explicit" should remain.

spec/ttml2.xml
nigelmegitt: It doesn't say that though. It says "active begin and end times of `B` within the root temporal extent", which means something different, or at least can be misread. If we're going to tweak this wording we should make it properly clear.

spec/ttml2.xml
nigelmegitt: That would be confusing because `body` is reparented to `region` in later processing. What we unambiguously mean is the document temporal coordinate space, which I consider to be a synonym for "relative to the document syncbase".

spec/ttml2.xml
skynavga: Active begin times of an element B within the root temporal extent/document temporal coordinate space/syncbase of body element all mean the same thing.

nigelmegitt: Okay that's your understanding, which is partly biased by knowing what the text is _meant_ to say. The question is what is clearest to a reader who hasn't spent so many hours poring over the details.

skynavga: In spec writing, one always assumes an expert reader, and fills in the gaps with notes only as required. My general policy is to avoid specification text that goes out its way to inform a non-expert reader. In other words, write only the absolute minimum (in spec text) that expresses the necessary information. I often use ISO/IEC 13818-1 (MPEG-2 Systems) as the ideal model for minimalist spec writing.

spec/ttml2.xml
skynavga: I'm afraid this has digressed into a study of how to best paraphrase the semantics that already apply to region as specified in 11.1.2:

>If begin and (or) end attributes are specified on an out-of-line region element, then they specify the beginning and (or) ending points of a time interval during which the region is eligible for activation and with respect to which animation child elements of the region are timed. If specified, these begin and end points are relative to the time interval of the nearest ancestor element associated with a time interval, irregardless of whether that interval is explicit or implied. The nearest ancestor element of an out-of-line region element that is associated with a time interval is the layout element.

Whatever language we use should be consistent with what begin/end means on an out-of-line region element.

spec/ttml2.xml
skynavga: I will update to make reference to the associate region procedure.

spec/ttml2.xml
nigelmegitt: Do you see any problems with my proposal at https://github.com/w3c/ttml2/pull/911#discussion_r203663883 ?

spec/ttml2.xml
nigelmegitt: Right, and that can be found by inspection.
* Out-of-line `region` elements are significant as children of `layout`.
* `layout` is a child of `head`
* `head` is a child of `tt`.

The document temporal coordinate space is the temporal coordinate space defined by the computed value of the `ttp:timeBase` attribute of the `tt` element. None of the elements in the hierarchy described above can modify that or set an alternative syncbase. Hence using "document temporal coordinate space" as the reference term is correct.

spec/ttml2.xml
skynavga: @nigelmegitt Yes, there is a problem. We do not use computed begin/end time anywhere else in the spec. On the other hand, we use active time interval 50 times in the spec, one of which is in the definition of [inline region], the subject of this section: "An inline region is assigned its parent element's time interval as its active time interval." However, I do agree the following comment below, that we can change "root temporal interval" to "document temporal coordinate space".

spec/ttml2.xml
skynavga: I can accept your original comment https://github.com/w3c/ttml2/pull/911#discussion_r203668328 above, and will update.

skynavga: ok, in which case it will effectively match the language found under `body`:

>If relative begin or end times are specified on the body element, then these times are resolved by reference to the beginning of the document temporal coordinate space.

note that if we decide in the future to allow `begin` on any of `tt`, `head`, or `layout`, then we will need to revisit this language

spec/ttml2.xml
skynavga: Upding to scope language to a specific ISD.

skynavga:
@nigelmegitt please re-review; I have addressed all comments

spec/ttml2.xml
```diff
@@ -17133,9 +17133,9 @@ inline regions. This procedure additionally binds <loc href="#terms-content-elem
 to the corresponding generated out-of-line regions.</p>
 <note role="elaboration">
 <p>A <loc href="#terms-content-element">content element</loc> can be associated with only a single
-<loc href="#terms-region">region</loc>, either by an attribute specification or via a child 
-<loc href="#layout-vocabulary-region"><el>region</el></loc> element. Consequently, if a <loc href="#terms-content-element">content element</loc>
-specifies a <loc href="#layout-attribute-region"><att>region</att></loc> attribute, then any 
+<loc href="#terms-region">region</loc> (within a given <loc href="#terms-intermediate-synchronic-document">intermediate synchronic document</loc>),
+either by an attribute specification or via a child <loc href="#layout-vocabulary-region"><el>region</el></loc> element. Consequently,
+if a <loc href="#terms-content-element">content element</loc> specifies a <loc href="#layout-attribute-region"><att>region</att></loc> attribute, then any
```
nigelmegitt: I'm not going to block this for this change, so we can proceed with review. I'd be interested to know if anyone else takes the "(within a given intermediate synchronic document)" qualifier to apply also to the text that follows - it is potentially unclear to me. Of course the attribute or child region element cannot be present in an ISD where the content element has been rehomed via the `body` element to be a descendant of a single `region` element, so I can deduce what it is supposed to mean, but it reads a bit oddly in my opinion.

spec/ttml2.xml
```diff
@@ -17185,7 +17185,7 @@ otherwise, let <emph>ID</emph> be the value of the <att>xml:id</att> attribute o
     attributes to values equivalent to the respective active begin and end times of <emph>B</emph> within the root temporal extent;</p></item>
```
nigelmegitt: Change to 17188 needs to be applied to 17185 also.

nigelmegitt:

PR #913, opened 2018-07-19
Update TTML1 reference (#884).

### LINKED ISSUE #884: Change TTML1 normative reference to TTML1 3ED.
The TTML1 3ED prose matches that of TTML2 for those capabilities they have in common. In contrast, there is divergence between TTML1 2ED and TTML2 for the same capabilities.

skynavga:
Closes #884.


skynavga:

PR #915, opened 2018-07-20
Switch to CR entities, update dates in status.

skynavga:


skynavga:
@nigelmegitt no; I will do that in another, issue driven branch

PR #920, opened 2018-07-21
Add @extent to isd:{sequence,isd} (#919).

### LINKED ISSUE #919: Add missing @extent attribute to isd:isd.
Both `ttp:displayAspectRatio` and `ttp:pixelAspectRatio` are defined for use on `isd:isd`; however, an `extent` attribute was neglected to be defined (but is already implemented and used in TTX/TTPE). In order to support the use of `rw` and `rh` units in ISDs, as well as complete the triangle for (SAR/DAR/PAR), we need to define an `extent` attribute to take two pixel based length expressions for use on `isd:isd`.

skynavga:
Closes #919.

### DIFF
--- a/spec/rnc/ttml2-isd.rnc
+++ b/spec/rnc/ttml2-isd.rnc
@@ -30,6 +30,7 @@ TTAF.isd.sequence =
   }
 
 TTAF.isd.sequence.attlist &=
+  attribute extent { string }?,
   attribute size { xsd:nonNegativeInteger }?,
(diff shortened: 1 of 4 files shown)


skynavga:
The last commit (https://github.com/w3c/ttml2/pull/920/commits/8fe2d1faa92a1775f5f89ce8a00761ca95b51c67) message should have said "inline definitions".

skynavga:

PR #923, opened 2018-07-22
Miscellaneous editorial improvements (#895).

### LINKED ISSUE #895: Miscellaneous editorial improvements.
+ section 3.2.3

   * add note indicating that the mandatory semantics of presentation
   processing include performing 11.3.1.3 Intermediate Document Construction
   and 11.3.1.4 Synchronic Flow Processing;

 + section 8.1.[3456]

   * remove "such as div, p, span and br" from notes

 + section 9.1.{1,5}, add note indicating that no text node descendant
   is significant for the purpose of performing [construct anonymous
   span]

 + section 10.4.2.1

   * change "is not associated with" to "does not specify" in note

 + section 10.4.3.2

   * add note (or text) indicating that computed value may be
   further resolved at either layout or rendering time when (1)
   such resolution is required by the semantics of a specific property
   or the information required to perform the resolution is not
   available until layout or rendering;

   * add termdef of "resolved computed value"; add termrefs
   at reference points (see tts:lineHeight and possibly elsewhere);

 + section 10.4.4.1

   * and extra line spacing between concept labels and definitions;

 + section 10.4.4.2

   * change "an element" to "a presentation related element" in the
   prologue clause;

   * step (3), change "for each nested style element" to "for each
   conditionally included nested style element" or add note to this
   effect;

   * step (5), remove "(set)" parenthetical;

   * step (5), change "for each immediate animation element" to "for
   each conditionally included immediate animation element" or add note
   to this effect;

 + section 10.4.4.3

   * change "an element" to "a presentation related element" in the
   prologue clause;

   * change "following sub-steps:" to "following sub-steps
   (if possible):"; add note indicating possible cases where it is
   not possible, and that further value resolution may occur;

 + section 10.4.4.4

   * change "each element" to "each presentation related element" in the
   prologue clause;

 + section 11.3.1.2

   * change "explicit inline" to "inline"; we no longer have implicit
   inline regions;

   * in [process inline regions], step (1), change "a region attribute"
   to "a region attribute that references a conditionally included
   region element";

   * in [process inline regions], step (2), change "a region element"
   to "a conditionally included region element";

   * add note under step (3)(a) that handles use of "dur"; change
   example to use end instead of dur;

 + section 11.3.1.3

   * add note at end of [construct intermedia document] that indicates
   an implementation may optionally perform style resolution as
   a post-processing step of this procedure (see note under 11.3.1.4
   step 2);

  + section 11.3.1.4

   * move paragraph "Each intermediate ... as follows:" at beginning
   of label [flow transformation] to immediately before the label
   [flow transformation], rewriting it as follows:

   "For the purpose of performing presentation processing, for each
   intermediate synchronic document $DOC_{inter_i}$ produced by
   11.3.1.3 Intermediate Synchronic Document Construction, invoke the
   following [flow transformation] procedure in order to map each
   $DOC_{inter_i}$ to a corresponding XSL FO document instance $F_i$."

   * incorporate paragraph "For each resulting document instance F, ..."
   at end of [flow transformation] into the text of the above rewritten
   paragraph;

   * [TTML2 2e] in step (6), handle "image in a block context"; in the
   mean time add note addressing this point in [TTML2 1e];

   * [TTML2 2e] in step (7), handle case of tts:displayAlign="justify"

   * [TTML2 2e] in step (8), subdivide into (a) and (b), copying
   current text into (a) and adding new text defining the mapping of
   inline images in (b); in the mean time, add note to this effect
   in [TTML1 1e]; N.B. also need to handle tts:display="inlineBlock"

 + section 11.3.1.5

   * incorporate both block level and inline level tt:image elements
   into examples;

   * in note, add that the intermediate documents shown do not use
   the ISD format (consider changing them to the ISD format!)

 + section 11.3.2.1

   * change "requires use" to "requires support or use";

 + acknowledgements

   * add Stefan Pöschel

skynavga:
Closes #895.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -5872,9 +5872,7 @@ attribute or inline style attributes or a combination thereof.</p>
 <note role="explanation">
 <p>Style properties that are associated with a <el>body</el> element
 in a <loc href="#terms-document-instance">document instance</loc> are available for style inheritance by
-descendant <loc href="#terms-content-element">content elements</loc> such as <el>div</el>, <el>p</el>,
-<el>span</el>
-and <el>br</el>.</p>
+descendant <loc href="#terms-content-element">content elements</loc>.</p>
 </note>
 <p>If no <att>timeContainer</att> attribute is specified on a
 <el>body</el> element, then it must be interpreted as having
@@ -5948,14 +5946,9 @@ element group apply semantically to the <el>div</el> element.</p>
 <p>An author may associate a set of style properties with a <el>div</el> element by means of either
 the <att>style</att> attribute or inline style attributes or a combination thereof.</p>
 <note role="explanation">
-<p>Style properties that are associated
-with
+<p>Style properties that are associated with
 a <el>div</el> element in a <loc href="#terms-document-instance">document instance</loc>
-are available for style inheritance by descendant <loc href="#terms-content-element">content elements</loc> such as
-<el>div</el>, 
-<el>p</el>,
-<el>span</el>, and
-<el>br</el>.</p>
+are available for style inheritance by descendant <loc href="#terms-content-element">content elements</loc>.</p>
 </note>
 <p>If no <att>timeContainer</att> attribute is specified on
 a <el>div</el> element, then it must be interpreted as having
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -16533,9 +16541,9 @@ merge <emph>P</emph> into the specified style
 set of <emph>E</emph>, <emph>SSS(E)</emph>;</p></item>
 <!-- 5. [animation styling] -->
 <item><p><phrase role="strong">[animation styling]</phrase> if the
-element type of <emph>E</emph> is not the animation element type
-<el>set</el>, then for each
-immediate animation (<el>set</el>) element child <emph>A</emph> of element <emph>E</emph>,
+element type of <emph>E</emph> is not an <loc href="#terms-inline-animation">inline animation</loc> element,
```
nigelmegitt: It is not clear where do out-of-line referenced animations get brought in. I've raised #927 regarding this point.

nigelmegitt:

PR #924, opened 2018-07-22
Address possible conflict in placement of ruby text containers (#882).

### LINKED ISSUE #882: Two text containers with same tts:rubyPosition (not "outside").
When a ruby annotation with two text containers is used, currently each text container could use the same `tts:rubyPosition` value different to `outside` (e.g. `after`). In that case the two containers would have to be rendered on the same side of the actual ruby base text.

Is this case intended? If so, which of the two text containers will be the one rendered adjacent to the base text?

skynavga:
Closes #882.


skynavga:
Prior to merge, wrap _resolved computed value_ with `<termref>...</termref>` from #923.

spoeschel:
The text looks good to me, too.

skynavga:

PR #928, opened 2018-07-23
Clarify that descendants of conditionally excluded elements are alway…

### LINKED ISSUE #926: Clarify that descendants of conditionally excluded elements cannot be included
See https://github.com/w3c/ttml2/pull/900/files/204ff46225830b23a2c6a4ca7f147cbdd68956d6..9c9a3efb6b2e938d098eff732dcdbca45f791943#r204399436

Need to clarify that the text in 8.2.1 condition (attribute):

> if an element specifies a `condition` attribute, and the its `<condition>` expression value evaluates to `false`, then the semantics of the element and its descendant elements must be ignored unless explicitly specified otherwise.

does _not_ mean that a descendant with a `condition` attribute which evaluates to `true` becomes conditionally included.

Also remove the qualifier "conditionally excluded" before "descendants" in the note introduced by #900 at the end of §8.2.1.

nigelmegitt:
…s excluded themselves (#926)

Closes #926.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -6285,18 +6285,6 @@ value evaluates to <code>false</code>:</p>
 <note role="clarification">
 <p>The exceptions noted above also apply to descendants of a
 <loc href="#terms-conditionally-excluded">conditionally excluded</loc> element.</p>
-<p>However conditional exclusion does not always follow references:
-for example, if a <loc href="#terms-conditionally-excluded">conditionally excluded</loc>
-<loc href="#styling-vocabulary-style"><el>style</el></loc> element
-<emph>S<sub>1</sub></emph>
-references a <loc href="#terms-conditionally-excluded">conditionally included</loc> style element
-<emph>S<sub>2</sub></emph>,
-then, since the <loc href="#semantics-style-association-chained-referential">chained referential semantics</loc>
-are excepted from the ignored semantics, 
-the style set of 
-<emph>S<sub>2</sub></emph> is not ignored, 
-whereas any style attributes specified on
-<emph>S<sub>1</sub></emph> are ignored.</p>
 </note>
 </div3>
 <div3 id="content-attribute-xlink-arcrole">
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -6262,7 +6262,7 @@ from evaluation cannot change or is known to not change during presentation proc
 <p>For example, if a condition expression depends on the value of a <loc href="#content-value-condition-function">condition function</loc>,
 then the condition expression's value may change during presentation processing, in which case eager evaluation cannot be used.</p>
 </note>
-<p>For the purpose of presentation processing, if an element specifies a <att>condition</att> attribute, and the
+<p>For the purpose of presentation processing, if an element specifies a <att>condition</att> attribute, and 
 its <loc href="#content-value-condition">&lt;condition&gt;</loc> expression value evaluates to <code>false</code>,
 or if an ancestor of the element has a <att>condition</att> whose 
 <loc href="#content-value-condition">&lt;condition&gt;</loc> expression value evaluates to <code>false</code>,
@@ -6291,6 +6291,18 @@ value evaluates to <code>false</code>:</p>
 <note role="clarification">
 <p>The exceptions noted above also apply to descendants of a
 <loc href="#terms-conditionally-excluded">conditionally excluded</loc> element.</p>
+<p>However conditional exclusion does not always follow references:
+for example, if a <loc href="#terms-conditionally-excluded">conditionally excluded</loc>
+<loc href="#styling-vocabulary-style"><el>style</el></loc> element
+<emph>S<sub>1</sub></emph>
+references a <loc href="#terms-conditionally-excluded">conditionally included</loc> style element
+<emph>S<sub>2</sub></emph>,
+then, since the <loc href="#semantics-style-association-chained-referential">chained referential semantics</loc>
+are excepted from the ignored semantics, 
+the style set of 
+<emph>S<sub>2</sub></emph> is not ignored, 
+whereas any style attributes specified on
+<emph>S<sub>1</sub></emph> are ignored.</p>
 </note>
 </div3>
 <div3 id="content-attribute-xlink-arcrole">
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -6267,9 +6267,12 @@ then the semantics of the element must be ignored unless those semantics are exp
 attribute that evaluates to <code>false</code>, then the content of that element and its descendants are ignored for presentation purposes.</p>
 </note>
 <note role="elaboration">
-<p>Even if an element specifies a <att>condition</att> attribute that evaluates to <code>true</code>, that
-element is nevertheless <loc href="#terms-conditionally-excluded">conditionally excluded</loc> if it has an ancestor element that that has a <att>condition</att>
-attribute that evaluates to <code>false</code>.</p>
+<p>Evaluation of the condition attribute is not required (and not relevant)
+on an element that is conditionally excluded
+due to one of its ancestor elements being conditionally excluded.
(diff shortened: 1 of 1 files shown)


nigelmegitt:
@skynavga Your comment in #926 is consistent with the edits I am proposing here. If you disagree with anything in the pull request then please comment on it, otherwise we are left with an unnecessary ambiguity where we both _agree_ on the intent of the specification.

nigelmegitt:
By the way, if you want to make the case that this is editorial, please go ahead.

skynavga:
@nigelmegitt let's not turn this into a pissing contest, ok; if you want to resolve this today, then please process the new PR; otherwise, you are going to blow out our schedule

nigelmegitt:
I raised the issue, it has several points in it, which #929 does not address. #928 does, and I raised it early in the hope of getting to a quicker resolution. I'm happy to work constructively here, but I can't do that on my own. Since you've actually made a review comment about the extended example, I'll remove that text for now, since it is editorial and not strictly necessary.

The only other change that I would consider on #928 is that adding both "or if an ancestor of the element has a condition whose `<condition>` expression value evaluates to false"  _and_ following that with "then the semantics of the element and its descendant elements must be ignored" could be considered tautological - the "and its descendant elements" would be removed to address that point.

spec/ttml2.xml
```diff
@@ -6237,7 +6237,7 @@ in the core vocabulary catalog:</p>
 </ulist>
 <div3 id="content-attribute-condition">
 <head>condition</head>
-<p>The <att>condition</att> attribute is used to conditionally exclude or include an element from semantic processing,
+<p>The <att>condition</att> attribute is used to conditionally exclude an element and its descendants from semantic processing,
```
skynavga: Adding "and its descendants" is not necessary. This follows from the meaning of conditional exclusion of an element, so it is mere repetition of the particulars.

Also, need to link "conditionally exclude" to term definition in terminology.

spec/ttml2.xml
```diff
@@ -6256,13 +6256,20 @@ from evaluation cannot change or is known to not change during presentation proc
 <p>For example, if a condition expression depends on the value of a <loc href="#content-value-condition-function">condition function</loc>,
 then the condition expression's value may change during presentation processing, in which case eager evaluation cannot be used.</p>
 </note>
-<p>For the purpose of presentation processing, if an element specifies a <att>condition</att> attribute, and the
+<p>For the purpose of presentation processing, if an element specifies a <att>condition</att> attribute, and 
 its <loc href="#content-value-condition">&lt;condition&gt;</loc> expression value evaluates to <code>false</code>,
-then the semantics of the element and its descendant elements must be ignored unless explicitly specified otherwise.</p>
+or if an ancestor of the element has a <att>condition</att> whose
```
skynavga: This is a redundant, wordier, and possibly incorrect rewrite of existing text, and inverts the application of exclusion from a conditionally excluded element to its descendants to an expression of inheritance from an conditionally excluded ancestor. In short, this adds nothing, muddies the language, and is much wordier in paraphrasing the original text.

spec/ttml2.xml
```diff
@@ -6256,13 +6256,20 @@ from evaluation cannot change or is known to not change during presentation proc
...
 its <loc href="#content-value-condition">&lt;condition&gt;</loc> expression value evaluates to <code>false</code>,
-then the semantics of the element and its descendant elements must be ignored unless explicitly specified otherwise.</p>
+or if an ancestor of the element has a <att>condition</att> whose 
+<loc href="#content-value-condition">&lt;condition&gt;</loc> expression value evaluates to <code>false</code>,
+then the semantics of the element and its descendant elements must be ignored unless those semantics are explicitly excepted.</p>
 <note role="elaboration">
 <p>For example, if a <loc href="#content-vocabulary-p"><el>p</el></loc> element specifies a <att>condition</att>
 attribute that evaluates to <code>false</code>, then the content of that element and its descendants are ignored for presentation purposes.</p>
 </note>
+<note role="elaboration">
```
skynavga: This note is unnecessary since it assumes incorrectly that conditional inclusion is defined as a 1s order predicate, when it is not.

spec/ttml2.xml
nigelmegitt: OK will change to add the link. It is not obvious that our definition of conditional exclusion applies to the descendants, since they _are_ still added to the tree even if excluded. Can we compromise by putting "and its descendants" in parenthesis?

spec/ttml2.xml
nigelmegitt: Unless we explicitly state a rule that applies conditional exclusion to descendants, or that excludes semantics in all cases if an ancestor is conditionally excluded, then there's no text from which to infer this normative rule. I don't think it's enough to assume it will be obvious that conditional exclusion applies hierarchically to a tree from the excluded element down to its descendants. This is our construct in TTML2, and we need to be clear about it.

spec/ttml2.xml
skynavga: ok

spec/ttml2.xml
nigelmegitt: Where is this (not defined as a 1st order predicate) stated? The note is intended to clarify for readers who have not captured this nuance the effect of the normative text. Is the note _wrong_?

spec/ttml2.xml
skynavga: Take the existing text, "For the purpose of presentation processing, if an element specifies a condition attribute, and the its <condition> expression value evaluates to false, then the semantics of the element and its descendant elements must be ignored unless explicitly specified otherwise." then add the following before the final period ", which is to say, the element and its descendant elements are <loc href="#terms-conditionally-excluded">conditionally excluded</loc>".

nigelmegitt: By the way, in terms of "possibly incorrect", it's trivial to show that all descendants of a conditionally excluded element have an ancestor that specifies a condition expression that evaluates to false, so I'm certain it is correct in the semantic.

spec/ttml2.xml
skynavga: [conditionally excluded (included)] ... If an element is not conditionally excluded, then it is said to be conditionally included.

spec/ttml2.xml
nigelmegitt: @skynavga your proposal at https://github.com/w3c/ttml2/pull/928#discussion_r204814592 does not cover the case of a descendant element of a conditionally excluded element having a condition that evaluates to true. I could modify it further to append "regardless of the evaluation of any `<condition>` expressions specified on those descendant elements". However at this stage the text is even wordier than my proposal. They would be equivalent though, by my reading.

spec/ttml2.xml
nigelmegitt: done

spec/ttml2.xml
nigelmegitt: We should put the normative requirements in the main section not the terminology section though. That definitional text "said to be" seems to be saying something about descriptive language rather than stating a normative rule.

skynavga: @nigelmegitt re: https://github.com/w3c/ttml2/pull/928#discussion_r204820496, terminology is normative (as definitions), which play a semantic role elsewhere; "is said to be" is equivalent to "is defined to mean as follows"; see also my proposed rewrite of this note, and make sure to wrap "conditionally included" and "conditionally excluded" with term links

nigelmegitt: done

nigelmegitt: and terms wrapped with appropriate links and styling

skynavga:
Changing this to editorial as the final changes are not substantive in nature.

skynavga:

tairt:
> @nigelmegitt let's not turn this into a pissing contest, ok; if you want to resolve this today, then please process the new PR; otherwise, you are going to blow out our schedule

@skynavga I would like to remembe you and us all that W3C has [Code of Conduct.](https://www.w3.org/Consortium/cepc/) .

PR #929, opened 2018-07-24
Clarify condition as pertains to exclusion vs inclusion (#926).

### LINKED ISSUE #926: Clarify that descendants of conditionally excluded elements cannot be included

skynavga:
Closes #926.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -6237,7 +6237,7 @@ in the core vocabulary catalog:</p>
 </ulist>
 <div3 id="content-attribute-condition">
 <head>condition</head>
-<p>The <att>condition</att> attribute is used to conditionally exclude or include an element from semantic processing,
+<p>The <att>condition</att> attribute is used to <loc href="#terms-conditionally-excluded">conditionally exclude</loc> an element from semantic processing,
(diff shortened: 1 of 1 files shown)


skynavga:
@nigelmegitt if you want to handle it this way, then we are either not going to make our schedule, or this issue isn't going to be addressed in TTML2 1e; which would you prefer?

nigelmegitt:
There's no reason we cannot handle this by amendments to #928 and meet our schedule. I've already made a couple, and suggested one more that could be made, if it would help.

skynavga:
See #928 for subsequent processing.

PR #930, opened 2018-07-24
Map out-of-line animation to inline animation to handle styling seman…

### LINKED ISSUE #927: Out-of-line animations are not included in Specified Style Set Processing
Whereas §10.4.4.2 Specified Style Set Processing includes inline animation elements that are conditionally included child elements of E, there doesn't seem to be any step in the process that merges in the out of line animations referenced by the `animate` attribute.

skynavga:
…tics (#927).

Closes #927.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -18355,9 +18355,7 @@ elements in the <loc href="#element-vocab-group-metadata"><code>Metadata.class</
 </tbody>
 </table>
 <p>An <loc href="#terms-out-of-line-animation">out-of-line</loc> <el>animate</el> element
-must specify an <loc href="#content-attribute-xml-id"><att>xml:id</att></loc> attribute,
-but must not specify any attribute from the <loc href="#attribute-vocab-group-timing">Timing Attributes</loc> Module, and, if specified,
-must be ignored for the purpose of performing the <loc href="#procedure-resolve-timing">[resolve timing]</loc> procedure.</p>
+must specify an <loc href="#content-attribute-xml-id"><att>xml:id</att></loc> attribute.</p>
 <p>Style property attributes targeted by an <el>animate</el> element are specified directly using
 attributes in the TT Style Namespaces or in a namespace that is not a TT Namespace, where the list (sequence) of animation values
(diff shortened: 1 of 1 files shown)


skynavga:
@nigelmegitt re: https://github.com/w3c/ttml2/pull/930#pullrequestreview-139995754 I have given some thought to timing on out-of-line (OOL) animations, and I have not come up with a rational model for how this might work; nominally, any timing on an out-of-line animation would be in relation to its hierarchy, e.g., animate/animation/head/tt, which would make that timing be expressed in the document temporal coordinate space; then the question is how to synchronize this with timing on possibly multiple referring elements; my conclusion was that it is better to avoid this problem, at least in TTML2 1e, by not permitting timing on OOL animations;

skynavga:
in other words, if you want timing on an animation that does not correspond with the referring elements timebase, then don't use out-of-line animation

nigelmegitt:
If the semantic seems unclear and we want to leave it to implementations, then we should simply say nothing about it. The newly proposed text that creates a copy of the `animate|set` element as an inline child gives a good opportunity to simply copy across the timing attribute values without prior evaluation. I would simply say nothing about this problem and allow timing on OOL animations, then we have parity with TTML1 for `set` at least, in the sense that TTML2 isn't saying what TTML1 didn't say.

skynavga:
@nigelmegitt ok, I can live with that for TTML 1e; will update in a minute

skynavga:
@nigelmegitt updated

spec/ttml2.xml
```diff
@@ -18630,6 +18630,8 @@ elements in the <loc href="#element-vocab-group-metadata"><code>Metadata.class</
 </tr>
 </tbody>
 </table>
+<p>An <loc href="#terms-out-of-line-animation">out-of-line</loc> <el>set</el> element
```
nigelmegitt: There used to be a line like this for OOL animate elements too - I think it got dropped by accident?

skynavga: Nope, it's still there. Just not modified (now).

skynavga: OTOH, `set` was missing this entirely.

nigelmegitt: Ah, right, I see, thanks.

skynavga:

PR #933, opened 2018-07-25
Update changes for CR2 to CR3 diffs (#931).

### LINKED ISSUE #931: Update changes file to add CR2 to CR3 differences.

skynavga:
Closes #931.

### DIFF
--- a/spec/ttml2-changes.html
+++ b/spec/ttml2-changes.html
@@ -31,31 +31,46 @@ div.exampleInner { background-color: #d5dee3;
       </a>
       <ul class="toc">
         <li class="tocline">
-          <a class="tocxref" href="#change-history-ttml2-cr1-to-ttml2-cr2"><span class="secno">1.1</span>TTML2 CR1 to CR2</a>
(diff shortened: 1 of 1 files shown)


spec/ttml2-changes.html
```diff
@@ -72,9 +87,166 @@ <h2><a id="change-history"></a>1 Change History (Non-Normative)</h2>
...
+  the value space consists of whole integers only.
+
+* In 10.2.55, do not support animation of tts:pitch.
+
+* In 10.2.56, do not support animation of tts:speak.
+
+* In 10.4.4.2, step (3), qualify nested style element as nested
+  conditionally included style element.
+
+* In 10.4.4.2, step (5), qualify animation element as inline animation
```
nigelmegitt: I can't see mention of the enhancements to describe the processing of out-of-line animation here - I think it's the last change we made.

skynavga: It's there, just not in the right place:

* In 13.2.1, specify mapping of out-of-line animation to synthetic
  (anonymous) inline animation in order to handle style application
  via out-of-line animation.

I will move to Technical Changes where it should be placed.

skynavga: @nigelmegitt fixed

PR #940, opened 2018-07-29
Fix <pitch> syntax (#939).

### LINKED ISSUE #939: Fix <pitch> syntax.
The definition of `<pitch>` syntax should not specify an optional sign, since the definition of `<percentage>` and `<number>` already include an optional sign.

I am marking this as editorial since it is obvious that the current definition does not intend there to be two sets of signs and therefore, this is an editorial fix and not a change in the intended value syntax.

skynavga:
Closes #939.


skynavga:

PR #942, opened 2018-07-30
Ensure audio styles are defined on style element (#941).

### LINKED ISSUE #941: Add audio styles to style element in schemas.
This is an editorial change to synchronize schemas with spec.

skynavga:
Closes #941.


skynavga:

PR #944, opened 2018-07-30
Improve non-negative-real interoperability (#943).

### LINKED ISSUE #943: Improve interoperability of non-negative-real and xs:decimal.
As presently defined in §10.3.27, `non-negative-real` includes `.0` and `0.0`, but does not include `0.` This creates an interoperability problem with the use of `xs:decimal` [1] in the schema for `tta:gain` and `tta:pan`. There is no reason we should exclude `0.`, so I propose we change the definition of `non-negative-real` to the following:

```
non-negative-real
  : <digit>* "." <digit>+
  | <digit>+ "."
```

[1] https://www.w3.org/TR/xmlschema-2/#decimal

skynavga:
Closes #943.


nigelmegitt:
Marking for the agenda, since as per https://github.com/w3c/ttml2/issues/943#issuecomment-409219624 the deadline for a change like this seems to have passed, even though it is (probably) a good idea. The alternative would be to defer to 2nd Ed.

skynavga:
Deferring to consideration under ttml.next.

palemieux:
The pull request is missing tests to exercise the change.

PR #953, opened 2018-07-31
The style attribute does not apply to animate or set (#952).

### LINKED ISSUE #952: The style attribute is not defined for animate/set.
When we removed `@style` from `animate` and `set` in #703, we neglected to remove these elements from the enumeration in §10.2.1.

skynavga:
Closes #952.


skynavga:

PR #956, opened 2018-07-31
Widen data type for tta:{gain,pan} (#955).

### LINKED ISSUE #955: Widen schema type for tta:{gain,pan}.
Unless and until we adopt #943, we need to widen the schema data type for values of `tta:{gain,pan}` to use `xs:string` instead of `xs:decimal`. Otherwise, if this schema is used to generate code for validation, then it will not be possible to detect `0.` as an invalid value.

skynavga:
Closes #955.

### DIFF
--- a/spec/rnc/ttml2-datatypes.rnc
+++ b/spec/rnc/ttml2-datatypes.rnc
@@ -170,7 +170,7 @@ TTAF.FrameRateMultiplier.datatype =
   xsd:string { pattern = "\p{Nd}+\s+\p{Nd}+" }
 
 TTAF.Gain.datatype =
-  xsd:decimal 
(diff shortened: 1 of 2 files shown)


skynavga:

skynavga:
@nigelmegitt re: your last paragraph of https://github.com/w3c/ttml2/pull/956#pullrequestreview-143131465, I have just checked the spec and schemas, and all continuously animatable styles except for `tts:opacity` are currently defined with a type of `xs:string`; therefore, to avoid false negatives during schema evaluation when using `tts:opacity` on `tt:animate`, we should change the type used with `tts:opacity` to `xs:string` in the schemas.

PR #960, opened 2018-08-01
Change resolved computed value to used value (#959).

### LINKED ISSUE #959: Change "resolved computed value" to "used value".
To improve consistency with CSS, change "resolved computed value" to read "used value", a term used by CSS that is functionally equivalent to our use of the term "resolved computed value".

Note that XSL-FO does not use the term "used value"; rather, it processes computed values to obtain **traits** on area objects, where such **traits** are very close if not equivalent to the CSS notion of a _used value_.

skynavga:
Closes #959.

When merging, change **resolved computed value** to **used value** at https://github.com/w3c/ttml2/pull/958/files#diff-5ac39ab39265717d36dfd789d254ce1aR11144.


skynavga:

PR #962, opened 2018-08-01
Remove xml:base and @condition from chunk (#961).

### LINKED ISSUE #961: Remove xml:base and @condition from chunk.
An `xml:base` attribute was inadvertently defined for use on the `chunk` element. However, the `chunk` element does not permit any child elements or metadata attributes to which an `xml:base` attribute could potentially apply. This is currently reflected in the schemas as well which do not define use of `xml:base`.

skynavga:
Closes #961.

### DIFF
--- a/spec/rnc/ttml2-embedded.rnc
+++ b/spec/rnc/ttml2-embedded.rnc
@@ -58,9 +58,7 @@ TTAF.chunk =
   }
 
 TTAF.chunk.attlist &=
-  TTAF.base.attrib,
   TTAF.id.attrib,
-  TTAF.condition.attrib,
   TTAF.encoding.attrib,
   TTAF.length.attrib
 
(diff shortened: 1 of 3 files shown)


skynavga:
Defer to ttml.next.

skynavga:
Reopened this PR to remove `@condition` and `xml:base` from `chunk` element.

nigelmegitt:
I think this pull request is somewhat premature - we haven't really discussed the issue yet or what we would like to do about it. Happy to add it to the agenda for Thursday.

nigelmegitt:
Discussed in today's call. Agreement in principle to proceed, with a request for accompanying tests in a PR to w3c/ttml-tests.

nigelmegitt:
Tests at w3c/ttml2-tests#210 look good.

PR #964, opened 2018-08-01
Fix animation element's content model in schemas (#963).

### LINKED ISSUE #963: Fix animation element content type in schemas.
The animation element's content type in the schemas does not match the element information item definition in the spec, and needs to be synchronized with the latter.

skynavga:
Closes #963.


skynavga:

PR #968, opened 2018-08-02
Clarify inheritance of font size by ruby text containers and ruby tex…

### LINKED ISSUE #967: Special inheritance rule of font size for ruby text content is contradictory.
Given the following example:

```
        <span tts:ruby="container" tts:fontSize="0.1rv">
         <span tts:ruby="baseContainer"><span tts:ruby="base">base</span></span>
          <span tts:ruby="textContainer"><span tts:ruby="text">text</span></span>
        </span>
```

The computed font size at _text_ will be 25% of `0.1rv` since Section 10.2.21.1 specifies that the computed font size for the text container is 50% of that the container, and the computed font size for the text is 50% of that of the text container.

Given the following example:

```
        <span tts:ruby="container" tts:fontSize="0.1rv">
         <span tts:ruby="base">base</span>
          <span tts:ruby="text">text</span>
        </span>
```

The computed font size at _text_ will be 50% of `0.1rv` since Section 10.2.21.1 specifies that the computed font size for the text is 50% of that of the text container.

Is that intended, i.e. that the font size of _text_ is different?

skynavga:
…t content (#967).

Closes #967.


spec/ttml2.xml
```diff
@@ -10504,27 +10504,21 @@ for which the computed value of <loc href="#style-attribute-ruby"><att>tts:ruby<
...
-Instead, if no <loc href="#style-attribute-fontSize"><att>tts:fontSize</att></loc> attribute is specified on the
-ruby text <loc href="#content-vocabulary-span"><el>span</el></loc> element, a value equal to 50% of the value of the computed value
-of <loc href="#style-attribute-fontSize"><att>tts:fontSize</att></loc> on the nearest ancestor
-<loc href="#content-vocabulary-span"><el>span</el></loc> element for which the computed value
-of <loc href="#style-attribute-ruby"><att>tts:ruby</att></loc> is <code>textContainer</code> or <code>container</code> is used,
-whichever is most immediate to the ruby text <loc href="#content-vocabulary-span"><el>span</el></loc> element;
+Instead, if it is an <termref def="defs-explicit-ruby-text-container">explicit ruby text container</termref> and does not specify
+a <loc href="#style-attribute-fontSize"><att>tts:fontSize</att></loc> attribute, or if it is an
+<termref def="defs-implied-ruby-text-container">implied ruby text container</termref>, then
+a value equal to 50% of the value of the computed value of <loc href="#style-attribute-fontSize"><att>tts:fontSize</att></loc>
```
nigelmegitt: I'm having trouble decoding this:

* can we change "50% of the value of the computed value" to "50% of the computed value"?
* why is it setting the _used_ value to 50% of the _computed_ value? That feels like a dimensional change - shouldn't we be setting a computed value based on a computed value or a used value based on a used value? Possibly this is not the intended reading - I got there by seeing the wording "is used" at the end of the sentence. A fix could be to reword to: "the computed value of the fontSize of a ruby text container that does not explicitly specify a fontSize shall be set to 50% of the fontSize of the nearest ancestor ruby container"

spec/ttml2.xml
```diff
@@ -12355,17 +12349,25 @@ further information.</p>
...
+namely, <termref def="defs-ruby-base-content">ruby base content</termref> or <termref def="defs-ruby-text-content">ruby text content</termref> that is specified without a <termref def="defs-ruby-base-container">ruby base container</termref> or <termref def="defs-ruby-text-container">ruby text container</termref>, respectively,
+is processed in such a manner as to imply the presence of an anonymous <termref def="defs-ruby-base-container">ruby base container</termref> and (or) anonymous <termref def="defs-ruby-text-container">ruby text container</termref> as required.
 An <termdef id="defs-implied-ruby-base-container" term=""><term>implied ruby base container</term></termdef> and (or)
 <termdef id="defs-implied-ruby-text-container" term=""><term>implied ruby text container</term></termdef> is required in order to generate and apply
 implied or collective styling semantics to an inline block area which contains the inline areas generated by the
 <termref def="defs-isolated-ruby-content">isolated ruby content</termref>.</p>
+<note role="clarification">
```
nigelmegitt: Term definitions should not be in non-normative notes, since they are referenced by normative text and that confuses the normative/non-normative status. Please could you remove the `<note>` wrapper @skynavga ?

spec/ttml2.xml
nigelmegitt: Comment above updated in place to propose "shall be set to 50% of the computed value of the fontSize" - I had accidentally omitted the "computed value of the" initially.

skynavga:
@palemieux re: https://github.com/w3c/ttml2/pull/968#pullrequestreview-143137480, that won't work since an explicit font size can be specified on a ruby text container which is intended to be the size of the font for all of its ruby text content children; also, this is how CSS works (if you specify font size on `rtc`; so we need to keep the language expressed here

palemieux:
> that won't work since an explicit font size can be specified on a ruby text container

Yes, and 50% of the parent's font size will be 50% of the ruby text container's font size.

spec/ttml2.xml
skynavga: @nigelmegitt just addressed these comments; I did not mean "used value" by "is used"; I have fixed by changing "is used" to "applies" which is common language we use elsewhere;

spec/ttml2.xml
skynavga: @nigelmegitt fixed

skynavga:
@palemieux please re-check keeping in mind that it is semantically inconsistent to define the meaning of font size on a ruby text container to be the base font size, i.e., to be 100% of the font size inherited from ruby container; the meaning of font sizes on the various ruby categories are:

ruby container (ruby) - ruby base (rb) font size
ruby base container (rbc) - ruby base (rb) font size
ruby base content (rb) - ruby base (rb) font size
ruby text container (rtc) - ruby text (rt) font size
ruby text content (rt) - ruby text (rt) font size

nigelmegitt:
Thanks, those edits address my comments; I'm not professing to be sure that this pull request addresses the issue though, and would like to see the debate between @skynavga and @palemieux resolved with mutual agreement if possible before going ahead with this. The issue seems to be agreed at least, it's just the solution...

Issue #978, opened 2018-08-23
Clarify that text combination is not defined in a ruby base or text context.


skynavga:

skynavga:
no; text contain usage is not defined anywhere in a ruby container, which is the equivalent to saying it is not defined in a ruby base or ruby text context;

also, I am not saying it is not permitted, I am saying it is not defined; it may be we wish to define such semantics in the future, but we have not done so yet, which is the point of adding such an editorial clarification

palemieux:
> no; text contain usage is not defined anywhere in a ruby container

You mean _text combination usage_, right?

palemieux:
>  it may be we wish to define such semantics in the future, but we have not done so yet, which is the point of adding such an editorial clarification

This is clearly more than an editorial clarification, and needs to be clarified in 2ED.

In the meantime, the TTWG can reach consensus on a resolution and the test vectors adapted accordingly.

palemieux:
To be sure, I do not object to either prohibiting or leaving undefined the presence of text combination within `<span tts:ruby="container"></span>`.

skynavga:
Yes, text combination. I agree that defining semantics for using text combination in a ruby container would have to be 2Ed or later. The intention of the present issue is to simply add a note saying that text combination is not defined in this context (in the 1Ed).

palemieux:
> The intention of the present issue is to simply add a note saying that text combination is not defined in this context (in the 1Ed).

Although I like the direction of the note, I do not think the current prose supports this editorial note. I would simply capture a consensus resolution on this thread, and defer its implementation until 2ED.

skynavga:
In what way does the current prose define the use of text combination in a ruby container? I contend that it says nothing, and, therefore, is not defined. For example, the following language:

>Further, let N<sub>R</sub> and N<sub>B</sub>, be, respectively, the number of glyph area descendants of these inline areas.

would count all of the glyph areas in a text combination inside a base context as base glyphs instead of counting the outer inline area (in cross-baseline progression orientation) as one base glyph, which would produce incorrect behavior with `withBase` alignment.

palemieux:
> In what way does the current prose define the use of text combination in a ruby container? 

`tts:textCombination` defines semantics.

`tts:ruby` defines semantics.

Neither prohibits their combined use, therefore both semantics apply -- for better or for worse.

> which would produce incorrect behavior with withBase alignment.

... which is why it needs to be corrected in 2ED.

palemieux:
[...]

PR #979, opened 2018-08-23
Clarify text combination semantics in a ruby container context (#978).

skynavga:


skynavga:
It does not define it because no specification text mentions this possibility even though it is syntactically possible. Therefore, there is no defined behavior. It falls into the same category of early implementers of TTML1 specifying `tts:origin` and `tts:extent` on a `div` or `p`, and inferring that such use was meant to create an anonymous region (or alter an associate, specified region).

skynavga:
@palemieux I disagree; the introduction of a note at this time is appropriate; what is your argument against doing so?

palemieux:
Read he thread in the issue

skynavga:
@palemieux are you disputing the facts stated in the comment? if so, then what is your rationale?

skynavga:
Closing without merging, but marking issue for ttml.next.

Issue #277, opened 2017-03-13
Incorporate CSS advances into TTML vertical text handling


r12a:
10.2.52 tts:writingMode
http://w3c.github.io/ttml2/spec/ttml2.html#style-attribute-writingMode

10.2.46 tts:textOrientation
http://w3c.github.io/ttml2/spec/ttml2.html#style-attribute-textOrientation

I think it would be useful to make the text in this section more closely resemble the approach in CSS, recent developments for which, i believe, have fixed some issues related to vertical text handling which apply to the model described for TTML.

CSS has the following values for vertical writing modes:
vertical-rl (equivalent to tbrl)
vertical-lr (equivant to tblr)
sideways-rl
sideways-lr

The latter two values are intended for use with 'horizontal scripts' when one wants to set them vertically (usually alongside something). They are useful additions because they effectively rotate the progression of lines in one direction or the other, which is what you typically need for such use cases. This affects where the start and end points lie, and how text will wrap to the next line, as well as the default orientation of the characters. For example, for sideways-lr, start is at the bottom of the vertical line, Latin characters would lie on their left side, and lines would wrap when they reach the top of the vertical line. These are things that don't happen with vertical-lr.

I believe there are some technical advantages to this approach, but also it will help authors if the models for TTML and CSS are similar.

Perhaps there will be some resistance because vertical text values such as sidewaysLeft and sidewaysRight existed previously in TTML, i'm not sure. If so, and therefore things can't be removed, then i suggest that sideways-rl and sideway-lr values be added, and the user be advised to use those rather than the sidewaysLeft and sidewaysRight values of tts:textOrientation.

(The TTML model also doesn't adopt all of the values used in XSL, and so the -rl and -lr parts of tb-rl and tb-lr don't have equivalents in the vertical text values. In CSS the inline direction of the block is not indicated by an explicit -rl/-lr on the container. That is set by the direction property. I'm guessing that you don't want to go there.)

nigelmegitt:
[scheduled for initial discussion 2017-03-23]

r12a:
[...]

PR #981, opened 2018-08-26
Remove #textOrientation-sideways-LR and related semantics (#980).

### LINKED ISSUE #980: Remove #textOrientation-sideways-LR
The semantics of `sidewaysLeft`, specifically, a 90 degree counter-clockwise rotation, is not well defined in the context of vertical, top-to-bottom inline progression line layout. For example, rotating latin text 90 degrees counter-clockwise would effectively make it a bottom-to-top inline progression layout, which is not supported by any (currently) defined writing mode in TTML. Reviewing both [UTR50](http://www.unicode.org/reports/tr50/tr50-17.html#vo_values) and [CSS Writing Modes Level 3](https://www.w3.org/TR/css-writing-modes-3/#text-orientation) reveals that in all cases, only 90 degree clockwise rotation is sanctioned.

Since we have marked `#textOrientation-sideways-LR` as at-risk, I suggest we remove this feature, including the semantics of `sidewaysLeft` which appears to sanction a less than well-defined CCW rotation.

Although this is a substantive change, it is permitted prior to PR without triggering a new CR since it has been marked at risk.

skynavga:
Closes #980.


r12a:
I believe this is in line with the recommendations in https://github.com/w3c/ttml2/issues/277.  On that basis, i agree with it.

spec/ttml2.xml
```diff
@@ -13417,22 +13414,16 @@ property has no effect.</p>
 they are rotated or not rotated in vertical writing modes, where this determination is made in accordance with the <code>Vertical_Orientation</code> 
 of each content character as determined by <bibref ref="utr50"/>.</p>
 <p>If the value of this attribute is <code>mixed</code>, then rotated <loc href="#terms-glyph-area">glyph areas</loc> are set sideways
-and non-rotated <loc href="#terms-glyph-area">glyph areas</loc> are set upright, where <emph>sideways</emph> means rotated either 90&deg; clockwise 
-or 90&deg; count-clockwise, respectively, in accordance with whether the applicable block progression direction is right-to-left or left-to-right, 
+and non-rotated <loc href="#terms-glyph-area">glyph areas</loc> are set upright, where <emph>sideways</emph> means rotated 90&deg; clockwise,
```
nigelmegitt: Note for a potential editorial tweak at some point in the future: we define _sideways_ in relation to `mixed` independently of the keyword `sideways` even though both have exactly the same definition, rotate 90º clockwise, so potentially we could make them both reference the same definition and define the behaviour in one place rather than two. Very minor point!

PR #1002, opened 2018-09-13
Update #audio-{description,speech} per WG resolution (#990).

### LINKED ISSUE #990: Audio related feature changes.
Having reviewed the BBC's ability to provide presentation implementations for the audio features, I would like to propose the following changes to at-risk features:

* remove the [`#embedded-audio`](https://www.w3.org/TR/ttml2/#feature-embedded-audio) feature designator, because it requires playback of audio resources embedded in the TTML2 file which BBC does not expect to be able to demonstrate. I would not like to remove the syntactical possibility of embedded audio resources directly, but arguably we may need to do that also. If I can find a way for BBC to implement this in the next day or so then I will propose "un-removing" `#embedded-audio` - I'm merely taking a conservative line here given the limited amount of remaining time.
* change `#embedded-audio` to `#audio` in [`#audio-description`](https://www.w3.org/TR/ttml2/#feature-audio-description) to match the previous change.
* Remove `#embedded-audio`, `#gain` and `#pan` from [`#audio-speech`](https://www.w3.org/TR/ttml2/#feature-audio-speech) because I do not think we can implement them all together straightforwardly, at least not so they are all applied at once to the same text using a combination of Web Audio API and Web Speech API, as they are in their current states; other implementation techniques could be made to work, such as making external asynchronous calls to a suitable text to speech service and playing back the resulting audio resource, but that requires a further degree of complexity.
* Remove `#speech` because it is not clear what it means for an implementation to "support" a speech synthesis processor, for example must the implementation include one or merely access one remotely? If the implementation can successfully implement `#speak` and `#pitch` then it has adequate support; further support is not required, in my view.

skynavga:
Closes #990.

This is a substantive change, but is covered under the at-risk process to occur without triggering a new CR. These changes are being made to permit audio description and audio speech functional constellations to be designated as identifiable features (1) without resorting to the functionality of embedded (versus externally referenced) audio and (2) without requiring the admixture of speech and non-speech audio functionality.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -19986,13 +19986,13 @@ same vocabulary enumerated above.</p>
 <p>A TTML processor supports the <code>#audio-description</code> feature if it
 supports the following features:</p>
 <ulist>
-<item><p><loc href="#feature-embedded-audio"><code>#embedded-audio</code></loc></p></item>
+<item><p><loc href="#feature-audio"><code>#audio</code></loc></p></item>
 <item><p><loc href="#feature-gain"><code>#gain</code></loc></p></item>
 <item><p><loc href="#feature-pan"><code>#pan</code></loc></p></item>
 </ulist>
 <p>It is an error for a content profile to require or permit use of the <code>#audio-description</code> feature but prohibit use of any of the following
 features:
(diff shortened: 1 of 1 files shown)


skynavga:
Merge early per WG resolution for PR processing.

PR #1005, opened 2018-09-23
Remove 'available' keyword value for <measure> expression (#1004).

### LINKED ISSUE #1004: Remove 'available' keyword from <measure>.
The `available` keyword was drafted in early 2015 based on the `fill-available` keyword defined in a 2012 WD  of CSS Sizing Level 3. It has since been moved out to CSS Sizing Level 4 (as `stretch-fit`) where its details are not yet specified.

Since use of this keyword falls under `#extent-measure`, `#ipd`, and `#bpd`, all of which are marked at risk, I propose we remove it now rather than deprecating/obsoleting it in 2e.

In any case, I think the current definition we have is not sufficiently well-defined to permit creating test content at this juncture, so it would have to remain untested if we leave it in 1e.

skynavga:
Closes #1004.


skynavga:
@nigelmegitt for the record, note that `<measure>` is used by three features, `#ipd`, `#bpd`, and `#extent-measure` as well, and all of these are marked as at risk in CR3, so I think we are covered to remove it

nigelmegitt:
@skynavga re https://github.com/w3c/ttml2/pull/1005#issuecomment-424023589 thank you for listing those.

skynavga:
Merged early for PR processing. Do not expect any further comments on this pull request/issue.

PR #1048, opened 2019-03-16
Change entities to 2e ED, and update TTML1 bibref.

skynavga:


spec/entitiesedcopy.dtd
```diff
@@ -15,16 +15,16 @@
...
 
 <!ENTITY w3c.tr  "">	
 
 <!ENTITY document.role "editors-copy">
 
 <!-- The following is used for document .-->
-<!ENTITY document.status "W3C Proposed Recommendation">
-<!ENTITY w3c.status "PR">
-<!ENTITY doctype "pr">
+<!ENTITY document.status "W3C Working Draft">
```
nigelmegitt: Shouldn't this be Editors Draft? I see that for WDs we have `entitieswd.dtd` available at line 4, in place of `entitiesedcopy.dtd`.

spec/entitiesedcopy.dtd
```diff
@@ -15,16 +15,16 @@
...
+<!ENTITY w3c.status "WD">
+<!ENTITY doctype "wd">
```
nigelmegitt: Likewise shouldn't the above two entries be `ED` and `ed`, respectively?

spec/entitiesedcopy.dtd
skynavga: No. It is an editor's draft of a WD, so we label it as WD, but the SoTD states "This document is an editor's copy that has no official standing," and the label on left of screen shows "W3C Editor's Draft". In actuality, I think we don't actually need (or use) `entitiesedcopy.dtd`.

spec/entitiesedcopy.dtd
skynavga: See above.

spec/entitiesedcopy.dtd
nigelmegitt: OK, I've got caught out by this before. I wish there was a way to say WD (but by the way this is an ED)!

PR #1050, opened 2019-03-20
Specify fixed, implied semantics for xlink:type and xlink:actuate (#1039).

### LINKED ISSUE #1039: Define fixed values for xlink:{actuate,type}.
Even though TTML2 does not specify the use of `xlink:actuate` or `xlink:type`, their semantics (as defined by [XLink]) still apply, and, in particular, their values (and consequent semantics) are presently defined in the TTML2 schemas as `#fixed`, with values:

```
xlink:actuate = "onRequest"
xlink:type    = "simple"
```

skynavga:
Closes #1039.


skynavga:
@nigelmegitt since this change pertains to implied semantics for a behavior that is essentially application dependent, I can't think of an explicit test we could write that would produce a consistent result either in a transformation or presentation processor

PR #1054, opened 2019-03-25
Clarify relative profile designator does not use xml:base (#1033).

### LINKED ISSUE #1033: Clarify that xml:base processing does not apply to <relative-profile-designator>.
Add note to §8.3.14 indicating that `xml:base` related processing does not apply when absolutizing a `<relative-profile-designator>`.

skynavga:
Closes #1033.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -6921,8 +6921,13 @@ the TT Profile Namespace, which serves as the base URI with which relative profi
 </tr>
 </tbody>
 </table>
+<note role="clarification">
+<p>When absolutizing a &lt;relative-profile-designator&gt;,
+<loc href="#content-attribute-xml-base">xml:base</loc> processing does not apply (because the TT Profile Namespace
+always serves as the base URI).</p>
+</note>
 <note role="example">
-<p>For example, if a relative profile designator is expressed as
+<p>For example, if a &lt;relative-profile-designator&gt; is expressed as
 <code>ttml2-presentation</code>, then the absolutized profile designator
 would be
 <code>http://www.w3.org/ns/ttml/profile/ttml2-presentation</code>.</p>


skynavga:
@nigelmegitt see https://github.com/w3c/ttml2/issues/1033#issuecomment-476710700

skynavga:
@nigelmegitt re: https://github.com/w3c/ttml2/pull/1054#pullrequestreview-218959049, please file a new issue so we can proceed with this PR; by the way, I believe  you are wrong about not needing `<fragment-profile-designator>`, but I prefer to have that conversation outside the scope of this PR;

nigelmegitt:
I'm really confused by your take on this @skynavga - are you saying that `<relative-profile-designator>` does not use xml:base, but that `<fragment-profile-designator>` _does_ use xml:base? If so, how can an implementation resolve this when the same URI can be interpreted equally as either one or the other?

spec/ttml2.xml
```diff
@@ -6976,8 +6976,13 @@ the TT Profile Namespace, which serves as the base URI with which relative profi
 </tr>
 </tbody>
 </table>
+<note role="clarification">
+<p>When absolutizing a &lt;relative-profile-designator&gt;,
+<loc href="#content-attribute-xml-base">xml:base</loc> processing does not apply (because the TT Profile Namespace
```
palemieux: RFC 3986 does not define "absolutizing". I think you might mean "resolving a reference as defined in RFC 3986".

nigelmegitt: The term "absolutize", or a derivative of it, is used 9 times in TTML2. This is probably worth opening a separate editorial issue for @palemieux .

nigelmegitt: @palemieux in the interests of moving this on, would it be okay for you to open a separate issue about the term "absolutize" used in relation to URLs, as I suggested at https://github.com/w3c/ttml2/pull/1054/files#r309211829 ?

palemieux: @nigelmegitt Sure. @skynavga or you should open an issue.

nigelmegitt: Raised #1150. @palemieux please could you re-review on this basis?

skynavga:
Hold pending merger of #1151.

PR #1069, opened 2019-04-12
Clarify non-applicability of character related styles in ruby container context (#1043).

### LINKED ISSUE #1043: Character-related style properties should not apply to ruby containers.
By definition, ruby containers (`tts:ruby='container'`, `tts:ruby='textContainer'` and `tts:ruby='baseContainer'`) do not have text node children. As a result, character-related style properties (see below) should not apply to them, just like they do not apply to `p`.

- textEmphasis
- textDecoration
- textShadow
- etc.

skynavga:
Closes #1043.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -12396,15 +12396,16 @@ likewise, an explicitly specified <loc href="#content-vocabulary-span"><el>span<
 <loc href="#style-attribute-ruby"><att>tts:ruby</att></loc> is <code>textContainer</code> is referred to as
 an <termdef id="defs-explicit-ruby-text-container" term=""><term>explicit ruby text container</term></termdef>.</p>
 <note role="clarification">
-<p>Style properties which apply to text children or <loc href="#terms-anonymous-span">anonymous span</loc>
+<p>Style properties which might apply to <loc href="#terms-anonymous-span">anonymous span</loc>
 children of a <el>span</el> element that functions as a <termref def="defs-ruby-container">ruby container</termref>,
 <termref def="defs-ruby-base-container">ruby base container</termref>, or <termref def="defs-ruby-text-container">ruby text container</termref>
-have no semantic affect during presentation processing since no significant text node appears as a child (or as a child of an
-<loc href="#terms-anonymous-span">anonymous span</loc> child) of such a <el>span</el> element. For example,
+have no semantic affect during presentation processing since no significant text node appears as a child of an
+<loc href="#terms-anonymous-span">anonymous span</loc> child of such a <el>span</el> element. For example,
 if a <loc href="#style-attribute-fontStyle"><att>tts:fontStyle</att></loc> attribute is specified on such a <el>span</el> element,
-then it has no significant text node in scope to which it may be applied. Note, however, that a value specified for
+then it has no significant text node in scope to which it may be applied; however, a value specified for
 <loc href="#style-attribute-fontStyle"><att>tts:fontStyle</att></loc> on such a <el>span</el> element still applies for the purpose of
-<loc href="#semantics-style-inheritance">style inheritance</loc>.</p>
+<loc href="#semantics-style-inheritance">style inheritance</loc> by other (non-anonymous) <el>span</el> element children of such a
+<el>span</el> element.</p>
 </note>
 </div4>
 </div3>
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -12094,7 +12094,10 @@ if the computed value of <att>tts:ruby</att> of a <loc href="#content-vocabulary
 <item><p>the computed value of <att>tts:ruby</att> of each of its child elements is not <code>none</code>;</p></item>
 <item><p>the computed value of <att>tts:ruby</att> of its first child element is <code>baseContainer</code>
 or <code>base</code>;</p></item>
-<item><p>each of its text node children contain only <loc href="#style-value-lwsp">&lt;lwsp&gt;</loc>;</p></item>
+<item><p>each of its text node children and <loc href="#terms-anonymous-span">anonymous span</loc> children
+contain only <loc href="#style-value-lwsp">&lt;lwsp&gt;</loc>, which must be ignored for the purpose of
+presentation processing regardless of whether the computed value of the <loc href="#content-attribute-xml-space"><att>xml:space</att></loc> attribute
+is <code>preserve</code> or not;</p></item>
 </ulist>
 <p>When using <att>tts:ruby</att>, the following nesting constraints apply to a <termref def="defs-ruby-base-container">ruby base container</termref>, that is,
 if the computed value of <att>tts:ruby</att> of a <loc href="#content-vocabulary-span"><el>span</el></loc> element is <code>baseContainer</code>, then:</p>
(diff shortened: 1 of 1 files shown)


palemieux:
I have proposed an alternative at #1071 , which explicitly specifies the style properties that do not apply to ruby containers, ruby base containers and ruby text containers.

In particular, it would be good to be clear about the applicability to audio-related properties.

spec/ttml2.xml
```diff
@@ -12094,7 +12094,10 @@ if the computed value of <att>tts:ruby</att> of a <loc href="#content-vocabulary
 <item><p>the computed value of <att>tts:ruby</att> of each of its child elements is not <code>none</code>;</p></item>
 <item><p>the computed value of <att>tts:ruby</att> of its first child element is <code>baseContainer</code>
 or <code>base</code>;</p></item>
-<item><p>each of its text node children contain only <loc href="#style-value-lwsp">&lt;lwsp&gt;</loc>;</p></item>
+<item><p>each of its text node children and <loc href="#terms-anonymous-span">anonymous span</loc> children
+contain only <loc href="#style-value-lwsp">&lt;lwsp&gt;</loc>, which must be ignored for the purpose of
+presentation processing regardless of whether the computed value of the <loc href="#content-attribute-xml-space"><att>xml:space</att></loc> attribute
+is <code>preserve</code> or not;</p></item>
```
palemieux: In the case of `div` elements, what causes LWSP to be pruned? Is it the `div` model, which does not admit `#PCDATA`? The ruby containers semantics should presumably follow the same approach.

skynavga: No, we cannot use the same approach. The content model for `div` is (https://www.w3.org/TR/xml/#sec-element-content)[element content] (elements only) while the content model for `span` is (https://www.w3.org/TR/xml/#sec-mixed-content)[mixed content] (elements and `#PCDATA`).

Whitespace in a `div` do not generate text nodes.

palemieux: > No, we cannot use the same approach.

What is the approach for `div`?

skynavga:
There is no “approach” for div. div is element content, not mixed content.
See the XML spec for the meaning of these terms.

spec/ttml2.xml
palemieux: @skynavga Where would I find statements constraining `div` to _element content_ or not _mixed content_?

skynavga: @palemieux see the content model for `div` under §8.1.4; if the content model does not include the token `#PCDATA`, then it is defined as _element content_; see also the ttml* schemas; on the other hand, the content model of `span` does include `#PCDATA1`, which means is defined as _mixed content_; when an element is defined as the former (_element content_), then no text nodes are generated for LWSP that happens to appear between its child elements; in contrast, for the latter (_mixed content_), LWSP does generate text nodes when parsed by a compliant XML parser;

skynavga: @palemieux BTW, only the `p` and `span` elements are defined as _mixed content_ element types;

palemieux: > which means it is defined as mixed content; when an element is defined as the former (element content), then no text nodes are generated
> for LWSP that happens to appear between its child elements; in contrast, for the latter (mixed content),
> LWSP does generate text nodes when parsed by a compliant XML parser;

An XML parser will in fact always return text nodes: it has no concept of mixed content in absence of a DTD, schema, etc. neither of which are part of the TTML content model.

I am suggesting that there is flexibility in clarifying, in the spec, whether #PCDATA is pruned or not.

skynavga:
We are in agreement. And PR #1069 does just that (which text you quote).
Are you suggesting a change is needed to the PR text?

palemieux:
> Are you suggesting a change is needed to the PR text?

I am comfortable with the handling of text nodes children within ruby containers: non LWSP text is forbidden and LWSP text is ignored. The alternative is to make the `span` data model conditional on attribute values, which is a significant departure from the current specification structure.

I do not think the note regarding style applicability is sufficient and accurate:
* it does not cover the applicability of styles in absence of LWSP
* style applicability is handled in the _Applies to_ section of each style property
#1071 resolves these shortcomings.

skynavga:
@palemieux I presume that, by

>it does not cover the applicability of styles in absence of LWSP

you mean,

>it does not cover the applicability of styles in absence of LWSP _when the author violates the constraint against non-LWSP text_

which, if correct, means you are asking for specification text to address error behavior semantics. Is that your intent?

Also, regarding "conditional" content models, this is not supported by either XSD or RNC schemas, so we can't go there.

palemieux:
>which, if correct, means you are asking for specification text to address error behavior semantics. Is that your intent?

No.

`tts:textDecoration` should not apply to ruby text container regardless of whether there are LWSP, according to https://github.com/w3c/ttml2/issues/1043#issuecomment-488472104.

> Also, regarding "conditional" content models, this is not supported by either XSD or RNC schemas, so we can't go there.

I agree.

spec/ttml2.xml
palemieux: In the balance, the modification suggested here seems preferable to making the `<span>` data model dependents on the value of an attribute.

spec/ttml2.xml
```diff
@@ -12386,6 +12395,18 @@ an <termdef id="defs-explicit-ruby-base-container" term=""><term>explicit ruby b
...
+children of a <el>span</el> element that functions as a <termref def="defs-ruby-container">ruby container</termref>,
+<termref def="defs-ruby-base-container">ruby base container</termref>, or <termref def="defs-ruby-text-container">ruby text container</termref>
+have no semantic affect during presentation processing since no significant text node appears as a child of an
+<loc href="#terms-anonymous-span">anonymous span</loc> child of such a <el>span</el> element. For example,
+if a <loc href="#style-attribute-fontStyle"><att>tts:fontStyle</att></loc> attribute is specified on such a <el>span</el> element,
+then it has no significant text node in scope to which it may be applied; however, a value specified for
+<loc href="#style-attribute-fontStyle"><att>tts:fontStyle</att></loc> on such a <el>span</el> element still applies for the purpose of
+<loc href="#semantics-style-inheritance">style inheritance</loc> by other (non-anonymous) <el>span</el> element children of such a
+<el>span</el> element.</p>
+</note>
```
palemieux: As suggested in #1071, I believe it is strongly beneficial to explicitly state that some properties do not apply to ruby containers, just as it is stated that `tts:rubyPosition` does not apply to some `span` elements.

nigelmegitt: Something we don't do in _any_ of the style attribute tables at the moment, is to clarify whether attributes apply to the character content children of (anonymous) `span` elements (e.g. `color`) or to the boxes generated by them (e.g. `backgroundColor`). Doing this might help resolve the discussion here and clarify for implementers and spec readers?

skynavga: I disagree. We do not need to add clutter to the tables. The meaning of _applies_ is delegated to the prose for each defined style. This is consistent with past TTML practice as well as XSL FO and CSS. Furthermore, we do already have general language regarding anonymous spans, i.e.,

§10 Preamble

>For the purpose of determining the applicability of a style property, if the style property is defined so as to apply to a span element, then it also applies to anonymous span elements.

nigelmegitt: > The meaning of _applies_ is delegated to the prose for each defined style.

This is not actually clear in the spec.

skynavga:
Closing this PR which content has been sub-divided into and augmented by: #1100, #1091, #1090, #1088, #1086, #1084, #1082, #1080, #1078, and #1076.

PR #1071, opened 2019-04-25
Clarify which style properties do not apply to ruby containers

### LINKED ISSUE #1043: Character-related style properties should not apply to ruby containers.

### LINKED ISSUE #1078: Application semantics of text decoration.
§10.2.43 specifies the semantics of _application_ (applies) as

> to apply to glyph areas or other inline areas

language that has been present since TTML1 1ed [1]. However, a review of both XSL 1.1 [1] and CSS2.1 [3] calls into question the reference to "or other inline areas", which, with the exception of the ``blink`` value, is clearly applied by both specifications **only** to glyph areas (text). Since TTML does not support the ``blink`` value, the semantics of application are essentially limited to application to glyph areas, and not inline areas that are not glyph areas. In order to address the ambiguity created by stating that text decoration applies to something that it cannot logically apply to, it is recommended that "or other inline areas" be removed.

[1] https://www.w3.org/TR/2010/REC-ttaf1-dfxp-20101118/#style-attribute-textDecoration
[2] https://www.w3.org/TR/2006/REC-xsl11-20061205/#text-decoration
[3] https://www.w3.org/TR/CSS21/text.html#propdef-text-decoration

### LINKED ISSUE #1080: Application semantics of text emphasis.
When authored in TTML2 1ed [1], the application (applies) semantics of text emphasis was copied from the definition of text decoration, inadvertently including the text "or other inline areas", which should be removed since text emphasis cannot be applied to inline areas that are not glyph areas. This is an editorial change to remove unintentional language which has no affect on conformance.

[1] https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-attribute-textEmphasis

### LINKED ISSUE #1082: Application semantics of font selection strategy.
Clarify that application semantics of the font selection strategy property applies to glyph areas, which may take into account more or less character context to determine that selection according to the value of the property.

### LINKED ISSUE #1084: Application semantics of font variant.
Clarify that application semantics of the font variant property applies to glyph areas.

### LINKED ISSUE #1086: Application semantics of text combine.
Clarify that application semantics of the text combine property applies to glyph areas.

### LINKED ISSUE #1100: Emphasize non-applicability of styles to ignored text nodes in ruby container spans.
For avoidance of doubt on the part of a reader of the specification, add an informative note that emphasizes the non-applicability of character or glyph area related styles to ignored text nodes in the context of ruby container spans.

palemieux:
Closes #1043 
Closes #1078 
Closes #1080 
Closes #1082 
Closes #1084 
Closes #1086 
Closes #1100


palemieux:
> as it violates editorial and design principles that have been consistently applied to the TTML specifications. 

I see no evidence behind that statement. In fact, TTML today clearly specifies to which vocabulary each style property applies.

skynavga:
@palemieux the general design/editorial principle in play here is: do not specify something (in normative text) which is a logical consequence of other normative statements; in the current case, you have proposed adding qualifying language (i.e., claiming non-applicability) for 18 style attributes;

a close reading of each of these attributes' specifications show that 3 of them still need to apply to ruby container spans (`tts:direction`, `tts:wrapOption`, and `tts:unicodeBidi`), and that all but one of the remaining 15, namely, `tts:textDecoration`, cannot logically apply since there are no generated glyph area children or glyph area children of an inline area generated by a child anonymous span;

that leaves us with only one possibly case to consider, `tts:textDecoration`; now, the current language in TTML2 ([§10.2.43](https://www.w3.org/TR/ttml2/#style-attribute-textDecoration)) says it applies "to glyph areas or other inline areas", and, reading XSL FO 1.1 ([§4.9.5](https://www.w3.org/TR/xsl11/#rend-intrinsic) and [§7.17.4](https://www.w3.org/TR/xsl11/#text-decoration)) and CSS 2 ([§16.3.1](https://www.w3.org/TR/CSS2/text.html#propdef-text-decoration)), and particularly the latter, we find the following

>Underlines, overlines, and line-throughs are applied only to text (including white space, letter spacing, and word spacing): margins, borders, and padding are skipped. User agents must not render these text decorations on content that is not text. For example, images and inline blocks must not be underlined.

now, since we have already specified that non-LWSP text is forbidden from ruby container spans and we will have added (in #1069) that LWSP is ignored for presentation processing, then there is no text (including white space, letter spacing, and word spacing), i.e., no _glyph areas_ to which text decoration could be logically applied;

in evaluating these details, it is quite clear that we do not have to specify that something does not apply to something that cannot be present or is explicitly ignored; consequently, the note added by #1069 to §10.2.35.1 suffices, by reminding the reader of this fact (and that does not need to be otherwise stated in normative text);

notwithstanding the above, I believe there are a few improvements we could make in #1069, such as

- make clear that `tts:fontSelectionStrategy`, `tts:fontVariant`, and `tts:textCombine` apply only to glyph areas;
- remove _or other inline areas_ from the preamble for `tts:textDecoration` and `tts:textEmphasis`

palemieux:
@skynavga Your argument is disproved by the simple fact that character-based style attributes do not apply to `region`, `div`, etc...

skynavga:
@palemieux you keep mixing up applicability of styles to elements and applicability of styles to elements in a specific context; character styles can never apply to `region`, `div`, etc., since they can never contain text to which it might apply; on the other hand a `span` can always contain text, and, therefore, character properties generally apply unless something makes such application logically impossible for other reasons, for example, not having any generated glyph area children to which it might apply; we don't say that "character styles" don't apply to empty spans, do we? in any case, you are neglecting the entirety of my detailed response by replying in such a flippant manner without considering the details

palemieux:
@skynavga According to CSS 2.1, _applies_ to" indicates whether _properties have no rendering effect on some types of elements_. `tts:textEmphasis`, for instance, have no rendering effect on ruby containers, and thus does not apply to ruby containers.

skynavga:
@palemieux I guess I need to say it for a 3rd, or is this the 4th time, a ruby container is not an element; a `span` is an element; what is the point you are trying to make in all of these repetitive claims about applicability? have you read my detailed response at https://github.com/w3c/ttml2/pull/1071#issuecomment-491101678

nigelmegitt:
I've re-read @skynavga 's https://github.com/w3c/ttml2/pull/1071#issuecomment-491101678 and agree that the text decorations are not supposed to create any marks other than on text content, and since there isn't any in a container span, we don't need to say anything. The screenshot at https://github.com/w3c/ttml2/issues/1043#issuecomment-488462399 therefore looks like a browser bug.

I haven't checked that similar wording applies to all the other style properties that seemingly should not cause any marks to be made if applied to a container span. @skynavga or @palemieux did you check that?

skynavga:
@nigelmegitt yes, I checked the specification text for each of the 18 properties, as I stated in https://github.com/w3c/ttml2/pull/1071#issuecomment-491101678, 3 of those actually should apply (bidi and wrap related), and the other 15 cannot apply because they semantically apply only to glyph areas

palemieux:
> the other 15 cannot apply because they semantically apply only to glyph areas

@skynavga You are missing my point. I am arguing for specification clarity, and providing unambiguous guidance to implementers. According to your [analysis](https://github.com/w3c/ttml2/pull/1071#issuecomment-491101678), the applicability of style properties is not trivial. The _applies to_ attribute of each style property definition provides a simple and straightforward means to communicating this applicability. If `tts:textDecoration` does not apply to `span` elements that are ruby containers, then `applies to` should note this,

skynavga:
@palemieux `tts:textDecoration` and all the other 17 properties you want to constrain do apply to `span`s including `span`s that serve as ruby containers; 3 of these 18 apply in general, and the other 15 apply to glyph areas generated by the affected `span`s, and, since there are no glyph areas, application is a NO-OP; so we don't need to say anything at all, but we have at least added a comment in §10.2.35.1 that points out this fact; so your claim that these properties do not apply is not correct, they do apply;

palemieux:
@nigelmegitt and @cconcolato I have introduced the defined term "ruby container span" as suggested.

spec/ttml2.xml
```diff
@@ -12094,7 +12100,10 @@ if the computed value of <att>tts:ruby</att> of a <loc href="#content-vocabulary
 <item><p>the computed value of <att>tts:ruby</att> of each of its child elements is not <code>none</code>;</p></item>
 <item><p>the computed value of <att>tts:ruby</att> of its first child element is <code>baseContainer</code>
 or <code>base</code>;</p></item>
-<item><p>each of its text node children contain only <loc href="#style-value-lwsp">&lt;lwsp&gt;</loc>;</p></item>
+<item><p>each of its text node children and <loc href="#terms-anonymous-span">anonymous span</loc> children
```
cconcolato: Same comment as in the other PR, I think the presentation processing behavior should not be mixed with the content model. I'd prefer having a separate paragraph.

palemieux:
Overtaken by #1100, #1129 and #1130.

PR #1072, opened 2019-05-02
Correct the sRGB EOTF gamma in the HDR compositing example (Annex Q.1)

### LINKED ISSUE #1070: Gamma approximation in luminance gain.
From a discussion with @joedrago:
> a minor note about SRGB's transfer function. SRGB's transfer function approximates much closer to a gamma of 2.2, despite the value of 2.4 being present in the only exponent of the function. Here is a reference from color.org indicating its approximation:
http://www.color.org/srgb.pdf
And here is a Desmos graph showing 2.2, 2.4, and SRGB: 
https://www.desmos.com/calculator/az6xrrvj8v

palemieux:
Closes #1070

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -28964,12 +28964,11 @@ application based on <bibref ref="html52"/>), rather than of the Document Instan
 <div2 id="pq-hdr">
 <head>Perceptual Quantizer</head>
 <p>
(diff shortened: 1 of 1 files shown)


palemieux:
Per https://github.com/w3c/ttml2/issues/1070#issuecomment-491936312, I recommend withholding the PR until progress is made on the ticket.

PR #1077, opened 2019-05-23
Ignore text node children of ruby container spans (#1076).

### LINKED ISSUE #1076: Text node children of ruby container spans.
Add language requiring text node children of ruby container spans or anonymous span children of ruby container spans to be ignored for presentation processing. See also #1043.

skynavga:
This PR (and associated issue) break out the specific changes found in PR #1069 with respect to presentation semantics of text node children of ruby container spans.

Closes #1076.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -12094,7 +12094,9 @@ if the computed value of <att>tts:ruby</att> of a <loc href="#content-vocabulary
 <item><p>the computed value of <att>tts:ruby</att> of each of its child elements is not <code>none</code>;</p></item>
 <item><p>the computed value of <att>tts:ruby</att> of its first child element is <code>baseContainer</code>
 or <code>base</code>;</p></item>
-<item><p>each of its text node children contain only <loc href="#style-value-lwsp">&lt;lwsp&gt;</loc>;</p></item>
+<item><p>each of its text node children contain only <loc href="#style-value-lwsp">&lt;lwsp&gt;</loc>,
+which must be ignored for the purpose of presentation processing regardless of whether the computed value of the
+<loc href="#content-attribute-xml-space"><att>xml:space</att></loc> attribute is <code>preserve</code> or not;</p></item>
 </ulist>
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -12094,7 +12094,9 @@ if the computed value of <att>tts:ruby</att> of a <loc href="#content-vocabulary
 <item><p>the computed value of <att>tts:ruby</att> of each of its child elements is not <code>none</code>;</p></item>
 <item><p>the computed value of <att>tts:ruby</att> of its first child element is <code>baseContainer</code>
 or <code>base</code>;</p></item>
-<item><p>each of its text node children contain only <loc href="#style-value-lwsp">&lt;lwsp&gt;</loc>;</p></item>
+<item><p>each of its text node children contain only <loc href="#style-value-lwsp">&lt;lwsp&gt;</loc>,
+which must be ignored for the purpose of presentation processing regardless of whether the computed value of the
```
cconcolato: Instead of "which must be ignored ... preserve or not", it should rather say "These text node children are not considered significant text node for the purpose of performing the [construct anonymous spans] procedure."

skynavga: @cconcolato your suggested rewrite won't work because the [construct anonymous spans] procedure is invoked before style resolution; consequently, that procedure would not (and should not) have any knowledge about the computed value of the `tts:ruby` that it could employ to perform the special case processing you suggest; the correct place for lwsp to be ignored is going to be in the context of performing layout of the already styled spans that serve as ruby containers;

skynavga:
Per https://github.com/w3c/ttml2/pull/1077#pullrequestreview-241647075, and since there are no objections and 14 days have passed, I am merging this approved PR. Note also that, per https://github.com/w3c/ttml2/pull/1077#discussion_r289967294, the suggested change in https://github.com/w3c/ttml2/pull/1077#discussion_r288693991 cannot be adopted.

skynavga:
Tests at w3c/ttml2-tests#247.

PR #1079, opened 2019-05-23
Clarify text decoration application (#1078).

### LINKED ISSUE #1078: Application semantics of text decoration.

skynavga:
Closes #1078.


nigelmegitt:
Please do not merge this until a 2 week review period has passed.

skynavga:
Regarding https://github.com/w3c/ttml2/pull/1079#pullrequestreview-243020084, a _ruby container_ is not an element, and, therefore, cannot be removed from the set `{ span }`. You comment is non-actionable.

PR #1081, opened 2019-05-23
Clarify text emphasis application (#1080).

### LINKED ISSUE #1080: Application semantics of text emphasis.

skynavga:
Closes #1080.


skynavga:
Regarding https://github.com/w3c/ttml2/pull/1081#pullrequestreview-243020299, a _ruby container_ is not an element, and, therefore, cannot be removed from the set `{ span }`. You comment is non-actionable.

PR #1083, opened 2019-05-24
Clarify application of font selection strategy (#1082).

### LINKED ISSUE #1082: Application semantics of font selection strategy.

skynavga:
Closes #1082.


skynavga:
Regarding https://github.com/w3c/ttml2/pull/1083#pullrequestreview-243020509, a _ruby container_ is not an element, and, therefore, cannot be removed from the set `{ span }`. You comment is non-actionable.

PR #1087, opened 2019-05-24
Clarify text combine application (#1086).

### LINKED ISSUE #1086: Application semantics of text combine.

skynavga:
Closes #1086.


spec/ttml2.xml
```diff
@@ -13108,11 +13108,12 @@ follows the semantics of the <code>text-align-last</code> property with a value
 <div3 id="style-attribute-textCombine">
 <head>tts:textCombine</head>
 <p>When a vertical writing mode applies, the <att>tts:textCombine</att> attribute is used to specify a style property that
-determines whether and how multiple characters are combined horizontally so that their 
-<loc href="#terms-glyph-area">glyph areas</loc> fit within the width of a single line of the surrounding vertical text.
+applies to <loc href="#terms-glyph-area">glyph areas</loc> generated by content flowed into a region, and which
```
nigelmegitt: It also applies to inline areas because it affects their layout.

spec/ttml2.xml
```diff
@@ -10689,7 +10689,7 @@ of typographic glyph variants.</p>
 </table>
 <p>In addition to applying to <loc href="#content-vocabulary-span"><el>span</el></loc>, the computed value of this property
 on a <loc href="#content-vocabulary-span"><el>p</el></loc> element is used to resolve the <code>normal</code> value of the
-style property associated with the <loc href="#style-attribute-lineHeight">tts:lineHeight</loc> attribute as specified in
+style property associated with the <loc href="#style-attribute-lineHeight">ttslin:eHeight</loc> attribute as specified in
```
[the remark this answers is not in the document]

spec/ttml2.xml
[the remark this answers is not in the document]
skynavga: Formulaic language used elsewhere.

skynavga: In what way does it apply to inline areas that are not glyph areas? Please be specific.

nigelmegitt: As per the text in the section of the spec about `tts:textCombine`, the dimension of line areas (and I assume therefore also potentially inline areas) can be affected in that their block progression dimension size can change as a result of using this attribute.

skynavga: I agree that layout (of the containing line area) may be affected, but that doesn't mean we need to say that this property applies to inline areas any more than we might need to say that, e.g., `tts:fontSize` applies to (non-glyph) inline areas, a statement that would be required if we follow your logic.

The reason we don't want to say that text combine applies to a (non-glyph) inline area, is that we have (non-glyph) inline areas generated by a number of other different elements, such as (1) a `span` with `tts:display="inlineBlock"`, (2) a `span` with `tts:ruby="textContainer"`, and `image` in an inline context, none of which we want to combine as with glyph inline areas on which the `tts:textCombine` style property is designed to operate.

nigelmegitt: OK I see, that makes sense, thanks @skynavga .

skynavga:
Regarding https://github.com/w3c/ttml2/pull/1087#pullrequestreview-243021217, a ruby container is not an element, and, therefore, cannot be removed from the set { span }. You comment is non-actionable.

PR #1089, opened 2019-05-24
Clarify use of 'applies to' in style property definition tables (#1088).

### LINKED ISSUE #1088: Expand on style property definition table.
In the preamble to §10.2, expand the description of style property definition table to refer to CSS definitions and summary nature of table, referring to prose for further details.

skynavga:
Closes #1088.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -8375,9 +8375,7 @@ whether (and how) style is animatable, and semantic basis (derivation).</p>
 <note role="elaboration">
 <p>For the purpose of interpreting the entries of a
 <termref def="defs-style-property-definition-table">style property definition table</termref>, the definition of <emph>applies to</emph>
-specified by <bibref ref="css2"/>, &sect;1.4.2.3, applies to this specification as well. Moreover, the information described by the
-<emph>applies to</emph> entries of these tables are intended to be of a general nature, where the details of the semantics of style application
-are delegated to the prose that defines each style property.</p>
+specified by <bibref ref="css2"/>, &sect;1.4.2.3, applies to this specification.</p>
 </note>
 <p>For animatable styles, the term <emph>discrete</emph> refers to the use of either a <loc href="#animation-vocabulary-set"><el>set</el></loc>
 element or an <loc href="#animation-vocabulary-animate"><el>animate</el></loc> element with the <code>discrete</code> value for its
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -8372,6 +8372,13 @@ that support inline style specifications:</p>
 specifies one or more of the following aspects of the style: value syntax, initial value,
 elements to which style semantically applies, whether style is inherited or not, how percentage values are interpreted (if applicable),
 whether (and how) style is animatable, and semantic basis (derivation).</p>
+<note role="elaboration">
+<p>For the purpose of interpreting the entries of a
+<termref def="defs-style-property-definition-table">style property definition table</termref>, the definition of <emph>applies to</emph>
+specified by <bibref ref="css2"/>, &sect;1.4.2.3, applies to this specification as well. Moreover, the information described by the
+<emph>applies to</emph> entries of these tables are intended to be of a general nature, where the details of the semantics of style application
+are delegated to the prose that defines each style property.</p>
+</note>
 <p>For animatable styles, the term <emph>discrete</emph> refers to the use of either a <loc href="#animation-vocabulary-set"><el>set</el></loc>
 element or an <loc href="#animation-vocabulary-animate"><el>animate</el></loc> element with the <code>discrete</code> value for its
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -8372,6 +8372,13 @@ that support inline style specifications:</p>
 specifies one or more of the following aspects of the style: value syntax, initial value,
 elements to which style semantically applies, whether style is inherited or not, how percentage values are interpreted (if applicable),
 whether (and how) style is animatable, and semantic basis (derivation).</p>
+<note role="elaboration">
+<p>For the purpose of interpreting the entries of a
+<termref def="defs-style-property-definition-table">style property definition table</termref>, the definition of <emph>applies to</emph>
+specified by <bibref ref="css2"/>, &sect;1.4.2.3, applies to this specification as well. Moreover, the information described by the
+<emph>applies to</emph> entries of these tables are intended to be of a general nature, where the details of the semantics of style application
```
nigelmegitt: I appreciate the intent here but I would like to be even more specific. Suggest:

```
The elements listed in the <emph>applies to</emph> entries of these tables are those to which the style
property has or potentially has some processing effect including to their layout or to their content; 
where the styling of character content is affected, only the parent elements of that character content
are listed. The details of the semantics of style application are delegated to the prose that defines each
style property, and any referenced semantic derivation.
```

skynavga: So you are suggesting that we don't refer to the CSS definition, but rewrite it (the CSS definition) instead? I think that is a bad approach, since those semantics are used widely by W3C specs. If you want to clarify the CSS definition as it applies to TTML, then could you propose additional text to be added in a new paragraph of this note rather than changing the first paragraph itself, which I don't see you disagreeing with.

nigelmegitt: Sorry I wasn't clear which text I wanted to be more specific about. The full paragraph I'm proposing is:

```
<p>For the purpose of interpreting the entries of a
<termref def="defs-style-property-definition-table">style property definition table</termref>,
the definition of <emph>applies to</emph>
specified by <bibref ref="css2"/>, &sect;1.4.2.3, applies to this specification as well.
Moreover, the elements listed in the <emph>applies to</emph> entries of these tables are those on which the style
property has or potentially has some processing effect including to their layout or to their content; 
where the styling of character content is affected, only the parent elements of that character content
are listed. The details of the semantics of style application are delegated to the prose that defines each
style property, and any referenced semantic derivation.
```

cconcolato:
The CSS specification seems clear that the "applies to" line is normative and not "intended to be of a general nature" without normative impact as the PR seems to suggest.
> 1.4.2.3 Applies to
This part lists the elements to which the property applies. All elements are considered to have all properties, but some properties have no rendering effect on some types of elements. For example, the 'clear' property only affects block-level elements."

What matters is for implementers to be sure that if the condition given in the "applies to" line is not meant, then they can ignore the property for rendering (not for inheritance).

nigelmegitt:
I could accept just the CSS definition. In other words, in the proposal, delete "Moreover" and all the new text after it.

palemieux:
The updated prose resolves my comment at https://github.com/w3c/ttml2/pull/1089#pullrequestreview-243022248 . I am keeping my blocker so that all related PRs can be evaluated in light of #1043

skynavga:
I'd like to point out that XSL-FO 1.1 includes the following elaboration in its use of _applies to_ in [Appendix B.4](https://www.w3.org/TR/xsl11/#prapply) (emphasis added).

>For each property the formatting objects it applies to is listed. **It should be noted, however, that for some formatting objects there are qualifications on applicability or values permitted.**

Since TTML based its original use of _applies to_ on XSL-FO, by way of analogy, we should probably include this elaboration in the note proposed by this PR.

nigelmegitt:
@skynavga I don't think the elaboration adds much of any use: if the Applies to row on the style attribute table further qualifies the set of elements, that is obvious from the context.  I don't think there's anything incorrect in the content of the elaboration though.

skynavga:
@nigelmegitt the point of my mentioning this was that XSL-FO is saying that the prose of style property definitions define further qualifications on _applies to_, and that the appendix B.4 (in XSL-FO), which effectively serves the same function of our (and CSS's) _applies to_ entry in style property definition tables, does not itself specify such qualifications: meaning, we do not need to place such qualifications in our _applies to_ table entries either, and might want to add similar language in this PR beyond simply citing CSS;

nigelmegitt:
@skynavga that may well be the intent of the text but taking it out of context and just looking at the additional sentence, there's nothing to indicate _where_ the additional qualifications are specified, so it doesn't serve the purpose you may have been hoping it does.

Even with the additional context behind the proposal I still think that transplanting it into TTML would not add anything useful, and for that reason I think we should not do so.

skynavga:

skynavga:
Based upon today's TTWG call, we have full approval to merge this (which I reopened above).

PR #1092, opened 2019-05-25
Clarify ruby align application (#1090).

### LINKED ISSUE #1090: Application of ruby align.
For the sake of consistency, remove the text "only if the computed value of ``tts:ruby`` is ``container``" from the _applies to_ entry of the property definition table for `tts:rubyAlign`. Moreover, make clear (in prose) that the semantics of application is the empty set for spans for which the ``tts:ruby`` attribute's computed value is not ``container``.

skynavga:
Closes #1090.


skynavga:
@nigelmegitt we need to either remove this editorial inconsistency (which gives rise to the misreading that _applies to_ in this context applies to something that is not an _element_) or create more editorial inconsistency (by adding similar language to other properties); as this is an editorial matter, and not a substantive matter, I invoke my privilege as editor to insist upon and ensure editorial consistency in this regard: in particular, I will not accept creating further inconsistency by adding incorrect and misleading language to a raft of style properties which does nothing more than create semantic chaos;

skynavga:
@nigelmegitt just to be clear, we formally object to adding this language (that we believe should be removed) elsewhere: there is no technical reason to add it and it is either wrong or misleading and inconsistent with the meaning of _applies to_ as used in style definition tables; while I find it disconcerting to not remove the language, I could live with it remaining; however, I cannot live with it being added elsewhere when it should not have been added in `tts:rubyAlign` and `tts:rubyPosition` in the first place, for all the reasons I have stated elsewhere;

spec/ttml2.xml
```diff
@@ -12412,9 +12412,7 @@ an <termdef id="defs-explicit-ruby-text-container" term=""><term>explicit ruby t
 </tr>
 <tr>
 <td><emph>Applies to:</emph></td>
-<td>
-<loc href="#content-vocabulary-span"><el>span</el></loc> only if the computed value of <loc href="#style-attribute-ruby">tts:ruby</loc> is <code>container</code>
-</td>
+<td><loc href="#content-vocabulary-span"><el>span</el></loc></td>
```
palemieux: This text was the result of careful discussions (#717 and #740), and remains correct: `tts:rubyAlign` does not apply to ruby containers, since they cannot contain child glyph areas (see #1043 and #1071).

skynavga: A _ruby container_ is not an element; therefore, by the CSS definition of _applies to_ in a style definition table, one cannot say that a property does or does not apply to a _ruby container_ in the _applies to_ row of a style definition table.

skynavga:

PR #1093, opened 2019-05-25
Clarify ruby position application (#1091).

### LINKED ISSUE #1091: Application of ruby position.
For the sake of consistency, remove the text "only if the computed value of `tts:ruby` is `textContainer` or `text`" from the applies to entry of the property definition table for `tts:rubyPosition`. Moreover, ensure (in prose) that the semantics of application is the empty set for spans for which the `tts:ruby` attribute's computed value is not `textContainer` or `text`.

skynavga:
Closes #1091.


skynavga:
Regarding https://github.com/w3c/ttml2/pull/1093#pullrequestreview-243023999, a _ruby container_ is not an element; therefore, by the CSS definition of _applies to_ in a style definition table, one cannot say that a property does or does not apply to a _ruby container_ in the _applies to_ row of a style definition table.

skynavga:

PR #1096, opened 2019-05-25
Add module framework (#1066).

### LINKED ISSUE #1066: Add hooks to TTML2 to simplify the creation of external modules.
* extend style resolution to include style properties defined in external modules
* broaden the definition of _TTML Content Document Type_ to include elements in external modules
* extend the definition of _content element_ to include elements defined in external modules
* extend the definition of `condition` to include elements defined in external modules

See also https://github.com/w3c/ttml3/issues/29

skynavga:
Closes #1066.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -870,8 +870,7 @@ and the <loc href="#element-vocab-type-image">Image Module</loc> when they occur
   <specref ref="element-vocab-table"/>.</p>
 </note>
 <p>In addition to the element types cited above, element types designated as content element types by an
-<loc href="#terms-external-module">external module</loc> that is listed in the <loc href="#terms-module-registry">module registry</loc>
-are also considered to be content elements.</p>
+<loc href="#terms-extension-module">extension module</loc>.</p>
 </def>
 </gitem>
 <gitem id="terms-content-image">
@@ -905,6 +904,13 @@ must not, must, or may be employed by Timed Text Markup Language content.</p>
   defined by an explicitly specified or an implied <loc href="#layout-vocabulary-region"><el>region</el></loc> element
   which defines the extent of the content region in relation to a <loc href="#terms-root-container-region">root container region</loc>
   as well as other stylistic and temporal characteristics.</p>
+</def>
+</gitem>
+<gitem id="terms-core-module">
+<label>[core module]</label>
+<def>
+<p>A <loc href="#terms-module">module</loc> defined within the body of this specification, specifically, the modules defined by
+<specref ref="element-vocab-table"/> and <specref ref="attribute-vocab-table"/>.</p>
 </def>
 </gitem>
 <gitem id="terms-data-binding-context">
@@ -1140,6 +1146,16 @@ labeled (using a extension designation) in another (public or private)
 specification.</p>
 </def>
 </gitem>
+<gitem id="terms-extension-module">
+<label>[extension module]</label>
+<def>
+<p>A <loc href="#terms-module">module</loc> that is not labeled as a <loc href="#terms-core-module">core module</loc>. An extension module
+is designated at the time of specification as either a W3C defined module, in which case any vocabulary or semantic
+behavior it defines is specified in the scope of one or more TTML Namespaces listed in <specref ref="namespace-vocab-table"/>,
+or as a non-W3C defined module, in which case its vocabulary and semantic behavior is specified in the scope
+of one or more namespaces not listed in <specref ref="namespace-vocab-table"/>.</p>
+</def>
+</gitem>
 <gitem id="terms-extension-specification">
 <label>[extension specification]</label>
 <def>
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -16675,7 +16733,7 @@ element type of <emph>E</emph> is not animation element type <el>animate</el> or
 and is not styling element type <el>style</el>,
 then for each style property
 <emph>P</emph> in the set of style properties defined
-above in <specref ref="styling-attribute-vocabulary"/>, perform the
+in <specref ref="styling-attribute-vocabulary"/> or in a supported <loc href="#terms-external-module">external module</loc>, perform the
```
nigelmegitt: I don't think this does enough to satisfy the issue request about adding hooks to allow modules to declare attributes as style attributes for the purposes of style processing. I would like to add some kind of classification of label that allows any attribute to be declared as participating in the style resolution algorithm, such that attributes defined in external modules can be declared in a uniform way as being "style attributes" for this purpose.

As drafted here, there is no common terminology that can be used to declare an externally defined attribute is a "style attribute".

skynavga:
@nigelmegitt you do realize that these changes are exactly what we have already accepted and merged into TTML3, and that making a change here would create an inconsistency with TTML3? see https://github.com/w3c/ttml3/pull/30 for confirmation.

nigelmegitt:
@skynavga I seem to recall there was disquiet about those changes in TTML3 also. In my view it may be that the best approach is to fix it here and then port these changes as incremental changes onto TTML3 also, or it may be that we think TTML3 has a different set of constraints on modules than TTML2. Either way, my comments still stand.

skynavga:
@nigelmegitt can we merge this PR and deal with your comments by way of another issue? since it proposes an additional follow-on substantive change that would affect both this spec and TTML3?

spec/ttml2.xml
palemieux: First, while I did not object to heavy module machinery suggested for TTML3, I do not think it is necessary in TTML2, where changes should limited. It would therefore be ideal if a simpler approach could be taken.

Second, if the argument is that TTML2 text should reflect TTML3 text, then let's resolve https://github.com/w3c/ttml3/issues/32 as part of this issue.

skynavga:
I have attempted to address various comments in https://github.com/w3c/ttml2/pull/1096/commits/2f67bbdd1c2b9c71f5de0b8675a20e5238e88c5c, and, in particular, I have changed the adjectives internal/external to core/extension w.r.t. to characterizing modules and simplified language.

I have also removed the definition and references to a TTWG controlled Module Registry, which I still think is a good idea that will assist interoperability and implementation activity.

The net effect of removing the registry is that anyone may define an extension module that adds a content element or an attribute that it wishes to be treated as a style property regardless of whether it is in a TTWG controlled namespace or not.

I firmly believe this will lead to interoperability problems in the future, but we shall see.

spec/ttml2.xml
```diff
@@ -1137,6 +1146,16 @@ labeled (using a extension designation) in another (public or private)
...
 </def>
 </gitem>
+<gitem id="terms-extension-module">
+<label>[extension module]</label>
+<def>
+<p>A <loc href="#terms-module">module</loc> that is not labeled as a <loc href="#terms-core-module">core module</loc>. An extension module
+is designated at the time of specification as either a W3C defined module, in which case any vocabulary or semantic
+behavior it defines is specified in the scope of one or more TTML Namespaces listed in <specref ref="namespace-vocab-table"/>,
+or as a non-W3C defined module, in which case its vocabulary and semantic behavior is specified in the scope
+of one or more namespaces not listed in <specref ref="namespace-vocab-table"/>.</p>
```
palemieux: Why is the distinction between W3C defined module and non-W3C defined module necessary?

skynavga: Addressed in https://github.com/w3c/ttml2/pull/1096/commits/98f3fa320e277ff9dbfca07012e5262d8ce7458d.

palemieux:
@skynavga I have approved the PR. Cyril and Nigel had much stronger opinions on this since they are writing "modules", so I encourage you to check with them.

skynavga:
@cconcolato @nigelmegitt please check

skynavga:
@palemieux please mark your conversation https://github.com/w3c/ttml2/pull/1096#pullrequestreview-261049337 as resolved, or ask me to do so

skynavga:
Even though this PR impacts both document syntax and processing semantics, it is not interoperably testable as such, since whether or not a given implementation implements a feature defining module is dependent upon that implementation, and, if it does not implement it, then the default pruning rules will prune any elements or attributes added by a given module specification. Note, however, that element and/or attribute syntax associated with a given module specification could be tested for validity on an implementation that does support the module. Such tests would also need to specify a profile that requires support for any feature defined by the associated module that is being tested.

skynavga:
Marking this issue as an enhancement since the omission of support for modularization was not an oversight.

spec/ttml2.xml
skynavga: @nigelmegitt see https://github.com/w3c/ttml2/pull/1096/commits/d54693b114ec31930523b31daa086b8baa29a774

nigelmegitt: Thanks @skynavga this works for me.

skynavga:
@cconcolato I've tried to address your comment in 7c3ef1c:

>the current definition of modules mixes everything from syntactical elements (attributes) to specifications, which is strange to me

there are no other outstanding, actionable comments at this time, so based on the current three approvals, I'm proceeding with merging

PR #1097, opened 2019-05-26
Add #presentation-audio feature (#1068).

### LINKED ISSUE #1068: Create an `#audio-presentation` feature.
The [`#presentation`](https://www.w3.org/TR/ttml1/#feature-presentation) feature designator requires support for region layout and presentation, and line layout, neither of which is relevant to a TTML2 presentation processor whose rendered output is only audio. 

This means that an audio-only profile of TTML2 would not necessarily support `#presentation` and therefore would not support the TTML2 Presentation Profile, the intent of which is to be a minimal profile for presentation. 

The consequence is it would be reasonable to define a kind of presentation processor profile that does not conform to the minimal presentation profile requirements, which sounds nonsensical.

I propose that we add a new feature designator that applies to audio presentation only, which does not include the constraints implied by visual text layout requirements. For completeness we also could consider creating a minimal TTML2 audio presentation processor profile.

skynavga:
Closes #1068.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -21728,6 +21728,25 @@ transforming the <loc href="#style-attribute-position"><att>tts:position</att></
 <p>A TTML content processor supports the <code>#presentation</code>
 feature if it supports the <code>#presentation</code> feature as defined by <bibref ref="ttml1"/>, &sect;D.1.69.</p>
 </div3>
+<div3 id="feature-presentation-audio">
+<head>#presentation-audio</head>
+<p>A TTML content processor supports the <code>#presentation-audio</code> feature if it
+(1) satisfies the generic processor criteria defined by <specref ref="conformance-generic-processor"/>,
+(2) implements support for audio presentation (rendering) semantics, and
+(3) implements audio presentation semantics for the following features:
+</p>
+<ulist>
+<item><p><loc href="#feature-audio"><code>#audio</code></loc></p></item>
+<item><p><loc href="#feature-content"><code>#content</code></loc></p></item>
+<item><p><loc href="#feature-core"><code>#core</code></loc></p></item>
+<item><p><loc href="#feature-profile"><code>#profile</code></loc></p></item>
+<item><p><loc href="#feature-structure"><code>#structure</code></loc></p></item>
+<item><p><loc href="#feature-time-offset"><code>#time-offset</code></loc></p></item>
+<item><p><loc href="#feature-timing"><code>#timing</code></loc></p></item>
+</ulist>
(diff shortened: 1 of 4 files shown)


skynavga:
I wasn't sure whether you @nigelmegitt wanted to include `#audio` as a subset of this new feature designator, but I put it in. just in case Let me know if you think it should come out, or if `#audio-description` or `#audio-speech` should be included.

nigelmegitt:
@skynavga Interesting question about `#audio` - I think the best starting point is to include it, however it does mean that a processor that _only_ supports `#audio-speech` and not `#audio` somehow does not count as supporting `#presentation-audio`.  This is a reflection of the fact that at present there is no easy way to generate reusable audio resources from the text to speech processor, in implementations.

One way around this is to define yet another feature, `#presentation-speech`, which is probably the consistent thing to do. That would not require any layout processing or `#audio` but would require support for `#audio-speech` as well as the other features that the `#presentation-XXX` features require - `#content`, `#core` etc.

spec/ttml2.xml
```diff
@@ -21728,6 +21728,25 @@ transforming the <loc href="#style-attribute-position"><att>tts:position</att></
...
+(2) implements support for audio presentation (rendering) semantics, and
+(3) implements audio presentation semantics for the following features:
+</p>
+<ulist>
+<item><p><loc href="#feature-audio"><code>#audio</code></loc></p></item>
+<item><p><loc href="#feature-content"><code>#content</code></loc></p></item>
+<item><p><loc href="#feature-core"><code>#core</code></loc></p></item>
+<item><p><loc href="#feature-profile"><code>#profile</code></loc></p></item>
+<item><p><loc href="#feature-structure"><code>#structure</code></loc></p></item>
+<item><p><loc href="#feature-time-offset"><code>#time-offset</code></loc></p></item>
```
cconcolato: Why is clock-time not supported?

nigelmegitt: This is a mirror of [`#presentation`](https://www.w3.org/TR/ttml1/#feature-presentation) except without requiring "support for the region and line layout semantics". If you look there, clock-time is not mentioned either. I suspect you aren't talking about clock time as a time base, but about the time expression format.

If you mean [`#time-clock`](https://www.w3.org/TR/ttml1/#feature-time-clock), I don't know why that's not present in `#presentation`, but my guess is that the intent is to require the minimal implementable functionality, and `#time-offset` is the simplest way to express times. Of course it's possible to require processor support for other time expression formats as well by requiring the relevant feature designators.

skynavga:
Tests at w3c/ttml2-tests#249.

PR #1098, opened 2019-05-26
Contextualize 'considered an error' (#1067).

### LINKED ISSUE #1067: Clarify "considered an error".
As discussed on today's call (w3c/ttwg#31) in the context of #1049, wherever a condition is described as "considered an error" without qualification of the processing context in which it is an error, this needs to be addressed.

For example, under tts:extent we have:

> All other syntactically legal values must not be used in this context, and, if used, must be considered an error for the purpose of validation processing and must be ignored for the purpose of presentation processing, in which case the initial value (auto) applies.

which is clear, whereas in other places such as under tts:rubyPosition, we have:

> If specified (on the former ruby text content), then it is considered an error, and must be ignored for the purpose of presentation processing.

There are currently 13 matches for "considered an error" in the spec, soon expected to be 14.

In all cases, it should be clear what the processor behaviour is if the incorrect usage occurs, for example:
* validation processor reports an error
* presentation processor either aborts or continues, and if continuing, what the error handling is, typically some syntax is ignored.

The fix for this should be forward-ported to TTML3 also.

If the fix results in normative changes then new or updated associated tests for validation processors and presentation processors may be needed, and are part of the fix for this issue.

skynavga:
Closes #1067.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -16234,7 +16234,7 @@ is referred to as <emph>chained referential styling</emph>.</p>
 <p>The use of referential styling is restricted to making
 reference to <el>style</el> element descendants of a <el>styling</el>
 element. It is <termref def="defs-considered-an-error">considered an error</termref> to reference a <el>style</el>
-element that is a descendant of a <el>layout</el> element.</p>
+element that is a descendant of a <el>layout</el> element, in which case, the presentation semantics are not defined by this specification.</p>
 <note role="explanation">
 <p>The use of chained referential styling encourages the grouping of style specifications
 into general and specific sets, which further aids in style specification reuse.</p>
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -3351,8 +3351,9 @@ while a <loc href="#semantics-validation-state-validation-exception">validation 
 of type <code>warning</code> is referred to as a
 <termdef id="defs-validation-warning" term="validation warning"><term>validation warning</term></termdef>.</p>
 <p>Unless a <loc href="#terms-higher-level-protocol">higher level protocol</loc> applies,
-(1) if a violation of a prescription or proscription of this specification is detected,
-then a <termref def="defs-validation-error">validation error</termref> exception applies; and
+(1) if a violation of a prescription or proscription of this specification by a <loc href="#terms-document-instance">document instance</loc> is detected,
+then it is <termdef id="defs-considered-an-error" term=""><term>considered an error</term></termdef>, in which case
+a <termref def="defs-validation-error">validation error</termref> exception applies; and
 (2) if a violation of a recommendation or disrecommendation of this specification is detected, then a
 <termref def="defs-validation-warning">validation warning</termref> applies.</p>
 <note role="example"><p>An example of a higher level protocol in this context is a system or end-user defined
@@ -7491,7 +7492,7 @@ the specified attribute value(s) are to be used instead of the font characterist
 must be considered to be equal to the value(s) of the same named font characteristics encoded in the <loc href="#terms-font-resource">font resource</loc>.</p>
 <p>If the computed value of the <code>family</code> attribute (whether explicitly specified or not) matches a
 <loc href="#style-value-generic-family-name">&lt;generic-family-name&gt;</loc> (by case-sensitive comparison),
-then it is considered an error, and the referenced font resource must be ignored for the purpose of presentation processing.</p>
+then it is <termref def="defs-considered-an-error">considered an error</termref> and the referenced font resource must be ignored for the purpose of presentation processing.</p>
 <note role="elaboration">
 <p>The effect of the previous paragraph is that a content author may not override a built-in font resource that corresponds to a generic font family.</p>
 </note>
@@ -9645,7 +9646,7 @@ layout and presentation.</p>
 <p>If the value of this attribute is <code>inlineBlock</code> and this attribute is applied
 to a <loc href="#content-vocabulary-span"><el>span</el></loc> element, then
 <loc href="#terms-inline-block-display">inline block display</loc> semantics apply and the element
-is a candidate for region layout and presentation. Furthermore, it is considered an error if the
+is a candidate for region layout and presentation. Furthermore, it is <termref def="defs-considered-an-error">considered an error</termref> if the
 <code>inlineBlock</code> value is applied to any element which is not a <loc href="#content-vocabulary-span"><el>span</el></loc> element.</p>
 <p>The <att>tts:display</att> style is illustrated by the following example.</p>
 <table id="style-attribute-display-example-1" role="example">
@@ -9893,7 +9894,7 @@ then the specified value is restricted to one of the following:
 (3) two <loc href="#style-value-length">&lt;length&gt;</loc> specifications,
 where these specifications are expressed as non-percentage, definite lengths using pixel units.
 All other syntactically legal values must not be used in this context, and,
-if used, must be considered an error for the purpose of <loc href="#terms-validation-processing">validation processing</loc> and must be ignored
+if used, must be <termref def="defs-considered-an-error">considered an error</termref> and must be ignored
 for the purpose of presentation processing, in which case the initial value (<code>auto</code>) applies.</p>
 <p>If a <att>tts:extent</att> attribute is specified on a <loc href="#layout-vocabulary-region"><el>region</el></loc> element,
 then its computed value determines the <loc href="#terms-width">width</loc> and <loc href="#terms-height">height</loc>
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -16815,7 +16816,7 @@ determine (obtain) the computed style set <emph>CSS</emph> of
 <label>[compute containment scaling](<emph>inout</emph> target rectangle TR, <emph>in</emph> reference rectangle RR)</label>
 <def>
 <olist>
-  <item><p>If the units of target rectangle TR and reference rectangle RR are not the same, then raise an error condition.</p></item>
+  <item><p>It is <termref def="defs-considered-an-error">considered an error</termref> if the units of target rectangle TR and reference rectangle RR are not the same, in which case exit this procedure without modifying the target rectangle.</p></item>
```
nigelmegitt: Not sure if this is a nit or a genuine problem: the definition of "considered an error"  introduced in this pull request specifically mentions document instances, but this algorithm is called from a context where the target rectangle is defined by the intrinsic height and width of the image, which may be from an external resource referenced by the document instance, and not the document instance itself.

Incidentally, it is hard to see how the units of TR and RR could _ever_ be the same, since from the calling concept RR is in the (logical) [document coordinate space](https://www.w3.org/TR/ttml2/#terms-document-coordinate-space) and TR is presumably always a pixel dimension. It's possible that we need a separate issue that requires mapping of TR into the document coordinate space before calling this algorithm, so that this error-checking step effectively becomes a no-op.

This comment applies both here, and on `[compute over scaling]` on line 16856 below.

skynavga: A nit. Readers should be able to read between the lines in this regard (IMO), however, if you insist, I can add something to the effect of _and referenced content_ to the definition.

Regarding _units_, your interpretation is incorrect, since by _units_ here we mean (as we do consistently elsewhere) the enumerated set of units defined by §10.3.22 `<length>`, i.e., _same_ does not mean _same when comparing absolutized relative pixel units_.

nigelmegitt: OK, so resolving the nit would involve changing the definition of "considered an error" so it does not refer only to the document instance but also to any referenced resources that are dereferenced at the time of validity checking. Or saying it's a nit and we don't need to do anything.

In terms of units, there's no way any image resource will sensibly have, say, `rw` or `%` units, unless some processing step upstream has deliberately taken the intrinsic dimensions and re-expressed them in those units to get to TR. If the intent is to constrain document instances to express positions and sizes in pixels if they use embedded images, it's a strange way to achieve that goal.

skynavga: @nigelmegitt this PR does not change existing language with respect to the interpretation of units: it merely changes _raise an error condition_ to a common definition of _considered an error_; if you have a problem with the underlying condition (which I don't see there to be a problem), then file another issue; in the mean time, the present change in the PR is consistent with the scope of its associated issue; also, regarding your _nit_, I prefer making no change to the definition of _considered an error_.

palemieux: @nigelmegitt Couldn't behavior in response to an error, regardless of the error, be defined as: (a) a "validation error" for validation processors and (b) left undefined for presentation processors?

spec/ttml2.xml
```diff
@@ -9645,7 +9646,7 @@ layout and presentation.</p>
 <p>If the value of this attribute is <code>inlineBlock</code> and this attribute is applied
 to a <loc href="#content-vocabulary-span"><el>span</el></loc> element, then
 <loc href="#terms-inline-block-display">inline block display</loc> semantics apply and the element
-is a candidate for region layout and presentation. Furthermore, it is considered an error if the
+is a candidate for region layout and presentation. Furthermore, it is <termref def="defs-considered-an-error">considered an error</termref> if the
```
cconcolato: What is the associated presentation behavior in case of error?

spec/ttml2.xml
```diff
@@ -16189,7 +16190,7 @@ evaluated in the specified order, and that order applies to resolution of the va
 style property in the case that it is specified along multiple reference paths.</p>
 <p>The use of referential styling is restricted to making
 reference to <el>style</el> element descendants of a <el>styling</el>
-element. It is considered an error to reference a <el>style</el>
+element. It is <termref def="defs-considered-an-error">considered an error</termref> to reference a <el>style</el>
```
cconcolato: What should be the associated presentation processing behavior?

spec/ttml2.xml
```diff
@@ -16229,10 +16230,10 @@ ref="semantics-style-inheritance-content"/> below.</p>
 other out-of-line style properties, thus creating a chain of references starting at the
 affected element. When expressed in this manner, the association of style information
 is referred to as <emph>chained referential styling</emph>.</p>
-<p>A loop in a sequence of chained style references must be considered an error.</p>
+<p>A loop in a sequence of chained style references must be <termref def="defs-considered-an-error">considered an error</termref>.</p>
 <p>The use of referential styling is restricted to making
```
cconcolato: The entire paragraph seems to be a repetition of a paragraph in the previous section. I suggest deleting it. Chained referential styling is based on referential styling.

spec/ttml2.xml
```diff
@@ -18213,17 +18214,17 @@ then the value of this term must be constrained to the interval
 [0&hellip;<emph>F-1</emph>], where <emph>F</emph> is the frame rate
 determined by the <att>ttp:frameRate</att> parameter as defined
 by <specref ref="parameter-attribute-frameRate"/>.
-It is considered an error if a <emph>frames</emph> term or
+It is <termref def="defs-considered-an-error">considered an error</termref> if a <emph>frames</emph> term or
```
cconcolato: What is the associated presentation processing behavior?

spec/ttml2.xml
nigelmegitt: Sounds like something to raise as a separate issue @cconcolato ?

spec/ttml2.xml
nigelmegitt: @palemieux it could, but that doesn't seem very helpful. I'd prefer to include a stage to normalise to the same units in the document coordinate space, and if for some reason that can't be done, then raise an error.

palemieux: @nigelmegitt I think it is impractical to specify a behavior for a presentation processor, which naturally assumes that valid documents are presented to it (which is why a validation processor exists).

nigelmegitt: @palemieux Not at all, I see all kinds of specs that describe behaviour in the presence of an unexpected input. We do this elsewhere in TTML too, for example [`tts:extent`](https://www.w3.org/TR/ttml2/#style-attribute-extent) specifies that syntactically legal values that are prohibited on the `tt` element must be ignored. It's true that validation processors are there to catch validation errors, but that doesn't let presentation processors off the hook for doing something sensible, and we should be able to improve interoperability by specifying what "sensible" is in known error conditions.

spec/ttml2.xml
skynavga: @nigelmegitt your comments may be worthwhile, but this PR is not about specifying presentation processing behavior, it is about normalizing the use of text that is intended to correspond with _considered an error_, which the text you refer to is doing; if you want to specify more concrete presentation processor behavior (something I believe is unnecessary), then file a separate issue and let us proceed to accept this PR;

spec/ttml2.xml
palemieux: @nigelmegitt Ok. I do not object to the specification describing handling of errors in the context of both presentation and validation processing (with my understanding that the former will have negligible impact in practice). This is however a pretty significant task given the scope of the spec. Did you have specific changes in mind?

spec/ttml2.xml
skynavga: my above comment is directed at @nigelmegitt; apologies for the wrong nesting

spec/ttml2.xml
nigelmegitt: @palemieux that question is out of scope of this pull request. We have done some trawls before for this kind of thing, and I don't think there's a strong need for a lot of changes right now.

spec/ttml2.xml
nigelmegitt: @skynavga I've opened #1108 as a separate issue about TR vs RR.

spec/ttml2.xml
nigelmegitt: I've always assumed it must be ignored, but we should say so.

spec/ttml2.xml
nigelmegitt: In a sense this is obvious - the only kind of element to which the `inlineBlock` style could be applied that is _not_ a `span` is already an inline block, so it doesn't make any difference.

nigelmegitt:
@skynavga I'd like us to have resolved @cconcolato 's comments before completing my review.

spec/ttml2.xml
skynavga: Saying it must be ignored would constitute a substantive change, and at present this PR and issue are editorial only. The best I can do here is state that presentation semantics are not defined, which I have done at https://github.com/w3c/ttml2/pull/1098/commits/298eb6f96435bb022f5293f49b94dc838787aff2.

spec/ttml2.xml
skynavga: Adding the same information (about presentation semantics not being define) as I did in the previous section. This section was written as a standalone method of styling without referring to the previous section on (non-chained) referential styling. Since there is nothing wrong with the paragraph (other than needing updating here as in the previous section), we should leave it in and not remove it.

spec/ttml2.xml
nigelmegitt: Is there a problem making such a substantive change here?

spec/ttml2.xml
```diff
@@ -16234,7 +16234,7 @@ is referred to as <emph>chained referential styling</emph>.</p>
 <p>The use of referential styling is restricted to making
 reference to <el>style</el> element descendants of a <el>styling</el>
 element. It is <termref def="defs-considered-an-error">considered an error</termref> to reference a <el>style</el>
-element that is a descendant of a <el>layout</el> element.</p>
+element that is a descendant of a <el>layout</el> element, in which case, the presentation semantics are not defined by this specification.</p>
```
nigelmegitt: This, undefined presentation semantics, is the situation we are trying to avoid, so this isn't a change I'd like to make if we can help it. I agree it's a reasonable statement of the status quo though.

spec/ttml2.xml
nigelmegitt: Moved to #1111.

spec/ttml2.xml
nigelmegitt: Moved to #1112.

spec/ttml2.xml
nigelmegitt: @cconcolato says he can live with this as is.

spec/ttml2.xml
nigelmegitt: Moved to #1113.

spec/ttml2.xml
nigelmegitt: This can be revisited in the context of #1112.

cconcolato:
Merged before the 10 days delay because editorial and agreed in meeting

spec/ttml2.xml
skynavga: I'm somewhat concerned that we are starting to litter the spec with text of the form: "X is considered an error, and must/should be ignored for the purpose of presentation processing". Perhaps we should add text under the (new inline) definition of _considered an error_ that says  this and not repeat it throughout the spec (meaning we can remove some of the existing instances of this text).

nigelmegitt: I'm happy with having explicit instructions for presentation processors throughout the spec. If you want to make it easier to understand and appear less verbose you could introduce a stylistic approach where there is a 3 line table:

|Error condition|_testable definition of bad thing_|
|---|---|
|**Validation behaviour**|Issue an error\|warning|
|**Presentation behaviour**|Ignore xxx\|Do yyy|

And add a document convention describing it.

PR #1101, opened 2019-05-26
Emphasize null style semantics in context of ruby container spans (#1100).

### LINKED ISSUE #1100: Emphasize non-applicability of styles to ignored text nodes in ruby container spans.

skynavga:
Closes #1100.


spec/ttml2.xml
```diff
@@ -12386,6 +12386,18 @@ an <termdef id="defs-explicit-ruby-base-container" term=""><term>explicit ruby b
 likewise, an explicitly specified <loc href="#content-vocabulary-span"><el>span</el></loc> element for which the computed value of
 <loc href="#style-attribute-ruby"><att>tts:ruby</att></loc> is <code>textContainer</code> is referred to as
 an <termdef id="defs-explicit-ruby-text-container" term=""><term>explicit ruby text container</term></termdef>.</p>
+<note role="clarification">
+<p>Style properties which might apply to <loc href="#terms-anonymous-span">anonymous span</loc>
+children of a <el>span</el> element that functions as a <termref def="defs-ruby-container">ruby container</termref>,
+<termref def="defs-ruby-base-container">ruby base container</termref>, or <termref def="defs-ruby-text-container">ruby text container</termref>
+have no semantic affect during presentation processing since no significant text node appears as a child of an
```
cconcolato: The important part is indicating how a reader can determine that "no significant text node appears".

skynavga: They do that by reading the normative specification text that has been approved in #1077 in combination with the already existing normative specification text that appears in the paragraph that appears immediately prior to the first Note in §10.2.35: 

>for content that violates some constraint, ...,  if some other content, then ignored (not presented).

The note proposed in this PR merely emphasizes this fact for the reader in the context of `span` elements that serve as ruby containers.

cconcolato: The term 'significant text node' is used here, but never really defined. Looking at the spec I see 2 usages:
a) One related to the [construct anonymous spans] procedure
b) One where it's simply a "text node in a #PCDATA context" (https://www.w3.org/TR/ttml2/#isd-vocabulary-region)

My understanding is that you want to refer to a) (and maybe the terminology in b) should be aligned, but that's a separate issue)

Anyway, #1077 does not talk about 'significant text node' so I think it's hard for the reader to deduce that "no significant text node appears as a child of an anonymous span child of such a span element".

skynavga: Good comment. I will clarify and/or define "significant" in this context.

skynavga:
Although I am generally against repeating the same text when the reader has access to a common declaration, I have, by way of a compromise, added (in https://github.com/w3c/ttml2/pull/1101/commits/9e2fe63da0b92285f70ebc40edccc13475b6bc03) the text "; and see Note" to each of 15 style property's _applies to_ declaration, where "Note" is a link to the note added by this PR to 10.2.35.1.

Note that three style properties that Pierre had proposed constraining cannot be constrained because these properties apply to ruby container spans as well:

- `tts:direction`
- `tts:unicodeBidi`
- `tts:wrapOption`

Therefore, I did not qualify the _applies to_ declaration for these properties.

spec/ttml2.xml
skynavga: @cconcolato added (in https://github.com/w3c/ttml2/pull/1101/commits/4a2c9cb3e1507721b0d988cc0db2a7f8d0e29fdf) links to _significant text node_ definition, as defined by #1107;

skynavga:
@palemieux Re: https://github.com/w3c/ttml2/pull/1101#pullrequestreview-247160031, this is not a legitimate request as applied to editorial PRs, at least as far as our standing protocol applies. If you need more time to review a specific PR that may subject to early merger, then please state your reasons as to why more time is required. Thanks.

palemieux:
@skynavga The standing protocol only applies to those changes that are determined editorial by consensus, not by unilateral action. In any case, this PR as a much larger deficiency, as discussed at https://github.com/w3c/ttml2/pull/1101#pullrequestreview-243028113 .

skynavga:
@palemieux if there is an approval on an issue marked as editorial, and there is no objection to it being marked as editorial, then it falls under the early merger policy; if you would like to review this policy, please raise it to the group in session; regarding this particular PR (#1101), I have no intention to merge it early, whether or not an approval has been given;

palemieux:
@skynavga I think you have two choices: take your time now to gather consensus over proposed resolutions, or faced potentially lengthy delays sooner or later.

skynavga:
@palemieux qui tacit consentire videtur; if I have an approval on an editorial PR, I apply my own judgment as to whether sufficient time has passed and the level of potential controversy on a case by case basis; I believe this is consistent with this WGs operating policy and past history; if you want to propose a change to that policy, then feel free to do so;

skynavga:
@palemieux I will leave your blocker in place for this issue; however, a simple request as a comment to allow this PR to go to full term would have sufficed, and would have been the more collegial approach

nigelmegitt:
It seems like now might be a good time to summarise, because the current state and the original direction may seem surprising (they do to me).

In response to a request to be more specific about the elements to which the style attributes apply, by changing the text in the Applies To column, @skynavga fought hard against doing that on the basis that (summarising from memory, apologies if this is inaccurate) it would add maintenance, and also that it is unnecessary. 

It seems like everyone is agreed that the application of the relevant style attributes to the "ruby containers" (shorthand term) has no effect, so it is academic whether they are included in the "applies to" or not, _except_ that the original issue said that it is helpful to implementers to know from the beginning that some style attributes never apply to those "ruby containers", because then they can be excluded from processing.

This pull request has now got to the point where there _is_ additional text on the "applies to" rows (so the maintenance requirement has not been avoided) and the text it points to says that they "have no semantic affect" during presentation processing, confirming that we are all in agreement on that point.

The decision we need to make _as a group_ is whether we want to express this non-effect as:
1. an informative consequence of other normative text (as in this pull request) or
2. as an additional normative categorisation as in #1071. 

For myself, I'm not sure that there is any significant difference between these approaches, or that there is anything testable that would result from either one or the other approach being taken (which suggests the informative approach is adequate). As always, views welcome. I would like us to come to a conclusion on this fairly rapidly in Thursday's call.

skynavga:
@nigelmegitt thanks for that summary, with which I largely agree; I would merely add the following remarks:

>it is helpful to implementers to know from the beginning that some style attributes never apply to those "ruby containers", because then they can be excluded from processing

- It is important to note that this "helpful" statement takes the form of suggesting an _optimization_ that an implementer may or may not choose to employ. That is to say, an implementation would be fully compliant if it completely ignores such an optimization. This point is at the core of why adding a normative constraint to applies to is untestable.

>This pull request has now got to the point where there is additional text on the "applies to" rows (so the maintenance requirement has not been avoided)

- My preference is to add nothing to the applies to entries; however, I proposed the current text "see also Note" as a compromise in the case that the group feels something is needed. I can live with this note, but I do believe it has a (possibly unwarranted) maintenance overhead; for example, do we need to add the same note to other properties, such as `tts:color`? or future `tts:font*` or `tts:text*` properties?

cconcolato:
@nigelmegitt you seem to indicate that the "applies to" row would be informative? Is that the case?

skynavga:
@cconcolato the "applies to" item of style property definition tables as a whole is normative; however, adding comments of the form "see also Note" to the "applies to" item is by itself informative, since Notes do not carry normative content;

nigelmegitt:
@cconcolato : what @skynavga said - the proposed addition here to the Applies row is informative; the row itself is normative.

skynavga:
@nigelmegitt re: https://github.com/w3c/ttml2/pull/1101#pullrequestreview-249881080, I am already **extremely** uncomfortable with repeating the new comment I have added 15 times, and now you want to expand that text; what I implemented in this recent change is exactly what we agreed to do yesterday; why are you modifying that agreement? if the agreed approach is open for revision, I would prefer going back to my original position, where the applies to entry point directly to (and only to) the generic note;

nigelmegitt:
@skynavga This is the benefit of having proposed text to review - it did not match my understanding of the agreement (because the 2nd-down-the-line note still says "might" in it and I thought we agreed to reword it), and seeing how it looks in reality gives something concrete to respond to.

Rather than threatening to retrench your position, perhaps you could consider my proposal on its merits on behalf of the reader rather than the Editor? I think they are:
1. There's only one note to dereference for each style attribute, so it is easier to navigate.
2. Overall there is one fewer note in the document.
3. The text of the (now duplicated) note is simpler than the removed long note.

I do agree with you that having duplicated text is not ideal, but given the conversation yesterday it seems that @palemieux would not be able to accept the previous proposal of a single generic note and duplicated pointer text.

skynavga:
@nigelmegitt we did not agree to take out the "that might apply" from the generic note yesterday; we agreed to add more concrete binding to that generic note so that "might" is effectively turned into "does"; the new notes do that by explicitly saying that the generic note "holds for this attribute"; again, if you want to insist on incorporating the entire text of the generic note into each of the 15 locations, then I will back out of the compromise agreement, which I am already uncomfortable with; I would rather see NO spec change for this entire suite of related PRs than copy the generic note 15 times.

palemieux:
I agree with @nigelmegitt that the indirection in the current proposal is not ideal. What about the following approach?

In `tts:fontWeight`:

_Applies to: span; see also Note related to the applicability of certain style properties to ruby container spans.-

In _Terms and Definition_:

Add definition for _ruby container spans: ruby container, ruby base container, or ruby text container_

In _10.2.35.1 Special Semantics of Ruby Annotations_:

Simplify the _Note_ to state: _Style properties that affect glyph areas have no semantic affect during presentation processing of a ruby container span since no significant text node appears as a child of a ruby container span. For example, if a tts:fontStyle attribute is specified on a ruby container span, then it has no significant text node in scope to which it may be applied; however, a value specified for tts:fontStyle on such a  ruby container span is still inherited by other (non-anonymous) span element children of the ruby container span._

skynavga:
@palemieux I can agree to that; I will update this PR accordingly in the next few days; [I'm departing for Manila today]

skynavga:
@palemieux see https://github.com/w3c/ttml2/pull/1101/commits/a9025fed6d93a9b6c69b485171f5b65cc5b4fe9c

skynavga:
@palemieux found time to implement https://github.com/w3c/ttml2/pull/1101#issuecomment-503430680; please check

palemieux:
LGTM. Can all the other PRs that _clarify application_ of style properties be merged into this one, so that we can make that the NOTE is present in all relevant style properties?

skynavga:
There is no need to merge them. Their changes do not intersect. All you
need  to do is drop your change requests on the other PRs and I will merge
all into master.

palemieux:
@skynavga I would like to review them together since all these changes were prompted by the same issue.

skynavga:
I am not going to merge them into one PR just to satisfy your review
preferences. You already gave approval to each of the other PRs  during
last week’s call. Feel free to merge them into a local branch if you want
to review them as a whole.

palemieux:
@skynavga What is the effect of applying `tts:direction` on a ruby container span?

palemieux:
@skynavga What is the semantic affect of `tts:color` on a ruby container span?

palemieux:
@skynavga What is the semantic affect of `tts:unicodeBidi` on a ruby container span?

palemieux:
@skynavga What is the semantic affect of `tts:wrapOption` on a ruby container span?

skynavga:
@palemieux for all of your four previous comments the answer is the same, which is "the same semantic affects of applying X to a span that is not a ruby container span"

palemieux:
@skynavga To be more specific by way of an example: what is the impact of applying `tts:color`, which modifies the color of text marks, on a ruby container span, given that the ruby container span does not contain text nodes?

skynavga:
@palemieux I think you mean _specifying_ `tts:color` rather than `applying`; in any case, it means what I said, that it applies to the ruby container span no differently than it would apply to a span that is not a ruby container span, which means it applies to any glyph area in scope (of which there are none since there are no text nodes to which it might apply) and it is inherited by descendent spans;

what other interpretation is possible?

palemieux:
@skynavga in which circumstances would different computed values of tts:color on a given ruby container span yield different rendering results?

skynavga:
@palemieux if you are asking what the semantic affects of changing the value of `tts:color` on a ruby container span are, then the answer is that is affects what `tts:color` is inherited by descendent spans; is there something else you think might be affected?

palemieux:
> then the answer is that it affects what tts:color is inherited by descendent spans;

@skynavga "inherits" is not the same as "applies", which is why the two concepts are treated differently for each style property and in _10.4.4.2 Specified Style Set Processing_.

For instance, `tts:fontWeight` does not *apply* to `div` because, in the following scenario, no value of the `tts:fontWeight` attribute on `div` can yield a different rendering result. In other words, changing the computed value of a style property on an element to which it does not apply can never yield a different results.

```
<div tts:fontWeight=XXX>
<p tts:fontWeight="bold">hello</p>
</div>
```

skynavga:
@palemieux of course it can produce different presentation results:

```
<div tts:fontWeight='bold'>
  <p>hello</p>
</div>
```

produces different results than

```
<div tts:fontWeight='normal'>
  <p>hello</p>
</div>
```

I'm failing to understand what point you are trying to make here or what information you are asking about. Is there an actual problem you are suggesting needs to be addressed?

palemieux:
@skynavga The example you provided above is different than the one I had provided (you removed `tts:fontWeight` on `<p>`).

Ask yourself, can modifying the *computed value* of `tts:fontWeight` on `div` after style resolution processing change rendering? If not, then `tts:fontWeight` does not apply to `div`.

skynavga:
@palemieux Why are you even asking about whether `tts:fontWeight` applies to `div`? Is `div` listed in the _applies to_ line? No. So what point/problem are you attempting to resolve here?

palemieux:
@skynavga Ask yourself, can modifying the computed value of `tts:color` on a ruby container span after style resolution processing ever change rendering? If not, then `tts:color` does not apply to ruby container span.

skynavga:
@palemieux I should also add that you are making a false assumption that presentation semantics is predicated wholly on _applies to_ statements in the property definition tables. It is clear that different visual renderings result from making changes to what style attributes are specified on what elements irregardless of whether the presentation semantics associated with those style attributes is said to _apply to_ an element.

palemieux:
@skynavga I am literally asking you: can modifying the computed value of tts:color on a ruby container span after style resolution processing ever change rendering?

palemieux:
> It is clear that different visual renderings result from making changes to what style attributes are specified on what elements irregardless
> of whether the presentation semantics associated with those style attributes is said to apply to an element.

Yes, but this is the result of inheritance, and regardless of whether the style property applies to the element.

skynavga:
@palemieux and I keep answering the same:  yes it can change rendering, because a change in the computed value results in a change to inherited values

palemieux:
> because a change in the computed value results in a change to inherited values

My question is: can modifying the computed value of `tts:color` on a ruby container span *after style resolution processing* ever change rendering?

skynavga:
@palemieux the computed value cannot change after style resolution processing, which means that inheritance can be performed once and only once; however, animation can cause changes to the _used value_, and, hence, the _actual value_, after style resolution processing

palemieux:
@skynavga Please focus on the question: can modifying the computed value of `tts:color` on a ruby container span after style resolution processing ever change rendering?

skynavga:
@palemieux the question is a non-sequitur: there is no mechanism for changing the computed value of a style property after style resolution processing

skynavga:
@palemieux regarding animation, though we don't explicitly say it in TTML2, the _animation sandwich model_ applies regarding continuous animations via the `animate` element; in particular, as described in [1] continuous animations change the `presentation value`, which, for us, translates to `used value`, so continuous animations do not cause changes to inheritance or the result of the style resolution processing procedure;

[1] https://www.w3.org/TR/2001/REC-smil-animation-20010904/#AnimationSandwichModel

palemieux:
@skynavga My objective is not to litigate resolution processing and/or animation processing, but instead develop a criteria to determine whether a style property applies to an element.

Rephrasing my question to make it palatable to you: can modifying the actual value of `tts:color` on a ruby container span ever change rendering?

skynavga:
@palemieux re:

>Rephrasing my question to make it palatable to you: can modifying the actual value of tts:color on a ruby container span ever change rendering?

No.

However, re:

>develop a criteria to determine whether a style property applies to an element

I do not consider it necessary to develop, articulate, or otherwise publish such a criteria, even if we could all agree to what that criteria might be, and I expect we would never reach such an agreement in the first place. If you want to develop such a criteria for your own thinking, that is a different matter, but I do not consider doing such a thing to be part or parcel of resolving any of the current outstanding issues on TTML2.

palemieux:
> I do not consider doing such a thing to be part or parcel of resolving any of the current outstanding issues on TTML2.

Well, I do, since the issue is specifically about whether some style properties apply to ruby span containers, and you have not succeeded in developing a consistent approach to making this determination.

skynavga:
@palemieux actually, I need to revisit my answer of

>Rephrasing my question to make it palatable to you: can modifying the actual value of tts:color on a ruby container span ever change rendering?
>>
>>No.

In particular, I need to qualify this answer with respect to _actual value_ and continuous animation. Without considering animation, there is no mechanism for changing _compute value_, _used value_, or _actual value_ after style resolution processing. However, if we consider animation, then both _used value_, and hence, _actual value_, may be changed after style resolution processing.

But nothing in this respect has any bearing on _applies to_ or vice-versa.

skynavga:
@palemieux re:

>>I do not consider doing such a thing to be part or parcel of resolving any of the current outstanding issues on TTML2.
>
>Well, I do, since the issue is specifically about whether some style properties apply to ruby span containers, and you have not succeeded in developing a consistent approach to making this determination.

Neither the filed issues nor the PRs that address these issues call for an algorithm for determining whether a style property _applies to_ an _element_ in the sense that this language is used in the TTML specifications, or, in XSL-FO, or in CSS for that matter.

If it is your intent to introduce this as a new requirement for our specification, then file a different issue. However, I can say ahead of time that I will oppose such a requirement even more strongly than I opposed your initial description of #1043.

skynavga:
@palemieux I want to add that we already have generic language in the spec (that you reviewed and agreed to) that is used to determine which elements are pruned during intermediate document construction; in a strong sense, this language is analogous to your desire to determine which style properties might be _effectively pruned_ for the purpose of optimizing a presentation processor;

the language I am referring to is step 3(a) of [construct intermediate document], which employs the term _presentation related element_, which is defined in general terms as

>Any element which (1) contains or expresses presentable content or (2) affects the presentation of content. 

in particular, take note of the usage _affects the presentation_, which we do not further specify or define in the specification; when we adopted this language, we were intentionally vague about what might _affect the presentation_;

you now seem to be asking us to back out of that former agreement by fully defining an algorithm for determining what is meant by _applies to_ as it affects presentation; you can't have it both ways

nigelmegitt:
### Going round in circles

@palemieux I'm not seeing much sign that this recent discussion since https://github.com/w3c/ttml2/pull/1101#issuecomment-504633371 has moved things forward. Note that we discussed and agreed the meaning of "applies to" in #1089.

@skynavga this may be an invalid assumption on my part, but it seems pretty obvious where @palemieux was heading with his questions - I appreciate your precision in answering exactly the questions asked, but it looks like you're trying to avoid answering the core question.

Nevertheless, **both of you**, does this really help? Aren't we rehashing a discussion already had? 

### What we all do seem to agree on

It's clear that the neither the specified, nor the computed value, nor the actual value of `tts:color` makes any difference to the presentation of an element that contains no character content, regardless of whether it is a ruby container, a `div`, or an ordinary `span`, but that the specified value can affect the presentation of descendant elements that do contain character content, via inheritance. 

Further, it is also clear that no test can reveal the difference in conformance of a document given either 1) a normative 'applies to' constraint excluding ruby containers or 2) (a reference to) an informative note that reminds the reader that even if `tts:color` is specified on a ruby container, it won't make any presentational difference. 

### Can we wrap this up please?

I had thought we had agreed to proceed with the informative note approach. @palemieux if I didn't understand your position correctly there, then please be really specific about what testable effect you want to generate by changing the normative text in the "applies to" rows. Otherwise can we close this with the consensus we seemed to have, and move on from this please?

### Afterthought

#### Presentation behaviour of invalid documents?

By the way, one aspect we have not considered until now is the presentation effect on an _invalid_ ruby container that erroneously does contain character content children. Since we don't define the presentation behaviour in that error state, I would argue that it is academic what the application status is of style attributes that don't have any effect on valid ruby containers. However, if we were to normatively exclude them on the "applies to" line then there could be some kind of hybrid state where a presentation processor attempts to present the text children (in an undefined way) but is required to honour the non-application of excepted style attributes. This is all just going too far into the weeds in my opinion: I mention it only because I can't envisage any other scenario in which it might matter.

skynavga:
@nigelmegitt re: your first remarks, I too was surprised that @palemieux appears to be reopening a delicate agreement that I thought we had already resolved, including his "LGTM" in https://github.com/w3c/ttml2/pull/1101#issuecomment-503828613 above;

re: your

>one aspect we have not considered until now is the presentation effect on an invalid ruby container that erroneously does contain character content children. Since we don't define the presentation behaviour in that error state

actually, we do have normative text that covers this scenario in 10.2.35 (emphasis added)

>A validating content processor must treat a violation of any of the above constraints as an error. For the purpose of presentation processing, the violation of any of these constraints should result in fallback (inline) presentation of ruby text annotations, and, for content that violates some constraint, ..., **or, if some other content, then ignored (not presented)**.

nigelmegitt:
Ah, thanks for that pointer @skynavga

palemieux:
> I too was surprised that @palemieux appears to be reopening a delicate agreement that I thought we had already resolved, including his "LGTM" in #1101 (comment) above;

@skynavga I in fact remain happy with the prose. I also believe it should be extended to cover the following style properties:

* `tts:direction` since _when applied to a span element (or anonymous span), the computed value of ``tts:direction`, in combination with the computed value of the `tts:unicodeBidi` style property, determines the directionality of a bidirectional embedding or override as specified by [UAX9], §4.3, Higher Level Protocol HL3._ In other words, `tts:direction` does not seem to have any effect unless the element has children text nodes.

* `tts:unicodeBidi` since it cannot apply across the children elements of a ruby container spans, which are rendered independently, e.g. how can the unicode bidi algorithm be applied on ruby text containers in double-sided ruby.

* `tts:wrapOption` since the ruby container span has no children text nodes where wrapping can occur. `tts:wrapOption` of course inherits to any children containers, where wrapping can occur -- but that is inheritance and not application.

* `tts:color` since, per https://github.com/w3c/ttml2/pull/1101#issuecomment-504715041, there is no children text nodes that `tts:color` can affect in a ruby container span

Looking forward to better understanding the analysis behind your comment at https://github.com/w3c/ttml2/pull/1101#issuecomment-498041055

palemieux:
> then please be really specific about what testable effect 

@nigelmegitt For instance, it do not think the unicode bidi algorithm should apply across the children spans of the ruby container, since each is laid out as an independent unit. Do you agree? If so, I would say that `tts:direction` and `tts:unicodeBidi` do not apply to the ruby container.

```
<span tts:ruby="container" tts:rubyAlign="spaceAround" tts:direction="rtl" tts:unicodeBidi="embed">
    <span tts:ruby="baseContainer">
      <span tts:ruby="base">東南</span>
    </span>
    <span tts:ruby="textContainer" tts:rubyPosition="before">
      <span tts:ruby="text">نشاط التدويل، W3C</span>
    </span>
    <span tts:ruby="textContainer" tts:rubyPosition="after">
      <span tts:ruby="text">1234</span>
    </span>
  </span>
```

skynavga:
@palemieux I take it you have never implemented the Unicode Bidi algorithm, for if you had, you would understand that nested embedding levels translate to changes in resolved bidi levels; e.g., if one has

```
<p tts:unicodeBidi='embed' tts:direction='ltr'>
  <span tts:ruby='container' tts:unicodeBidi='embed' tts:direction='rtl'>
    <span tts:ruby='baseContainer' tts:unicodeBidi='embed' tts:direction='ltr'>
      <span tts:ruby='base' tts:unicodeBidi='embed' tts:direction='rtl'>A</span>
      <span tts:ruby='base' tts:unicodeBidi='embed' tts:direction='ltr'>B</span>
    </span>
    <span tts:ruby='textContainer' tts:unicodeBidi='embed' tts:direction='ltr'>
      <span tts:ruby='text' tts:unicodeBidi='embed' tts:direction='rtl'>a</span>
      <span tts:ruby='text' tts:unicodeBidi='embed' tts:direction='ltr'>b</span>
    </span>
  </span>
</p>
```

will produce resolved bidi levels of

```
ABab
3434
```

and not

```
ABab
1010
```

as your logic would produce.

For ``tts:wrapOption`` the following

```
<p>
  <span tts:ruby='container'>
    <span tts:ruby='baseContainer' tts:wrapOption='noWrap'>
      <span tts:ruby='base' tts:wrapOption='wrap'>ABC</span>
      <span tts:ruby='base' tts:wrapOption='wrap'>DEF</span>
    </span>
    <span tts:ruby='textContainer' tts:wrapOption='wrap'>
      <span tts:ruby='text' tts:wrapOption='noWrap'>abc</span>
      <span tts:ruby='text' tts:wrapOption='noWrap'>def</span>
    </span>
  </span>
</p>
```

allows wrapping within `ABC`, `DEF` and between `abc` and `def`, but does not allow wrapping between `ABC` and `DEF` or within `abc` or `def`.

As for `tts:color`, it applies to `<span tts:ruby='textContainer'/>` in the same way that it applies to `<span/>`.

nigelmegitt:
> I take it you have never implemented the Unicode Bidi algorithm, for if you had, you would understand that ...

@skynavga this against-the-person phrasing falls short of our standards of respect and professional conduct. Since you are providing the benefit of your own implementation experience, a better phrasing would be "I have learned from implementing the Unicode Bidi algorithm that ..."

palemieux:
> allows wrapping within ABC, DEF and between abc and def, but does not allow wrapping between ABC and DEF or within abc or def.

@skynavga Ok, `tts:wrapOption` can have an effect on a ruby base container. I do not see how `tts:wrapOption` can have an effect on ruby containers (since wrapping does not happen between ruby base and ruby text) and on ruby text containers (since each of the child of a ruby text container is associated with a different ruby base and wrapping does not happen between them).

palemieux:
> you would understand that nested embedding levels translate to changes in resolved bidi levels; e.g.,

@skynavga Does TTPE render the following two `spans` differently?

```
				<span tts:direction="rtl" tts:unicodeBidi="embed">
				<span>1234 </span><span>نشاط التدويل، W3C</span>
				</span>
```

```
				<span tts:direction="rtl">
				<span>1234 </span><span>نشاط التدويل، W3C</span>
				</span>
```

palemieux:
> As for tts:color, it applies to <span tts:ruby='textContainer'/> in the same way that it applies to <span/>.

Since `<span tts:ruby='textContainer'/>` has no children text nodes, `tts:color` has no effect, as discussed above.

skynavga:

skynavga:
Having left a comment above, I see that I cannot mark this PR for "changes requested" since I was the one who created this PR in the first place. Since I cannot now serve as a proper reviewer of this PR, I am closing this PR once again. If @palemieux or @nigelmegitt would like to create a new PR based on the current changes proposed in this PR so that I can serve as a proper reviewer, then you may do so.

nigelmegitt:
I understood the interop problem is with the three direction-related attributes we agreed to move to a separate issue, `wrapOption`, `direction` and `unicodeBidi`. @palemieux is that correct, or is there an interop problem with the remaining attributes whose text is informatively modified by this pull request?

skynavga:
@nigelmegitt I object to moving forward with this PR for two reasons, one technical, the other procedural: (1) it has not been demonstrated that there exists an interoperability issue, a specification ambiguity, or a technical reason for disclaiming the application of 15+1 (now 16) properties to the `span` element type in certain contingent document specific contexts, and, since such demonstration is not forthcoming, it is impossible to present a technical objection to 16 separate hypothetical author-dependent use case scenarios; (2) as the original submitter of this PR, I am unable to serve as a reviewer, including, for example, exercising my prerogative to submit a review that requests changes to what is now, in my opinion, an unfriendly amendment to my initial PR, and, which has, in effect, been adopted by @palemieux while excluding (under protest) the original submitter of this PR.

nigelmegitt:
@skynavga this does not seem like a constructive way for us to make progress.

> (1) it has not been demonstrated that there exists an interoperability issue, a specification ambiguity, or a technical reason for disclaiming the application of 15+1 (now 16) properties to the span element type in certain contingent document specific contexts

I do not understand this, from a consistency perspective: you were apparently happy with 15, but the 16th somehow breaks the pull request? You have not argued that `tts:color` might apply to ruby containers in a way that, say, `tts:fontFamily` does not. Both affect glyph areas only. The technical reason has been discussed ad nauseam. The changes made here are informative only.

> (2) as the original submitter of this PR, I am unable to serve as a reviewer, including, for example, exercising my prerogative to submit a review that requests changes to what is now, in my opinion, an unfriendly amendment to my initial PR, and, which has, in effect, been forcibly taken over by @palemieux while excluding (under protest) the position and opinion of the original submitter of this PR.

What changes are you requesting? Why is increasing the consistency by bringing `tts:color` in an "unfriendly" amendment?

skynavga:
@nigelmegitt if it is true that

>The changes made here are informative only.

then you (and @palemieux) must agree that they are not necessary changes; do you and @palemieux agree?

then, if they are not necessary, you must agree that there are other ways this same strictly editorial point can be made; do you agree? for example, we can return to the previous state of this PR which added a general, informative note that covers all cases that present these circumstances, in particular, all 15 properties that I had previously agreed to add an explicit note to along with yet another property that @palemieux seems to have left off his initial hit list;

I have never been comfortable repeating the same information in 15 places where one general note covers all cases, and now that @palemieux has opened the door once again to what was a previous tenuous compromise, am no longer willing to proceed with that compromise and prefer to return to the single general note. My technical (and editorial) reason for this is that adding language that disclaims applicability abuses the other agreement we had reached, which is that the _applies to_ information follows the CSS convention, which is to say, it applies to elements, or, more precisely, to element types. Since `span` is an element type, then all 15+1 properties apply to `span`, as already is indicated by the specification text. And that is the end of the story.

nigelmegitt:
> CSS convention, which is to say, it applies to elements, or, more precisely, to element types

@skynavga this is not the case. CSS does not list only element types in its "applies to" entries, in some places it uses other ways to qualify the kind of object to which the property applies, which may be qualified.

I still have no idea why you consider it to be a problem to add a 16th note where 15 already exist, which by the way are not the same because their context differs. What is special about `tts:color` that means it should not also get the note that is present on other attributes that have an effect only on glyph areas?

skynavga:
@nigelmegitt we had already agreed (in #1089) that "the definition of _applies to_ specified by [CSS2] §1.4.2.3 applies to this specification", and that definition is

>This part lists the elements to which the property applies. All elements are considered to have all properties, but some properties have no rendering effect on some types of elements. For example, the 'clear' property only affects block-level elements.

This language does not qualify elements (by which it means element types). It either applies to an element or it does not. Furthermore, all of the 15+1 properties being discussed here clearly have rendering effects on elements of type `span`. Therefore, all of the 15+1 properties apply to the element type `span`.

Adding disclaimers to _applies to_ that attempt to carve out specific contingent documents contexts where a given `span` element happens to have nothing to which a property applies is an abuse of the definition of _applies to_ and can only lead to misunderstandings and confusion on the part of readers. Do the apply to the element type `span` or not? The answer is it depends on the particular instance of the `span` element type. CSS2 most certainly does not attempt to articulate such disclaimers in the _applies to_ entries.

If you want to add notes to 15+1 property specifications in the bodies of their definitions without attempting to qualify the _applies to_ entries in the style property definition tables, then I might willing to consider this, even though the single note originally proposed to be added to §10.2.35.1 serves the same purpose without resorting to wholesale copying, an entirely ill-advised course of action in my opinion and contrary to our past efforts to eliminate duplication.

nigelmegitt:
@skynavga re your assertion that CSS never qualifies elements to which a property _applies_, this is easily disprovable as mentioned way back at https://github.com/w3c/ttml2/pull/1101#issuecomment-501792296. See https://drafts.csswg.org/css-writing-modes-4/#property-index for just one of many examples:

* "applies to" in one case specifies "non-replaced inline elements" which is a qualified subset of elements, since some `span` elements are, and others are not, non-replaced inline elements, depending (in TTML2) on their contents and the value of `tts:display` - it is also a qualified subset of elements in HTML.
* in another case: "All elements except table row groups, table column groups, table rows, table columns, ruby base container, ruby annotation container " - since ruby base container is an "element" in terms of CSS (and HTML) but an attribute-qualified element in TTML2, the equivalent is to do exactly what is proposed here (actually it would be to do so normatively, but this is a start).

nigelmegitt:
As Chair, I want to draw this to a close. As I see it, we have consensus to merge as is, with one objection for which I have requested an explanation multiple times and so far never been satisfied. We have spent more than enough time on this and need to be able to address other open issues and pull requests.

For that reason I'm declaring this to be good enough to do the job needed to satisfy issue #1100, and requesting the Editors to merge this in the near future please.

I note that there is one objection, despite my reservations about how well formed it is, and I am deciding to carry that forward. If it remains in place when we request the next transition I will explain the situation to the Director as needed.

skynavga:
@nigelmegitt the chair does not own this PR; I am the one who created it, and it is my prerogative to close it, which I will do again now (yet again); I have repeatedly asked that you or @palemieux create another PR that affords me the opportunity to serve as a reviewer, since I cannot do so on this PR (due to having created it); it should be straightforward for you to create a new PR using the existing branch associated with this PR; you can reference this PR's conversation thread from the new PR as well, so you won't lose any information or context;

skynavga:
@nigelmegitt I should add that it appears to me that, by continuing to reopen this PR and not creating a new PR, you are attempting to circumvent my ability to review the changes now represented by this PR; that appears to completely circumvent our standing protocol and agreements that afford members the opportunity to review PRs they didn't author; even if you do create a new PR using the changes from this PR, and I post a "changes requested" review, you can still go through a process of dismissing my review without it being satisfied; however, I would expect that is an action you cannot take unilaterally as chair, but must have the WG collectively decide to dismiss a review.

skynavga:
@nigelmegitt regarding https://github.com/w3c/ttml2/pull/1101#issuecomment-509266297, I explicitly cite CSS2 in https://github.com/w3c/ttml2/pull/1101#issuecomment-508819799; as far as I know, no subsequent version of CSS or a CSS module has redefined _applies to_ as defined in CSS2; for example, CSS2.2, currently in FPWD status retains the same language (https://www.w3.org/TR/CSS22/about.html#applies-to);

regarding your second bullet "in another case", could you provide a link to the text you cite?

skynavga:
@nigelmegitt regarding your claim in https://github.com/w3c/ttml2/pull/1101#issuecomment-509269002

>I have requested an explanation multiple times and so far never been satisfied

Did I not ask in https://github.com/w3c/ttml2/pull/1101#pullrequestreview-258177330

>For example, what precisely is the interoperability problem with tts:color that these changes propose to solve? The same question applies for the other 15 properties. What is the interoperability problem?

I have not seen an answer to this question. How can I provide a technical counter-argument when the proponent of the change hasn't provided a technical argument for the change in the first place?

I have not seen any response from @palemieux to your question in https://github.com/w3c/ttml2/pull/1101#issuecomment-508688056:

> @palemieux is that correct, or is there an interop problem with the remaining attributes whose text is informatively modified by this pull request?

Is there an interop problem with these 15+1 properties or not? I would like to see an answer from @palemieux either way: yes or no. If yes, then what is the interop problem? If no, then what is the technical argument in favor of the change (to these 15+1 properties)?

Unless I have answers to these questions, then I can begin to write a technical counter argument, and it is inconsistent of you to demand that I do so.

nigelmegitt:
@skynavga responses to multiple comments:

> you are attempting to circumvent my ability to review the changes now represented by this PR

On the contrary, I explicitly noted your objection. The ability to review and object are not well supported by GitHub pull requests in the case that a pull request is modified by someone other than the original author, I agree, but you still had the opportunity to object, and took it. I asked you in https://github.com/w3c/ttml2/pull/1101#issuecomment-508794196 what changes you wanted to make, and you did not answer that question.

> regarding your second bullet "in another case", could you provide a link to the text you cite?

It's the same link, the two cases I mentioned were different entries in the table. The first case was for text-combine-upright, the second was for writing-mode.

Regarding interop, this was discussed on Thursday and it was my understanding from that call that the interop issue primarily affects the attributes that we agreed to move to a separate issue. Is that not correct @palemieux ?

skynavga:
@palemieux please do not reopen this PR; I opened this PR, therefore, I own this PR, and I am closing it; feel free to open another PR which makes use of the same branch as this PR;

skynavga:
@palemieux note that you can continue to add or edit comments on this PR even while closed;

palemieux:
> Regarding interop, this was discussed on Thursday and it was my understanding from that call that the interop issue
> primarily affects the attributes that we agreed to move to a separate issue.

This pull request is intended to address issue #1043, which was filed following a concrete interop issue (https://github.com/w3c/ttml2/issues/1043#issuecomment-488462399): does `tts:textDecoration` apply to ruby span containers? The same question applied to style properties beyond `tts:textDecoration`.

I would personally prefer tackling `tts:direction`, `tts:unicodeBidi` and `tts:wrapOption` today, in addition to those style properties already addressed by this pull request, but do not object waiting until later (as discussed last Thursday) since:

* these three properties are unlikely to be used with ruby span containers in practice (at least in the short-term)
* communication with CSS WG might be useful
* interaction between a style property like `tts:wrapOption` and ruby containers might require careful consideration

palemieux:
> please do not reopen this PR; I opened this PR, therefore, I own this PR, and I am closing it;

@skynavga This PR is no more yours than mine. It belongs to the group.

skynavga:
@palemieux there is no longer an interop issue with `tts:textDecoration` since the merger of PR #1077; are you saying there is another interop issue that is outstanding? what interop issue remains unaddressed with any of the 15+1 properties modified by this PR?

skynavga:
@palemieux it is silly of you to keep opening this PR when I am just going to keep closing it; please discuss with the chair before reopening again;

skynavga:
After reviewing the state of the changes proposed in this PR, including those made by @palemieux, I have determined I can accept an additional (informative) qualification of the _applies to_ entry of the 16 style property definition tables modified by this PR provided that the redundant paraphrase of the referenced note is removed in each case. Accordingly, I have updated this PR to include only the additional text "; see also Note" in each of the 16 modifies property definition tables. I believe the removal of the redundant paraphrase of the Note "related to the applicability of certain style properties to ruby container spans" does not reduce the utility of the reference to the Note since a reader can merely follow the link to determine that it is indeed "related to the applicability of certain style properties to ruby container spans".

Accordingly I will re-open this PR for further review by the WG. I also want to note for the record that @palemieux has not provided any substance to back up his claims that any interoperability problem remains with respect to these 16 properties. Nevertheless, I recognize the potential utility in drawing the reader's attention to the Note that explains why there is no interoperability problem.

spec/ttml2.xml
skynavga: Marking as resolved since definition of significant text node has already been added (in master branch).

PR #1106, opened 2019-05-31
Specify that font selection strategy is used by line height algorithm (#1102).

### LINKED ISSUE #1102: Use of font selection strategy in line height algorithm.
Currently the `tts:fontSelectionStrategy` attribute is not listed in the algorithm for `normal` in `tts:lineHeight` but since that algorithm (step 4) is based on actual fonts chosen, it should be.

### Algorithm should take include `tts:fontSelectionStrategy`

The algorithm for computing the used value at https://www.w3.org/TR/ttml2/#style-attribute-lineHeight lists in step 4:

> the computed values of the `tts:fontStyle`, `tts:fontWeight`, and `tts:fontVariant` style properties of P. 

This should additionally include `tts:fontSelectionStrategy`.

### Consistency of referencing back to the algorithm.

[`tts:fontFamily`](https://www.w3.org/TR/ttml2/#style-attribute-fontFamily) and [`tts:fontWeight`](https://www.w3.org/TR/ttml2/#style-attribute-fontWeight) and
[`tts:fontVariant`](https://www.w3.org/TR/ttml2/#style-attribute-fontVariant) all include text like:

> In addition to applying to span, the computed value of this property on a `p` element is used to resolve the normal value of the style property associated with the `tts:lineHeight` attribute as specified in 10.2.27 `tts:lineHeight`.

This should be added to [`tts:fontSelectionStrategy`](https://www.w3.org/TR/ttml2/#style-attribute-fontSelectionStrategy)

skynavga:
Closes #1102.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -10057,7 +10057,7 @@ by content flowed into a region.</p>
 further described in <specref ref="style-value-generic-family-name"/> below.</p>
 </note>
 <p>In addition to applying to <loc href="#content-vocabulary-span"><el>span</el></loc>, the computed value of this property
-on a <loc href="#content-vocabulary-span"><el>p</el></loc> element is used to resolve the <code>normal</code> value of the
+on a <loc href="#content-vocabulary-p"><el>p</el></loc> element is used to resolve the <code>normal</code> value of the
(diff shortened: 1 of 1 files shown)


skynavga:
Tests at w3c/ttml2-tests#252.

PR #1109, opened 2019-06-03
Remove error condition from scaling procedures (#1108).

### LINKED ISSUE #1108: Semantics of units employed by [compute containment|cover scaling].
The requirement for success in the first step of `[compute containment scaling]` and `[compute over scaling]` is that the units of TR and RR are the same, but this is rarely likely to be the case.

Propose replacing with a step to re-express image dimensions in the document coordinate space so that the resulting units are always the same.

> In terms of units, there's no way any image resource will sensibly have, say, `rw` or `%` units, unless some processing step upstream has deliberately taken the intrinsic dimensions and re-expressed them in those units to get to TR. If the intent is to constrain document instances to express positions and sizes in pixels if they use embedded images, it's a strange way to achieve that goal.

_Originally posted by @nigelmegitt in https://github.com/w3c/ttml2/pull/1098_

skynavga:
Closes #1108.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -15234,7 +15234,7 @@ in a cartesian space or a distance between two points in a cartesian space.</p>
 scalar
   : <loc href="#style-value-number">&lt;number&gt;</loc> units
 
-units
+<phrase><termdef id="defs-units" term=""><term>units</term></termdef></phrase>
   : "px"
   | "em"
   | "c"                                     // abbreviation of "cell"
@@ -15245,8 +15245,7 @@ units
 </tr>
 </tbody>
 </table>
-<p>It is an error to omit the <termdef id="defs-units" term=""><term><phrase><code>units</code></phrase></term></termdef>
-<loc href="#terms-component">component</loc> of a scalar length value.</p>
+<p>It is an error to omit the units <loc href="#terms-component">component</loc> of a scalar length value.</p>
 <p>The unit <code>px</code> (pixel) corresponds to a <termref def="defs-logical-pixel">logical pixel</termref> of
 the <loc href="#terms-document-coordinate-space">document coordinate space</loc>,
 as defined by <specref ref="root-container-region-semantics-coordinate-space"/>; however,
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -15245,7 +15245,8 @@ units
 </tr>
 </tbody>
 </table>
-<p>It is an error to omit the units <loc href="#terms-component">component</loc> of a scalar length value.</p>
+<p>It is an error to omit the <termdef id="defs-units" term=""><term><phrase><code>units</code></phrase></term></termdef>
```
nigelmegitt: Putting a term definition in an error declaration feels like a bad idea. Looking a bit deeper at this, the whole statement is unnecessary and arguably meaningless because the syntax representation does not show `units` as being optional; by definition a scalar length value contains units. It's the equivalent of saying "a thing with units that is simultaneously without units is an error"!

Proposal:

1. put the term definition around the use of `units` in the syntax representation of `scalar`? (i.e. `scalar: &lt;number&gt; <termdef id="defs-units" term=""><term><phrase><code>units</code></phrase></term></termdef>`)
2. Restate the error declaration as a note: "No length expression without units is defined; relative lengths can be expressed using percentages only."

skynavga: I just addressed your first in https://github.com/w3c/ttml2/pull/1109/commits/1960e3271820d5ced573b37af96e381b166f0293. However, your second point is superfluous to the present issue, and introduces something new (the issue of relative lengths). FYI, the reason that we added this text in the first place (some time ago) has nothing to do with relative units; rather, we had entertained dropping the requirement to have explicit units and defaulting to logical pixels (i.e., `px`) as is done with SVG; however, we decided against that path and decided instead to add this statement (which I agree could be viewed as redundant) as being an error to omit units. To conclude, I think we should not change this language at this time.

nigelmegitt: I've opened #1116 about the error statement.

skynavga:
Marking this change as untestable since unit conversion in this context is entirely implementation dependent and not visible to either application content or the end user.

PR #1110, opened 2019-06-05
Correct and clarify #extent-root-version-2 (#1027).

### LINKED ISSUE #1027: Clarify #extent-root-version-2.
The `#extent-root-version-2` feature designator cannot make use of `#length-version-2` because only pixel lengths are supported on the root element.

Furthermore, use of `#extent-auto-version-2` on the root element is no different from TTML1, syntactially speaking. However, there is a subtle difference semantically, which maps `auto` to `contain`, and `contain` on root means using the rules of Appendix H, which, in this case, uses arbitrary document processing context rules as long as they respect the resolved SAR.

Note also that `#extent-root-version-2` does not, by itself, sanction the explicit syntactic use of `contain`.

Finally, for testing purposes, `#extent-auto-version-2` on the root element must make use of explicit DAR and/or PAR to make it potentially different semantically from TTML1.

skynavga:
Closes #1027.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -20995,7 +20995,7 @@ when applied to the <loc href="#document-structure-vocabulary-tt"><el>tt</el></l
 <ulist>
 <item><p><loc href="#feature-extent-root"><code>#extent-root</code></loc></p></item>
 <item><p><loc href="#feature-extent-auto-version-2"><code>#extent-auto-version-2</code></loc></p></item>
-<item><p><loc href="#feature-extent-length-version-2"><code>#extent-length-version-2</code></loc></p></item>
+<item><p><loc href="#feature-extent-contain"><code>#extent-contain</code></loc></p></item>
 </ulist>
 <p>The <code>#extent-root-version-2</code> feature is a syntactic and semantic extension of the
 <loc href="#feature-extent-root"><code>#extent-root</code></loc> feature.</p>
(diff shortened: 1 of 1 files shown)


skynavga:
This PR entails a substantive change only to the extent that it redefines the meaning of the `#extent-root-version-2` feature designation. It does not add or change any underlying feature as such.

spec/ttml2.xml
```diff
@@ -20995,15 +20995,20 @@ when applied to the <loc href="#document-structure-vocabulary-tt"><el>tt</el></l
...
 <p>The <code>#extent-root-version-2</code> feature is a syntactic and semantic extension of the
 <loc href="#feature-extent-root"><code>#extent-root</code></loc> feature.</p>
 <p>It is an error for a content profile to require or permit use of the <code>#extent-root-version-2</code> feature
 but prohibit use of any of the following features:
 <loc href="#feature-extent-root"><code>#extent-root</code></loc>,
 <loc href="#feature-extent-auto-version-2"><code>#extent-auto-version-2</code></loc>, or
-<loc href="#feature-extent-length-version-2"><code>#extent-length-version-2</code></loc>.</p>
+<loc href="#feature-extent-contain"><code>#extent-contain</code></loc>.</p>
+<note role="clarification">
+<p>Support for the <code>#extent-root-version-2</code> feature is intended to imply support for the
```
nigelmegitt: This feels like we're building ourselves a problem for the future. Why not create `#root-extent-contain` to imply support for `contain` specifically on `tts:extent` on `tt` and be explicit about it?

skynavga: For one thing, we don't have any feature designations starting with `#root-` (but we could name it `#extent-root-contain`, but also because adding a new feature designation is a significantly more extensive change than tweaking the existing designation of `#extent-root-version-1`. As it is, I think we have far too many feature designations related to `tts:extent`, and adding yet another (when the alternative of tweaking a current designation) exacerbates the situation further.

nigelmegitt: The 'root' (sorry!) cause is that we use `tts:extent` in two different contexts and apply very different constraints to usage depending on where it is used. Unless we want to change that, e.g. deprecating its use on `tt` and replacing it with a new attribute, we should at least recognise that implementations may well support different functionality for its use on `tt` or `region` and allow some precision in the feature designators.

I really don't like having context-dependent feature inclusion without any syntactical help. Another option could be to allow profile documents to include an extra attribute on the feature elements specifying the context of support, but that's even worse in my opinion than defining the features with finer granularity because it introduces additional complexity to the profile processing.

nigelmegitt: `#extent-root-contain` would work for me.

skynavga: Since we don't have an implementer or profile author driving the inclusion of a new designation, I will defer adding one and go with the proposed (and approved) change that tweaks the current language.

skynavga:
Tests at w3c/ttml2-tests#253.

PR #1114, opened 2019-06-07
Remove redundant feature dependency from #ruby-full (#1025).

### LINKED ISSUE #1025: Remove redundant dependency on #rubyAlign-withBase.
For `#ruby-full`, remove dependency on `#rubyAlign-withBase` since it is already included in `#rubyAlign`.

skynavga:
Closes #1025.


nigelmegitt:
No test can distinguish between the before and after cases here, so I concur this is editorial.

PR #1123, opened 2019-06-25
Add IMSC profile designators (#1034).

### LINKED ISSUE #1034: Add IMSC profile designators to Table 5-2.
At present, IMSC's use of profile designators in the TT Profile Namespace is technically out of compliance with TTML2 due to the following language:

**TTML2 (1e) §5.2.3**

>All profile designators which have the TT Profile Namespace as a prefix but are otherwise not listed in Table 5-2 – Profiles are reserved for future standardization, and must not appear in a conformant document instance.

To fix this, we need to add entries to Table 5-2 for IMSC's text and image profiles (both 1.0 and 1.1).

skynavga:
Closes #1034.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -25814,16 +25814,6 @@ European Broadcasting Union (EBU), March 2015.
 W3C Recommendation, 14 December 2017. (See
 <xspecref href="http://www.w3.org/TR/2017/REC-html52-20171214/">https://www.w3.org/TR/html52/</xspecref>.)
 </bibl>
-<bibl id="imsc10" key="IMSC1.0">Pierre Lemieux,
-<titleref href="https://www.w3.org/TR/2018/REC-ttml-imsc1.0.1-20180424/">TTML Profiles for Internet Media Subtitles and Captions 1.0.1 (IMSC1)</titleref>,
-W3C Recommendation, 24 April 2018. (See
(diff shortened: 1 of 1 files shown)


nigelmegitt:
Thanks for addressing https://github.com/w3c/ttml2/pull/1123#pullrequestreview-254002495 @skynavga , this looks good to me now.

nigelmegitt:
I see that the normative reference to the profile registry is dated, which matches our normal policy for normative references, but it does mean that, formally, any new profile added to a future revision of the profile registry is excluded from those permitted, by the text:

> All profile designators which have the TT Profile Namespace as a prefix but are otherwise not listed in Table 5-2 – Profiles or in [TTML Media Type], §4, Registry, are reserved for future standardization, and must not appear in a conformant document instance. 

This seems contrary to the intention of having a registry. @skynavga can we change the reference link to https://www.w3.org/TR/ttml-profile-registry/ for this reference, exceptionally? Alternatively can we modify the text to say

> ... or in [TTML Media Type], §4, Registry, or in a a future revision of [TTML Media Type], are reserved...

?

skynavga:
@nigelmegitt ok, I made it generic (exceptionally), and also changed the label used in the bibliography to better match the short name of the document; we should probably discuss this exception in a TTWG call to get approval from the WG;

skynavga:
Tests at w3c/ttml2-tests#250.

PR #1125, opened 2019-07-01
Permit audio|image in body content model (#1014).

### LINKED ISSUE #1014: Allow audio element as a child of body.
The `body` element does not permit a child `audio` element, but the exclusion seems unmotivated. Either motivate its exclusion or permit its inclusion.

skynavga:
Closes #1014.


skynavga:
Tests at w3c/ttml2-tests#251.

PR #1126, opened 2019-07-03
Change xlink:show default to 'none' (#1040).

### LINKED ISSUE #1040: Define default value for xlink:show.
Specify a default value for `xlink:show` in §8.2.5, where the possible values are `new`, `replace`, `embed`, `other`, `none`. In the case of TTML, my suggestion is to use `none`, which is assigned the following semantics in [XLink 1.1]:

>The behavior of an application traversing to the ending resource is unconstrained by this specification. No other markup is present to help the application determine the appropriate behavior.

Alternatively, consider that SVG 1.1 restricts the value of `xlink:show` to `new` and `replace` (in similar linking contexts) and assigns a default value of `replace`. However, for TTML captions, I suspect that the default linking behavior should not replace the subtitle presentation; therefore, it seems better to make show semantics application dependent, which is accomplished by `none` (or perhaps `other`).

skynavga:
Closes #1040.


skynavga:
N.B. although this a substantive change to normative semantics, it is not testable for two reasons: (1) the semantics of the previous default value `new` were entirely implementation dependent, and (2) the semantics of the new default value `none` are also entirely implementation dependent, and, as a consequence the semantics of `new` is a subset of the semantics of `none` and vice-versa, making them members of an equivalence class.

PR #1132, opened 2019-07-14
Elaborate reference data embedding content model for data element (#1021).

### LINKED ISSUE #1021: Permit metadata in otherwise empty data element.
For `data` element, if `@src` specified, then content should not include non-lwsp `#PCDATA`, chunk, or source elements; i.e., it may include only metadata.class content; content model should be changed to read:

```
 #PCDATA |
 (Metadata.class*, chunk+) |
 (Metadata.class*, source+) |
 Metadata.class*                         // add this choice
```

As presently defined, if `@src` is specified, the only acceptable content model appears to be `#PCDATA`, where `#PCDATA` is `EMPTY`, which means that metadata cannot appear in a data element that uses `@src`.

Also, we need a term to refer to usage that expresses `@src`, in particular, I suggest the term _reference data embedding_, about which see also #1022.

skynavga:
Closes #1021.

### DIFF
--- a/spec/rnc/ttml2-embedded.rnc
+++ b/spec/rnc/ttml2-embedded.rnc
@@ -86,7 +86,7 @@ TTAF.data.content.extra = empty
 TTAF.data.content =
   (text |
     (TTAF.Metadata.class*,
-     (TTAF.chunk+ | TTAF.source+),
+     (TTAF.chunk+ | TTAF.source+)?,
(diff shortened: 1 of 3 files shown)


skynavga:
Note that the name of the branch associated with this PR has a typo: it should have been issue-1021-empty-_data_-metadata, and not issue-1021-empty-_date_-metadata.

PR #1133, opened 2019-07-14
Add #opacity-image feature designation (#883).

### LINKED ISSUE #883: Add #opacity-image feature.
The TTML2 features related to the `tts:opacity` attribute currently don't cover the case that this attribute can also apply to an `image` element. So a respective feature has to be added, e.g. `#opacity-image`.

skynavga:
Closes #883.

### DIFF
--- a/spec/profiles/ttml2-full.xml
+++ b/spec/profiles/ttml2-full.xml
@@ -149,6 +149,7 @@
     <feature value="required">#nested-span</feature>
     <feature value="required">#opacity</feature>
     <feature value="required">#opacity-block</feature>
+    <feature value="required">#opacity-image</feature>
     <feature value="required">#opacity-inline</feature>
     <feature value="required">#opacity-region</feature>
     <feature value="required" extends="#opacity">#opacity-version-2</feature>
--- a/spec/profiles/ttml2-presentation.xml
+++ b/spec/profiles/ttml2-presentation.xml
@@ -158,6 +158,7 @@
     <feature value="optional">#nested-span</feature>
     <feature value="optional">#opacity</feature>
     <feature value="optional">#opacity-block</feature>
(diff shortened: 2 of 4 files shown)


nigelmegitt:
This pull request needs an associated test in ttml2-tests.

skynavga:
This PR does not actually add any new functionality, so unless you have a processor that actually does something with profiles, then it is a moot point. The only test would be whether a profile aware processor recognizes the new feature designation defined by this PR.

nigelmegitt:
I was thinking we should make sure we have a test that verifies that opacity as applied to `image` does have an effect, as a minimum. I will check to see if one is already present.

skynavga:
@nigelmegitt I just checked, and it appears we do not, since our goal in TTML2 1ed testing was to test the set of features associated with the defined feature designations, and, since there was no feature designation for applying opacity to image, no test was required; so perhaps it is legitimate to ask what tests need to be added for features for which new feature designations are to be added; in any case, I don't believe this should block moving forward on the PR, but will be handled normally during the CR process for TTML2 2ed;

nigelmegitt:
@skynavga don't forget that we did agree to write tests alongside pull requests so that a) we would not have a big push to write tests at CR stage and b) as a secondary mechanism for verifying that we are agreed on the intent of any substantive change.

skynavga:
@nigelmegitt yes, I recall that, but I interpret that agreement as a guideline or goal, and not a concrete process rule; in general, we have not been overly diligent in following this guideline;

nigelmegitt:
@skynavga true, and I promised to keep encouraging us to meet that goal, which is what I'm doing here!

skynavga:
@nigelmegitt see https://github.com/w3c/ttml2-tests/pull/205

PR #1135, opened 2019-07-23
Eliminate inconsistent use of xsd:nonNegativeInteger (#983).

### LINKED ISSUE #983: Change use of type xsd:nonNegativeInteger to <non-negative-integer> for consistency.
The XML schema data type `xsd:nonNegativeInteger` is used (inconsistently) in the definitions of `chunk` and `isd:sequence`, when other definitions employ the internally defined `<non-negative-integer>`. For the sake of consistency, only one of these syntaxes should be used, namely, the internally defined syntax `<non-negative-integer>`. Note also that the XML schema syntax permits the lexical representation `-0` for zero (0), while the internal definition does not.

skynavga:
Closes #983.

In addition to eliminating the use of `xsd:nonNegativeInteger`, this PR

- eliminates use of `xsd:positiveInteger`
- adds notes regarding value equivalence of `-0`, `0`, `+0`, etc.
- improves editorial consistency

### DIFF
--- a/spec/rnc/ttml2-datatypes.rnc
+++ b/spec/rnc/ttml2-datatypes.rnc
@@ -72,7 +72,7 @@ TTAF.DataFormat.datatype =
   TTAF.URI.datatype
 
 TTAF.DataLength.datatype =
-  xsd:nonNegativeInteger
+  string
 
 TTAF.DataSource.datatype =
   TTAF.URI.datatype
@@ -164,7 +164,7 @@ TTAF.FragmentIdentifier.datatype =
   xsd:anyURI
 
 TTAF.FrameRate.datatype =
-  xsd:positiveInteger
+  string
 
 TTAF.FrameRateMultiplier.datatype =
   xsd:string { pattern = "\p{Nd}+\s+\p{Nd}+" }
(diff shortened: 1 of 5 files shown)


skynavga:
This PR modifies the normative document syntax in only the following sense: by removing certain redundant lexical representations from (1) the `@length` attribute of the `data` and `chunk` element types and (2) the `@size` attribute of the `isd:isd` element type. In particular, the following redundant lexical representations are excluded (for use by the above attributes) by this change:

- the use of `-0` as a lexical alias for `0`
- the use of `+` as sign indicator

In the editor's opinion, this change is of a zero or near-zero risk nature due to (1) the early state of industry roll-out of TTML2, particularly the features associated with these element types, and (2) the extremely unlikely possibility that any author has gone out of their way to use any of these redundant lexical representations when their canonical form (without any sign indicator) would suffice.

All other changes in this PR are of an editorial nature and have no conformance impact.

nigelmegitt:
In general I prefer to make use of native XSD-defined types where possible, so in that sense changing from `xsd:nonNegativeInteger` to `xsd:string` seems like the wrong direction, but I'd like to discuss this to understand the motivation better. Adding an agenda label.

skynavga:
@nigelmegitt in none of these cases is there an issue to file, since the value of the attribute only includes `<non-negative-integer>` as a sub-component of its overall syntax, and its overall syntax requires use of `xs:string` due to the presence of other syntactic components; e.g., `tts:zIndex` is typed as `<integer>` which includes an optional sign component, `tts:repeatCount` may have the value `indefinite`, etc.

PR #1146, opened 2019-08-30
Define alterability of initial values; specify that tts:ruby is unalterable (#1140).

### LINKED ISSUE #1140: Ruby constraints cannot be validated prior to ISD construction
Validation of ruby constraints can only happen after ISD construction since:

* the `tts:ruby` style property can be set using `initial` and impacted by conditionalized styles
* ruby constraints are specified in terms of the computed value of the `tts:ruby` style property

Is that the intent, or are should ruby style properties be specified solely using:
* inline styling
* referential styling
* chained referential styling
* nested styling

edit.: added impact of conditionalized styles

skynavga:
Closes #1140.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -8366,8 +8366,12 @@ such usage should be reported as a warning.</p>
 </ulist>
 <div3 id="styling-vocabulary-initial">
 <head>initial</head>
-<p>The <el>initial</el> element is used to modify the initial value of one or more style properties, i.e, to specify use of different value(s) than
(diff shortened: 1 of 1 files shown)


skynavga:
Tests at w3c/ttml2-tests#221.

nigelmegitt:
This pull request was opened before the TTWG worked out what approach it wanted to take to resolving the issue, or even discussing it at all - note that the TTWG did decide that it would be worthy of face to face discussion at TPAC since the issue is potentially complex.

skynavga:
Per TPAC 2019 TTWG meeting, closing this PR without merging.

PR #1149, opened 2019-08-31
Constrain @type on data element with reference data embeddings (#1022).

### LINKED ISSUE #1022: Constrain use of @type on data element for reference data embeddings.
For `data` element, if `@type` is specified for a _reference data embedding_, then reference must not be to an _internal data source_; conversely, if reference is to an _external data source_, then `@type` should be specified.

Note that _reference data embedding_ needs to be added as a fourth content model described in the paragraph that follows the second note in §9.1.3. See also #1021.

skynavga:
Closes #1022.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -7474,7 +7474,9 @@ If <loc href="#terms-reference-data-embedding">reference data embedding</loc> is
 a <loc href="#embedded-content-attribute-type"><att>type</att></loc> attribute should be specified, and,
 if specified, must correspond with the media (content) type of the referenced data resource;
 otherwise (an <loc href="#terms-internal-data-resource">internal data resource</loc> is referenced),
-a <loc href="#embedded-content-attribute-type"><att>type</att></loc> attribute must not be specified.
+a <loc href="#embedded-content-attribute-type"><att>type</att></loc> attribute must not be specified,
+in which case the content type is determined by the value of the <loc href="#embedded-content-attribute-type"><att>type</att></loc> attribute
+on the referenced <loc href="#embedded-content-vocabulary-data"><el>data</el></loc> element.
 If <loc href="#terms-sourced-data-embedding">sourced data embedding</loc> is used,
 a <loc href="#embedded-content-attribute-type"><att>type</att></loc> attribute must not be specified,
 in which case the media (content) type of the resolved <loc href="#embedded-content-vocabulary-source">source</loc> element is used as the type.</p>
@@ -7482,10 +7484,6 @@ in which case the media (content) type of the resolved <loc href="#embedded-cont
 such as <code>application/octet-stream</code>, and additional format information is known about a referenced data resource that cannot be
 expressed in a Media Type <bibref ref="media-types"/> parameter,
 then a <loc href="#embedded-content-attribute-format">format</loc> attribute may be specified as a hint to the content processor.</p>
-<note role="clarification">
-<p>In the preceding prose, the term <quote>specified</quote>, if not otherwise qualified, is to be read as specified on the current <el>data</el> element,
-and not specified on another element, for example, an element referenced by the current <el>data</el> element.</p>
-</note>
 <p>The use of <loc href="#terms-simple-data-embedding">simple data embedding</loc> is illustrated by the following examples.</p>
 <table id="embedded-content-vocabulary-data-example-1" role="example">
 <caption>Example Fragment &ndash; Simple Data Embedding in Data Defining Context</caption>
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -1173,9 +1173,9 @@ a <loc href="#profile-vocabulary-extension"><el>ttp:extension</el></loc> element
 <gitem id="terms-external-data-resource">
 <label>[external data resource]</label>
 <def>
-<p>A <loc href="#terms-data-resource">data resource</loc> external to a
+<p>A <loc href="#terms-data-resource">data resource</loc> the representation of which is external to (not contained within) a
 <loc href="#terms-timed-text-content-document-instance">timed text content document instance</loc>, referenced by means
-of a <loc href="#embedded-content-vocabulary-source"><el>source</el></loc> element or a <loc href="#embedded-content-attribute-src"><att>src</att></loc>
+of a <loc href="#embedded-content-attribute-src"><att>src</att></loc>
 attribute of a <loc href="#embedded-content-vocabulary-source"><el>source</el></loc> element or <loc href="#terms-embedded-content-element">embedded
 content element</loc>.</p>
 </def>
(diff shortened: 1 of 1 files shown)


skynavga:
Tests at w3c/ttml2-tests#222.

spec/ttml2.xml
```diff
@@ -7457,10 +7467,15 @@ data integrity during transport, i.e., the data could be modified without modify
...
-In the case of <loc href="#terms-sourced-data-embedding">sourced data embedding</loc>, a <loc href="#embedded-content-attribute-type"><att>type</att></loc> attribute must not be specified, in which case the media (content) type of the resolved <loc href="#embedded-content-vocabulary-source">source</loc> element
-is used as the type.</p>
+If <loc href="#terms-reference-data-embedding">reference data embedding</loc> is used, then, if it refers to an
+<loc href="#terms-external-data-resource">external data resource</loc>,
+a <loc href="#embedded-content-attribute-type"><att>type</att></loc> attribute should be specified, and,
+if specified, must correspond with the media (content) type of the referenced data resource;
+otherwise (an <loc href="#terms-internal-data-resource">internal data resource</loc> is referenced),
+a <loc href="#embedded-content-attribute-type"><att>type</att></loc> attribute must not be specified.
+If <loc href="#terms-sourced-data-embedding">sourced data embedding</loc> is used,
+a <loc href="#embedded-content-attribute-type"><att>type</att></loc> attribute must not be specified,
```
nigelmegitt: I read this at first to be inconsistent: referencing an externally represented resource using a `src` attribute on an embedded content element, the `type` **should** be present, but using a `src` attribute on a `source` element, it **must not** be? 

Perhaps this text needs to be edited to say that the `type` must not be specified _on the embedded content element_?

skynavga: This sentence is governed by (1) the immediate context of use, which is the semantics of the `data` element (only) and (2)  the initial "If reference data embedding is used". This means that this language is only referring to the use of `@src` on a `data` element. There is other (congruent) language to be found regarding `@type` in the definitions of `audio`, `font`, and `image` elements, and there is more elaborate language regarding `@type` under the definition of `source`.

nigelmegitt: Two thoughts on this:

1. if we keep this semantic then adding a "Note that the referenced internal resource is expected to specify its type" would help.

2. this has taken me down a different path of understanding the problem. Why would a `data` element have `@src` that points to an internal data resource? Such a data resource must be defined by a `data` element (including any child `chunk`s), this simply means that it is permitted to chain references from `data` to `data`. That seems to be pointless (and potentially circular?). Wouldn't a reasonable alternative be to prohibit internal data referencing from `data@src` altogether?

skynavga: Regarding (1), it is already mandatory that internal resource definitions specify `@type`, so saying it again in a note is redundant; for (2), because I might want to do:

```
<head>
  <resources>
    <data xml:id="backupIconDataDefinition" type="image/png">
      VGhlIHF1aWNrIGJyb3duIGZveA==
    </data>
  </resources>
</head>
...
<image>
  <source src="icon1.png"/>
  <source>
    <data src="#backupIconDataDefinition"/>
  </source>
</image
...
<image>
  <source src="icon2.png"/>
  <source>
    <data src="#backupIconDataDefinition"/>
  </source>
</image
```

nigelmegitt: OK, I understand the use case, thank you.

I still think it would help the reader to be clear if the type being specified or not specified is on the _referencing_ `data` or on the _referenced_ `data`, and it could be clearer in the text; the scoping only partially helps since the `type` attribute being defined is on the same element, differing only in usage context, as opposed to specification context.

skynavga: @nigelmegitt for the sake of expediency, can we go with what we have now?

nigelmegitt: @skynavga since I think an edit to fix this ambiguity would be small, I'd prefer to take that path instead of leaving it as is.

For example, here's one proposal for your Editorial consideration:

```xml
If <loc href="#terms-reference-data-embedding">reference data embedding</loc> is used, then
a <loc href="#embedded-content-attribute-type"><att>type</att></loc> attribute should be specified
on the <loc href="#embedded-content-vocabulary-source"><el>source</el></loc> or 
<loc href="#terms-embedded-content-element">embedded content element</loc> whose 
<loc href="#embedded-content-attribute-src"><att>src</att></loc> attribute specifies 
the external resource location, and, if specified,
the value of the <loc href="#embedded-content-attribute-type"><att>type</att></loc> 
attribute must correspond with the media (content) type of the referenced data resource;
otherwise (an <loc href="#terms-internal-data-resource">internal data resource</loc> is referenced),
a <loc href="#embedded-content-attribute-type"><att>type</att></loc> attribute must not be specified.
```

skynavga: I cannot accept the edit for four reasons: (1) it substantively changes the current specification (by removing the qualification regarding external data resource), (2) it is perfectly obvious that the context for specification is the current `data` element, i.e., the _referencing_ `data` element where the reference data embedding is used, and could not by any stretch of the imagination be the context the _referenced_ element, (3) it breaks symmetry with the language regarding the other three other embedding methods which say "must (not) be specified" none of which have language that qualifies the context, i.e., which likewise imply the current `data` element, and, finally, (4) it expands "be specified" unnecessarily (in an already overly long paragraph). At most, I might add a note that rules out such a blatantly non-obvious reading that assumes it is the _referenced_ element that is being constrained here. I will add such a note (which, personally, I think is unnecessary).

nigelmegitt: > could not by any stretch of the imagination be the context of the _referenced_ element

The fact that I found this to be ambiguous is a proof to the contrary. The problem is with the wording "If reference data embedding is used", which brings into context the idea of both the referencing and the referenced element, and to my mind introduces the potential confusion.

I agree that the text should not need that note.
 
Why don't we make this even simpler, and just say:

> If the `src` attribute of a `data` or `source` element references an external resource then the `type` attribute should be specified, and if specified must correspond with the media (content) type of the referenced resource. Otherwise, (an internal data resource is referenced), a `type` attribute must not be specified.

?

skynavga: I'm opposed to changing the normative prose as you suggest. I believe the "alternate" reading you suggest to be unreasonable and not justified by the prima facie definition of reference data embedding. It is abundantly clear that the normative prose could *only* be addressing the current `data` element, i.e., the referencing element. The note that I have added here makes that quite clear and obviates any need to change the normative prose.

skynavga: @nigelmegitt if you feel an additional note is needed to resolve your concerns, then please suggest one.

skynavga: Amending my above, I compared similar language under `source`, and have pulled the appropriate bits into the normative prose and removed the now redundant note.

skynavga: See https://github.com/w3c/ttml2/pull/1149/commits/cb13ae044986d1076f468acf1f5d3d980be0fb7a.

skynavga: @nigelmegitt ping

nigelmegitt: Sorry, was away, this works for me, thank you.

PR #1151, opened 2019-09-03
Define and reference inline term absolutize (#1150).

### LINKED ISSUE #1150: "Absolutize" has no defined meaning in relation to URL.
> RFC 3986 does not define "absolutizing". I think you might mean "resolving a reference as defined in RFC 3986".

_Originally posted by @palemieux in https://github.com/w3c/ttml2/pull/1054/files_

The term "absolutizing" is used 9 times in TTML2 - this should be editorially fixed for clarity.

skynavga:
Closes #1150.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -2932,7 +2932,7 @@ When defined internally (inline), a profile is referenced either implicitly
 or by means of a <loc href="#content-value-fragment-profile-designator">&lt;fragment-profile-designator&gt;</loc> expression.</p>
 <p>A profile designator must adhere to the <loc href="#content-value-uri">&lt;uri&gt;</loc> value syntax.
 If the profile designator is expressed as a relative URI,
-then it must be absolutized by using the TT Profile Namespace value as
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -2932,7 +2932,7 @@ When defined internally (inline), a profile is referenced either implicitly
 or by means of a <loc href="#content-value-fragment-profile-designator">&lt;fragment-profile-designator&gt;</loc> expression.</p>
 <p>A profile designator must adhere to the <loc href="#content-value-uri">&lt;uri&gt;</loc> value syntax.
 If the profile designator is expressed as a relative URI,
-then it must be absolutized by using the TT Profile Namespace value as
+then it must be <termdef id="defs-absolutize" term=""><term>absolutized</term></termdef>, i.e., resolved as per <bibref ref="uri"/>, &sect;5, by using the TT Profile Namespace value as
```
nigelmegitt: This inline term definition is ambiguous since it is not clear if the subclause "by using the TT Profile Namespace value as the base URI." is included or not. Please could you look for a way to clarify that @skynavga ?

skynavga: See https://github.com/w3c/ttml2/pull/1151/commits/29698b141d19538e72ca526da6c1358ce3d9d0a4.

nigelmegitt: Thank you!

PR #1152, opened 2019-09-03
Add 'ignore' value to ttp:profile@combine, and make it the default (#1028).

### LINKED ISSUE #1028: Add ignore value to @combine on tt:profile.
Add `ignore` value to `@combine` attribute on `tt:profile`. It would also be better to change the default from 'replace' to 'ignore' as well to be consistent with `ttp:{content,processor}ProfileCombination`, but this may not be feasible in TTML2 2e. However, take note that we have a precedent for changing a default value (see `ttp:markerMode`).

skynavga:
Closes #1028.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -4115,7 +4115,7 @@ otherwise it is referred to as a <loc href="#terms-non-nesting-profile">non-nest
 <td>
 <eg xml:space="preserve">
 &lt;ttp:profile
-  combine = (<emph>"leastRestrictive"</emph> | <emph>"mostRestrictive"</emph> | <emph>"replace"</emph>) : replace
+  combine = (<emph>"leastRestrictive"</emph> | <emph>"mostRestrictive"</emph> | <emph>"replace"</emph> | <emph>"ignore"</emph>) : ignore
   designator = <loc href="http://www.w3.org/TR/xmlschema-2/#string">xsd:string</loc>
   type = (<emph>"processor"</emph> | <emph>"content"</emph>) : processor
   use = <loc href="#content-value-absolute-profile-designator">&lt;absolute-profile-designator&gt;</loc> | <loc href="#content-value-relative-profile-designator">&lt;relative-profile-designator&gt;</loc>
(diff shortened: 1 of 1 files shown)


skynavga:
FYI, it turns out that both XSD and RNC schemas already permitted `ignore`; so they were ahead of the spec in this regard.

spec/ttml2.xml
```diff
@@ -4154,7 +4154,10 @@ elements in post-order traversal order.</p></item>
 if the value is <code>mostRestrictive</code>, then the most restrictive specification value
 applies. The order of restrictiveness is as follows (from least to most):
 <emph>optional</emph>, <emph>required</emph>, <emph>prohibited</emph>.</p>
-<p>If the <att>combine</att> attribute is not specified, then replacement semantics apply.</p>
+<p>If the value of the <att>combine</att> attribute is <code>ignore</code>, then a
+lexically subsequent feature or extension specification is ignored,
```
nigelmegitt: The definition of ordering is currently scoped only to `replace` - please could you look at factoring it out so that the same ordering requirement applies to both `ignore` and `replace` @skynavga ?

skynavga:
Tests at https://github.com/w3c/ttml2-tests/pull/224.

nigelmegitt:
I've just left a request changes review on w3c/ttml2-tests#224.

spec/ttml2.xml
skynavga: I could, but would rather not do that in this PR, or even in this release. What you ask for is an editorial exercise that is not required to satisfy this issue.

nigelmegitt: How is it editorial only? Currently there is no definition of ordering for `ignore`, but there is one for `replace`. It would make no sense for the ordering definition to differ between the two, but the scoping means that as it stands, a different ordering would appear to be okay.

skynavga: In that case, I don't understand your original comment. This PR doesn't add/change anything related to `replace`, so why are you mentioning `replace`?

nigelmegitt: The specification elements order of processing for `replace` is listed explicitly; it is omitted in this pull request for `ignore`, therefore at present it is unclear which feature or extension specification should be used and which others must be ignored. I'm arguing that we should specify the exact same ordering for `ignore` as `replace`, and should do so by common reference rather than by duplication.

skynavga: I don't know what you mean by "it is omitted". In what way does the two new lines 4157 and 4158 omit the order of processing? Is "lexically subsequent" not sufficient for you? That is the same language we use above when describing `replace` above. In any case, I strongly oppose attempting to combine the two paragraphs dealing with `replace` and `ignore`. In particular, it will make it more difficult to parse on the part of the reader, and thus I cannot accept your editorial suggestion.

nigelmegitt: It is not the same as the language we use when describing `replace`. [Section 6.1.1](https://www.w3.org/TR/ttml2/#profile-vocabulary-profile) adds further detail to the concept of lexical ordering in relation to `replace`:

> If the value of the `combine` attribute is `replace`, then a lexically subsequent feature or extension specification replaces a lexically prior specification, where specification elements are ordered as follows:
>
>1.    specifications defined by a [baseline profile](https://www.w3.org/TR/ttml2/#terms-baseline-profile) referenced by a `use` attribute;
>
>2.    specifications defined by descendant `ttp:profile` elements in post-order traversal order;
>
>3.    specifications defined as descendant `ttp:feature` or `ttp:extension` elements in post-order traversal order.

skynavga: See https://github.com/w3c/ttml2/pull/1152/commits/379f5a9cdbb12b2470bcf3fd4f3372193e812e0a.

nigelmegitt: Works for me.

PR #1156, opened 2019-09-08
Clarify luminance gain prose (#1117).

### LINKED ISSUE #1117: Add informative note to tts:luminanceGain that tts:luminanceGain is not relative to graphics white of the display
Currently the definition of luminanceGain states:
_"The tts:luminanceGain attribute is used to specify the absolute luminance of a region when its presentation requires a greater dynamic range than that provided by the [SRGB] color space, and when that presentation requires an absolute luminance value."_

To provide further clarity, consider adding an informative note to the above to advise that when tts:luminanceGain is applied during presentation it defines an absolute color & brightness, and it does not reflect a luminance level which is in relation to the graphics white of the display.

skynavga:
Closes #1117.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -11553,7 +11553,7 @@ images whose processing generates <bibref ref="srgb"/> pixel values, and an abso
 only the <bibref ref="srgb"/> pixel values are transformed.</p></note>
 <p>When applying <att>tts:luminanceGain</att>, the linear optical output values
 (C<sub>r</sub>, C<sub>g</sub>, C<sub>b</sub>) of the <loc href="#terms-component">components</loc> of a pixel of a region are defined as
-80 cd m<sup>-2</sup> ∙ <att>tts:luminanceGain</att> ∙ (r, g, b), where (r, g, b) are the normalized linear <bibref ref="srgb"/> <loc href="#terms-component">components</loc> of the pixel.
+80 ∙ <att>tts:luminanceGain</att> ∙ (r, g, b), where (r, g, b) are the normalized linear <bibref ref="srgb"/> <loc href="#terms-component">components</loc> of the pixel.
 </p>
 <p>If a computed value of the property associated with this attribute is not supported,
 then a <loc href="#terms-presentation-processor">presentation processor</loc> must use the closest supported value.</p>
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -11485,12 +11485,14 @@ of the paragraph.</p>
 </div3>
 <div3 id="style-attribute-luminanceGain">
 <head>tts:luminanceGain</head>
-<p>The <att>tts:luminanceGain</att> attribute is used to specify the absolute luminance of a region when
+<p>The <att>tts:luminanceGain</att> attribute is used to determine the absolute luminance of a region when
 its presentation requires a greater dynamic range than that provided by the <bibref ref="srgb"/> color space, 
 and when that presentation requires an absolute luminance value.</p>
 <note role="clarification">
-<p>The <bibref ref="srgb"/> color space specifies an absolute reference white point luminance of 80 cd∙m<sup>-2</sup>. 
-As a result, colors specified in the <bibref ref="srgb"/> space cannot be
+<p>Absolute luminance is determined by means of a gain multiplier applied to the absolute reference white
+point luminance of 80 cd∙m<sup>-2</sup> of the <bibref ref="srgb"/> color space (and not the actual white luminance
+of some output display device).</p>
+<p>Without the ability to specify this gain, colors specified in the <bibref ref="srgb"/> space cannot be
 reliably composited with high-dynamic range (HDR) content intended for display with significantly
 greater peak luminance, such as content that uses the perceptual quantizer electro-optical transfer function (PQ EOTF)
 specified in <bibref ref="bt2100_1"/>. Specifically,</p>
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -11485,12 +11485,14 @@ of the paragraph.</p>
 </div3>
 <div3 id="style-attribute-luminanceGain">
 <head>tts:luminanceGain</head>
-<p>The <att>tts:luminanceGain</att> attribute is used to specify the absolute luminance of a region when
+<p>The <att>tts:luminanceGain</att> attribute is used to determine the absolute luminance of a region when
```
nigelmegitt: I think it is specifying, not determining. However to compute the end value, some additional calculation is needed.

spec/ttml2.xml
```diff
@@ -11485,12 +11485,14 @@ of the paragraph.</p>
...
 its presentation requires a greater dynamic range than that provided by the <bibref ref="srgb"/> color space, 
 and when that presentation requires an absolute luminance value.</p>
 <note role="clarification">
-<p>The <bibref ref="srgb"/> color space specifies an absolute reference white point luminance of 80 cd∙m<sup>-2</sup>. 
-As a result, colors specified in the <bibref ref="srgb"/> space cannot be
+<p>Absolute luminance is determined by means of a gain multiplier applied to the absolute reference white
```
nigelmegitt: There's another layer of potential confusion here, since some presentation systems may not be able to present the specified absolute luminance. In this case I think we're providing a rule for calculating the used value, which may then be further modified to become an actual value based on the capabilities and settings of the presentation device.

Using the phrasing "Absolute luminance is determined by" does not make it clear which value is being specified or determined.

Therefore I'd suggest that we express this in terms of the _used value_ being computed by multiplying the value of the attribute by the 80cd.m<sup>-2</sup> specified by [[SRGB]] as an alternative to the wording proposed here.

spec/ttml2.xml
skynavga: No, it doesn't specify the absolute luminance. It specifies a relative gain to apply to a reference absolute luminance to _determine_ the desired absolute luminance, which is why I changed _specified_ to _determined_ in this context.

spec/ttml2.xml
skynavga: Here, it is the computed value that is determined which happens to be the same as the used value; however, the actual value may be different. This is the default reading that applies to all of our style definitions, so we don't bother to repeat it under each style.

spec/ttml2.xml
nigelmegitt: I don't understand the difference between those two things @skynavga . Isn't it only a difference in the unit of luminance?

skynavga: `tts:luminanceGain` _specifies_ a unit-less gain multiplier which, when multiplied by 80 cd/m2, _determines_ an absolute luminance; `tts:luminanceGain` does not specify a value that is equal to absolute gain, in which case it would have been incorrectly named (as a gain), and should be renamed to `tts:luminance`.

nigelmegitt: Are you making the distinction between an attribute value that includes a units component vs one that does not? 

Since there's no established symbol for the "80cd.m<sup>-2</sup>" unit, it's hard for authors to write it down. I've been thinking of it as something that has these units which just happen to be defined by specification instead of including them in the value. But it still ends up with units one way or another. I think there's a term for this kind of "dancing around terminology distinctions", and it's probably not very complimentary! We probably shouldn't spend too long doing it. (_we already have_)

skynavga: No. By _units_ and _unit-less_, I am talking physics, not ttml syntax. If we had an attribute that specified absolute luminance, then it would be called `tts:luminance` not `tts:luminanceGain`, and it would specify a number that has an implied physical unit of candelas per square meter, e.g., `tts:luminance="80"` would mean a specified absolute luminance of 80 cd/m2. But that is not what we have, so we can't say that `tts:luminanceGain` specifies an absolute luminance. It expresses a multiplier (like `ttp:frameRateMultiplier`) that is multiplied by 80 cd/m2 to obtain an absolute luminance.

spec/ttml2.xml
```diff
@@ -11551,7 +11553,7 @@ images whose processing generates <bibref ref="srgb"/> pixel values, and an abso
 only the <bibref ref="srgb"/> pixel values are transformed.</p></note>
 <p>When applying <att>tts:luminanceGain</att>, the linear optical output values
 (C<sub>r</sub>, C<sub>g</sub>, C<sub>b</sub>) of the <loc href="#terms-component">components</loc> of a pixel of a region are defined as
-80 ∙ <att>tts:luminanceGain</att> ∙ (r, g, b), where (r, g, b) are the normalized linear <bibref ref="srgb"/> <loc href="#terms-component">components</loc> of the pixel.
+80 cd m<sup>-2</sup> ∙ <att>tts:luminanceGain</att> ∙ (r, g, b), where (r, g, b) are the normalized linear <bibref ref="srgb"/> <loc href="#terms-component">components</loc> of the pixel.
```
palemieux: Please remove the units attached to `80` per https://github.com/w3c/ttml2/pull/283#issuecomment-291964708

spec/ttml2.xml
```diff
@@ -11485,12 +11485,14 @@ of the paragraph.</p>
...
+<p>The <att>tts:luminanceGain</att> attribute is used to determine the absolute luminance of a region when
 its presentation requires a greater dynamic range than that provided by the <bibref ref="srgb"/> color space, 
 and when that presentation requires an absolute luminance value.</p>
 <note role="clarification">
-<p>The <bibref ref="srgb"/> color space specifies an absolute reference white point luminance of 80 cd∙m<sup>-2</sup>. 
-As a result, colors specified in the <bibref ref="srgb"/> space cannot be
+<p>Absolute luminance is determined by means of a gain multiplier applied to the absolute reference white
+point luminance of 80 cd∙m<sup>-2</sup> of the <bibref ref="srgb"/> color space (and not the actual white luminance
+of some output display device).</p>
+<p>Without the ability to specify this gain, colors specified in the <bibref ref="srgb"/> space cannot be
```
palemieux: I suggest writing:

_ applied to the constant absolute reference white point luminance of 80 cd∙m-2 of the [SRGB] color space (and not the actual white luminance of the ultimate output display device, which can vary across devices and viewing environments and configurations)._

spec/ttml2.xml
skynavga: Marking this for discussion at TPAC.

spec/ttml2.xml
skynavga: This expansion doesn't add anything to the original parenthetical, i.e., "ultimate" is redundant, and "which can vary across ..." is obviously implied.

spec/ttml2.xml
skynavga: My request for discussion is predicated on a need to understand https://github.com/w3c/ttml2/pull/283#issuecomment-290472127. In particular, my understanding of the current expression

>80 ∙ tts:luminanceGain ∙ (r, g, b)

is that (r, g, b) is a vector (triplet), which means that the result of this calculation is also a vector, where each component of the calculated vector is itself an absolute luminance with respect to each component's color alone (in cd/m2); in any case, if we say that we are computing absolute luminance, then the resulting value must be in cd/m2; otherwise, it is not absolute luminance.

skynavga: Addressed in https://github.com/w3c/ttml2/pull/1156/commits/858e50843cb854f99d391b1070f582599a6bd302.

skynavga:
@palemieux pls resolve conversations and approve

PR #1157, opened 2019-09-08
Update content models w.r.t. permissible styling attributes (#1155).

### LINKED ISSUE #1155: Allow for styling attributes and animatable styling attributes in the content model.
Extracted from #1153:

1. The introduction to the sections on set and animate has been changed to refer to "styling attributes", but the content model does not reflect that. Although this is syntactically correct (since attributes in other namespaces are excluded from the content model and pruned prior to validation) we can do something more useful: I propose adding a comment to the content model highlighting that styling attributes in other namespaces are also considered for processing in the animation semantic, even though they are excluded from the content model from a formal specification perspective.

As per https://github.com/w3c/ttml2/pull/1154#discussion_r321253150 where @skynavga wrote:

> [snip] however, point (1) is a wider problem, since there are 13 instances of `{any attributes in TT Style Namespaces}` in various content models that need to be dealt with as a whole. I suggest opening another issue for point (1), with the suggestion to change these to read `{any styling attribute}` or `{any animatable styling attribute}` depending on the context (the latter for `animate` and `set`, the former for the 11 other cases.

skynavga:
Closes #1155.

This PR is untestable for the same reasons described at https://github.com/w3c/ttml2/pull/1096#issuecomment-513468050.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -6051,7 +6051,7 @@ element group apply semantically to the <el>body</el> element.</p>
   <loc href="#content-attribute-xml-lang">xml:lang</loc> = <loc href="http://www.w3.org/TR/xmlschema-2/#string">xsd:string</loc>
   <loc href="#content-attribute-xml-space">xml:space</loc> = (<emph>"default"</emph> | <emph>"preserve"</emph>)
   {<emph>any attributes in TT Metadata Namespace</emph>}
-  {<emph>any attributes in <termref def="defs-tt-style-namespaces">TT Style Namespaces</termref></emph>}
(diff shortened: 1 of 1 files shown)


nigelmegitt:
I think this is in fact testable, since it is constraining the set of styling attributes that are permitted. Therefore an invalidity test can be constructed that includes an `animate` or `set` element (or both) that targets `tts:ruby`, for example.

skynavga:
@nigelmegitt yes, that is true, but only for the changes to the content models of `animate` and `set`; however, I planned to add tests for that constrain (can't target non-animatable styles) in the context of #1154, which adds other language proscribing this usage without needing the change in content model to do so

nigelmegitt:
@skynavga okay, that would work too.

PR #1162, opened 2019-09-22
Remove constraint on presence of xml:id (#989).

### LINKED ISSUE #989: Back out requirement for xml:id on out-of-line region.
The following in 11.1.2 inadvertently invalidates valid TTML1 content where this constraint did not appear. Suggest changing 'must' to 'should'.

>An out-of-line `region` element must specify an `xml:id` attribute.

Note that there is no harm in having an out-of-line region without an ID. It just means that no `@region` attribute can reference it. So, while such usage is inefficient, it doesn't break anything.

skynavga:
Closes #989.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -1537,9 +1537,7 @@ non-nesting profile is <code>null</code>.</p>
 the <loc href="#terms-content-element">content element</loc> or <loc href="#layout-vocabulary-region"><el>region</el></loc> element associated with the animation.
 An out-of-line animation appears as a child of an <loc href="#animation-vocabulary-animation"><el>animation</el></loc> element in the header
 (<loc href="#document-structure-vocabulary-head"><el>head</el></loc> element)
-of a <loc href="#terms-document-instance">document instance</loc>, and specifies an <loc href="#content-attribute-xml-id"><att>xml:id</att></loc> attribute
-which value is referenced by an <loc href="#animation-attribute-animate"><att>animate</att></loc> attribute
-on the associated element to be animated. There is a one-to-many relation between
+of a <loc href="#terms-document-instance">document instance</loc>. There is a one-to-many relation between
 a referenced out-of-line animation element and referencing <loc href="#terms-content-element">content elements</loc> and
 <loc href="#layout-vocabulary-region"><el>region</el></loc> elements.</p>
 </def>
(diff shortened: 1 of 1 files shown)


skynavga:
@nigelmegitt see https://github.com/w3c/ttml2/pull/1162/commits/e716bcb60f070c70dbf0858b26c93d71c4ed7984. It is not necessary to add any informative text re-stating the XML specification that an IDREF must reference a specified ID, and that, as a consequence, the elements referenced by `@region` and `@animate` must indeed specify `xml:id`. Note that the same can be said of `@style`.

skynavga:
Need to remove remnants that refer to removed requirement for `xml:id` in definitions of out-of-line region (animation) terms.

palemieux:
@skynavga To summarize, the requirement for `@xml:id` to be present on out-of-line region is simply removed, right?

skynavga:
Correct

cconcolato:
This would result the same (strange) behavior in TTML1 and TTML2 that if `layout` contains any `region` (with or without `xml:id` they are considered as out-of-line regions), then content does not use a `region` attribute will not be rendered because no default region would be created. Am I correct?

skynavga:
@cconcolato correct; I agree it is strange, but that is the current TTML1 behavior, and it is best that TTML2 be consistent with TTML1

skynavga:
@cconcolato giving this some more thought, we could add a new rule to the [associate region] procedure just after the current rule 3 and before the current rule 4 that states:

>if a single, unidentified, i.e., lacking an `xml:id` attribute, out-of-line region is present, then the element is associated with that region;

this (or its equivalent) could also be back-ported to TTML1 4e.

nigelmegitt:
@skynavga then what to do if there is more than one unidentified `region` element child of `layout`?

I'm not sure that specifying an anonymous default region like that is something that anyone has asked for - it does not seem like a bad idea _per se_, I'm just aware that we should be resistant to adding features without any documented requirement or use case.

skynavga:
@nigelmegitt we can conclude this thread at the upcoming call; i'm happy to proceed with the current PR as is, which is consistent with TTML1 as far as I know;

palemieux:
I do not see a clear use case for the scenario at https://github.com/w3c/ttml2/pull/1162#issuecomment-536098929 , so the resulting, perhaps weird, behavior is not fatal. As indicated in https://github.com/w3c/ttml2/pull/1162#issuecomment-536126863 , we should strive to keep alignment with TTML1. I therefore like the idea of proceeding with the PR as-is.

skynavga:
Tests at w3c/ttml2-tests#234.

PR #1167, opened 2019-09-23
Add special semantics for tta:speak (#1013).

### LINKED ISSUE #1013: Add a syntactic shortcut for an implied audio element.
If a `span` has `tta:speak` set to something other than `"none"` then currently it generates an audio resource but it is not clear that the audio resource should play, unless the `span` has a child `audio` element with `src="http://www.w3.org/ns/ttml/resource/#speech"`. That has the benefit that the `audio` element provides a home for potential `tta:gain` and `tta:pan` attributes, but if they are not needed, because their default values are fine, then it is rather verbose so I would like to introduce a syntactic shortcut that in the absence of such an `audio` child element, one is assumed.

I'd also like to avoid the need for the `src="http://www.w3.org/ns/ttml/resource/#speech"` attribute altogether for an `audio` element as a child of a `span` for which speech synthesis is enabled. I would propose that omitting both the `src` attribute and a `source` child in that case should suffice.

skynavga:
Closes #1013.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -14760,13 +14760,13 @@ character content.</p>
 and <emph>A</emph> does not reference any source <loc href="#terms-audio-resource">audio resource</loc> by means of either
 <loc href="#terms-reference-data-embedding">reference data embedding</loc> or <loc href="#terms-sourced-data-embedding">sourced data embedding</loc>,
 then a reference to the <loc href="#terms-speech-data-resource">speech data resource</loc> with a
-<loc href="#terms-reference-data-embedding">reference data embedding</loc> is implied.</p></item>
+<loc href="#terms-reference-data-embedding">reference data embedding</loc> is implied;</p></item>
 <item><p>if no child of <emph>S</emph> is an <loc href="#embedded-content-vocabulary-audio"><el>audio</el></loc> element and
 no ancestor of <emph>S</emph> is an <loc href="#terms-audio-generating-element">audio generating element</loc> for which a referenced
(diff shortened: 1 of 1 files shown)


nigelmegitt:
I should note that we need tests for this too.

spec/ttml2.xml
```diff
@@ -14751,6 +14751,80 @@ that defined by <bibref ref="ssml"/> &sect;3.2.4 where the equivalent of a <code
 element with a <code>rate</code> attribute equal to the computed value is applied to the spoken
 character content.</p>
 </note>
+<div4 id="style-attribute-speak-special-semantics">
+<head>Special Semantics of Speak</head>
+<p>If the computed value of <code>tta:speak</code> on some <loc href="#content-vocabulary-span"><el>span</el></loc> element or
+<loc href="#terms-anonymous-span">anonymous span</loc> <emph>S</emph> is not <code>none</code>, then</p>
+<olist>
+<item><p>if <emph>S</emph> has a child <loc href="#embedded-content-vocabulary-audio"><el>audio</el></loc> element <emph>A</emph>
```
cconcolato: Just to make sure. You used 'child' and not 'descendant'. Is that correct? Is it because tta:speak is inherited so you would simply apply the same special semantic on the child of S to determine the behavior for the grandchild of S, right?

skynavga: Correct.

nigelmegitt: Also worth noting that there is no intermediate element permitted between a `span` and its child `audio` element; there's no possibility of `<span><newElement><audio></audio></newElement></span>` at present.

skynavga:
Tests at w3c/ttml2-tests#236.

PR #1170, opened 2019-10-02
Fix syntax coloring problem by using character ref for apostrophe.

skynavga:
Trivial editorial change only.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -8,7 +8,7 @@
 <!ENTITY prevloc "https://www.w3.org/TR/2018/REC-ttml2-20181108/">
 <!ENTITY latestedloc "https://w3c.github.io/ttml2/index.html">
 <!ENTITY versionOfTT "2">
-<!ENTITY tbd "<phrase role='tbd'>To Be Defined</phrase>">
+<!ENTITY tbd "<phrase role=&apos;tbd&apos;>To Be Defined</phrase>">
 <!ENTITY sp "&#x0020;">
 <!ENTITY sect "&#x00a7;">
 <!ENTITY times "&#x00d7;">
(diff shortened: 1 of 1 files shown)


nigelmegitt:
I don't see a syntax coloring problem - is this to help you work in your editing environment @skynavga ? (just wondering if a better solution isn't to use an IDE that doesn't get confused by apostrophes...)

skynavga:
@nigelmegitt yes, emacs; i'm not open to changing editors, as I've been using this one for 45 years

nigelmegitt:
Hmm, sorry to be a pain on this one but I just opened up the spec XML in a vanilla emacs installation and there's no problem, which suggests it is a local emacs configuration issue:

image

On the grand scale of things this change is extremely minor; I'm worried that adding a new style rule for editing TTML2 that requires that the entity reference is used instead of apostrophes is going to be hard to remember and will make spec editing harder for all the Editors, and it doesn't seem like a great precedent to work around a local IDE issue with a global spec edit.

skynavga:
Yes you are being quite the pain. My gnu enacts config, newly installed
from macports on macOS Mojave is broken with no configuration fix in sight
and if you don’t bloody approve this I shall quit editing now.

nigelmegitt:
OK I've expressed a concern about the proposal with an understanding that the issue is minor, and received a disproportionate threat in response. This isn't good. I'm going to back off this now; perhaps another Editor has a view and could approve if they're happy with this? I won't block it. pinging @cconcolato @palemieux 

For reference I'm using emacs 22.1.1 that appears to have come pre-installed on my Mac; certainly I have no recollection of installing it, nor any evidence that I did so.

skynavga:
@nigelmegitt I use emacs 23.6 from macports rather than the apple default build (22.1). The newer version has a much changed syntax parsing configuration for xml and nxml modes. I use the nxml mode. This problem has been present for a while, which I have dealt with in the past by inserting ``<!-- " -->`` and ``<!-- ' -->`` comments. But this workaround has become untenable, and the only reasonable alternative is to use &apos; and &quot; as needed to ensure the syntax parser is happy.

When I am no longer editor, then someone can change these back to the non-entity characters if they wish, but, in the mean time, I need this change to be able to do my day to day editing. Furthermore, given the length of the ttml2.xml source file, it has become progressively more difficult to track down syntax problems that crop up from time to time. And to do this, I need my editor (emacs 23.6 with nxml mode) to deal with this situation.

What I find most disconcerting is that you should decide to withhold your approval for a trivial editorial change that should be routine by nature and which requires no comment at all. You are not the one doing day to day editing, and I am the only active editor. Would you prefer to delay my work and delay going to FPWD simply because you find my customary editing practice different than your own?

We have already invested far more time and energy into this routine editor's PR than is warranted. Is that what you would insist we do to make progress?

PR #1173, opened 2019-10-04
Clarify escape in literal convention (#987).

### LINKED ISSUE #987: Clarify that `\\` in specification literals means a single backslash.
In particular, need to clarify this in the context of the following which appears in 2.3 _Document Conventions_

>a literal term (specified within quotation marks), when present, must appear exactly as specified (excluding the quotation marks themselves)

skynavga:
Closes #987.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -2185,12 +2185,9 @@ SPACE (U+0020), TAB (U+0009), CARRIAGE RETURN (U+000D), or LINE FEED (U+000A),
 which corresponds to production [3] <code>S</code> as defined by <bibref ref="xml10"/>.</p>
 <p>The following conventions are used in the specification of value syntax expressions:</p>
 <ulist>
-  <item><p>a literal term (specified within double or single quotation marks), when present, must appear exactly as specified
-  (excluding the quotation marks themselves and after performing escape processing);</p>
-  <note role="example">
-    <p>For example, the literal <code>&apos;\\&apos;</code> is interpreted as a single backslash, i.e., REVERSE SOLIDUS (U+005C).</p>
-  </note>
-  </item>
+  <item><p>a literal term (specified within quotation marks), when present, must appear exactly  as specified
+  (excluding the quotation marks themselves);</p></item>
+  <item><p>the literal <code>&apos;\\&apos;</code> is interpreted as a single backslash, i.e., REVERSE SOLIDUS (U+005C);</p></item>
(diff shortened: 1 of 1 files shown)


skynavga:
> Why does `\` require escaping in string literals?

Because '\' in syntax literals is considered an escape character.

skynavga:
Moving this PR (and related Issue) to 3ED, as the commenter is asking for a change unrelated to the original issue which was merely to clarify the meaning of `\\` and not to question the existence of the convention.

spec/ttml2.xml
```diff
@@ -2185,8 +2185,12 @@ SPACE (U+0020), TAB (U+0009), CARRIAGE RETURN (U+000D), or LINE FEED (U+000A),
 which corresponds to production [3] <code>S</code> as defined by <bibref ref="xml10"/>.</p>
 <p>The following conventions are used in the specification of value syntax expressions:</p>
 <ulist>
-  <item><p>a literal term (specified within quotation marks), when present, must appear exactly  as specified
-  (excluding the quotation marks themselves);</p></item>
+  <item><p>a literal term (specified within double or single quotation marks), when present, must appear exactly as specified
+  (excluding the quotation marks themselves and after performing escape processing);</p>
```
nigelmegitt: If we're going to refer to an escape processing algorithm we need to define that algorithm. Since the document conventions define the value syntax expressions without reference to any other specification, there is no external definition on which to rely, and we must do that locally, i.e. it needs to be added in this pull request.

nigelmegitt: That's a surprise to me! I had no clue from the specification that the document conventions used in the specification are intended to be based on the CSS2.1 content syntax. It is not even clear to me that the referenced part of CSS2.1 has any implication for the document conventions in CSS specifications themselves: rather, it is part of the section that defines the grammar and syntax within stylesheet documents.

skynavga: We normatively reference XSL 1.1, which includes the `<string>` definition (with `text-align`) from CSS 2.0 (1998), which already introduces the backslash convention. We don't bother to explicitly call out every dependency on longstanding CSS technology that has pre-existed and been adopted in so many web specs.

nigelmegitt: I remain baffled by the idea that any reader might guess that the document conventions in TTML2 are based on the content syntax and grammar of CSS. They don't resemble CSS grammar in any other significant respect.

skynavga:
@palemieux can you approve or drop you change request?

PR #1179, opened 2019-10-04
Improve anonymous span prose, clarify rule ordering (#1139).

### LINKED ISSUE #1139: Improve anonymous span prose.
`[construct anonymous span](3)` specifies:

_for each span element whose child is a single anonymous span, replace the anonymous span with its sequence of child text nodes_

This means that `<span><anonymous_span></anonymous_span></span>` becomes `<span><text_node/></span>`.

However, `[construct anonymous span](1)` specifies:

_for each significant text node in a content element, synthesize an anonymous span to enclose the text node, substituting the new anonymous span for the original text node child in its sibling and parent hierarchy_ 

This means that `<span><text_node/></span>` becomes `<span><anonymous_span></anonymous_span></span>`.

Doesn't this yield an infinite loop? Perhaps `[construct anonymous span](3)` was intended to apply to nested anonymous span, but when is a nested anonymous span ever created in the first place?

skynavga:
Closes #1139.

Implement https://github.com/w3c/ttml2/issues/1139#issuecomment-526928441 modulo the following:

* generalize convention regarding ordering of rules (steps) in procedural (algorithmic) specifications;
* update like prose in §8.1.5 as well as §8.1.6;

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -17977,7 +17977,7 @@ reference to the related <loc href="#terms-image-resource">image resource</loc>;
 <item>
 <p>for the resulting <el>fo:block</el> formatting object
 produced in the previous step that corresponds to the <el>body</el>
-element, invoke the following sub-steps:</p>
+element, invoke the following ordered sub-steps:</p>
 <olist>
 <item>
 <p>if the <att>display-align</att> style property of this <el>fo:block</el>
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -17742,6 +17742,8 @@ some element becomes temporally active or inactive.</p>
 <gitem id="procedure-construct-anonymous-spans">
 <label>[construct anonymous spans]</label>
 <def>
+<p></p>
+<p>Perform the following ordered steps:</p>
 <olist>
 <item><p>for each <loc href="#terms-significant-text-node">significant text node</loc> child of a <loc href="#terms-content-element">content element</loc>,
 synthesize an <loc href="#terms-anonymous-span">anonymous span</loc> to enclose the <termref def="defs-text-node">text node</termref>, substituting
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -6253,9 +6253,7 @@ a <el>p</el> element, then it must be interpreted as having
 <emph>parallel</emph> time containment semantics.</p>
 <p>If a sequence of children of a <el>p</el> element
 consists solely of <loc href="#terms-character-information-item">character information items</loc>, then that sequence
-is subject to <phrase role="strong"><loc href="#procedure-construct-anonymous-spans">[construct anonymous spans]</loc></phrase> processing,
-which may result in their being wrapped in an <loc href="#terms-anonymous-span">anonymous span</loc> for the purpose of
-applying style properties that apply to <el>span</el> elements.</p>
+is subject to <phrase role="strong"><loc href="#procedure-construct-anonymous-spans">[construct anonymous spans]</loc></phrase> processing.</p>
 <note role="elaboration"><p>The presentation semantics of TTML effectively
 implies that a <el>p</el> element constitutes a line break. In particular,
 it is associated with a block-stacking constraint both before the first
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -2229,6 +2229,8 @@ value syntax, no whitespace is permitted unless explicitly marked in a value syn
 <p>The phrase &quot;and (or)&quot; denotes an <emph>inclusive disjunction</emph> (<emph>p&lor;q</emph>), while the syntax
 &quot;either&hellip;or&hellip;&quot; denotes an <emph>exclusive disjunction</emph> (<emph>p&xor;q</emph>). Otherwise, the sense of &quot;or&quot; is
 implied by its context of use.</p>
+<p>Unless stated to the contrary, all procedures (algorithms) that appear in the form of an ordered (numbered) list of rules, steps, or sub-steps
+are intended to be performed in the specified order.</p>
 <p>All content of this specification that is not explicitly marked as
 non-normative is considered to be normative. If a section or appendix
 header contains the expression "Non-Normative", then the entirety
@@ -4208,7 +4210,7 @@ of the <el>ttp:profile</el> element must be considered to be
 the empty (null) profile, i.e., an undesignated profile definition containing no
 feature or extension specifications.</p>
 <p>The <loc href="#semantics-profile-state-combined-profile-specification-set">combined profile specification set</loc> <emph>CPSS</emph> of features and extensions of a profile <emph>P</emph>
-is determined according to the following ordered rules, where merging a specification <emph>S</emph>
+is determined according to the following, where merging a specification <emph>S</emph>
 into <emph>CPSS</emph> entails applying a combination method in accordance with the specified
 (or default) value of the <att>combine</att> attribute, and where merging a <loc href="#semantics-profile-state-combined-profile-specification-set">combined profile specification set</loc> <emph>CPSS&apos;&nbsp;</emph> into
 <emph>CPSS</emph> entails merging each ordered specification of <emph>CPSS&apos;&nbsp;</emph> into <emph>CPSS</emph>:</p>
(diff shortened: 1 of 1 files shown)


skynavga:
@palemieux 

In what way does replacing instances of specific language with a general rule constitute a "significant change"?

In what way does this remove an important clarification?

skynavga:
Moving this PR (and related Issue) to 3ED, as this is clearly going to become another philosophical debate about editorial style, and only serve to delay 2ED publishing schedule.

nigelmegitt:
I agree with @palemieux here - it is clearer to state that the steps need to be executed in the specified order every time that is the case than to state exceptions where it is not the case.

skynavga:
@nigelmegitt that's letting the cart drive the horse; i'm not sure there are any exceptions to the rule

nigelmegitt:
> i'm not sure there are any exceptions to the rule

@skynavga that "I'm not sure" is the strongest argument yet for retaining the qualification locally to each list. If you as the editor are not sure if there are any exceptions, I reckon it's likely that most other readers will not be sure in the opposite direction, i.e. they will be unclear if the list of steps must be executed in the specified order or in any order.

skynavga:
@nigelmegitt if there is an exception (and I very much doubt it), then you can darn well be sure there is very explicit language pointing out that exception; but I would say the burden of proof is on someone who presumes there is an exception to the rule that would require us to explicitly state in each instance that rules are ordered (as opposed to stating that it is always the case for numbered steps/rules);

skynavga:
I should also note that the underlying _semantic_ markup is `olist` vs `ulist`, which makes it explicit, so I never would have used `olist` if I intended the list to be unordered.

nigelmegitt:
I thought about the difference between `olist` and `ulist` and realised that even with steps that can be executed in any order we would still need to use `olist` in the specification, otherwise we would find it difficult to refer to individual steps. I don't know if there is a good semantic match in HTML for a list that has unique numbers for referencing but whose members are not sorted.

My preference here remains not to make the change.

skynavga:
At present the spec is inconsistent because we have inadvertently added explicit language about explicit language about ordering in some but not all places where we should have not done so and stated a general rule or should have done so everywhere. It is redundant to fix this by taking the latter approach IMO when the former is safer (since we would not need to worry about forgetting to add explicit language in every new ordered list in the future since the general convention would cover it).

nigelmegitt:
> some but not all places 

which places have a list of steps but don't say if they must be executed in order?

skynavga:
- §3.1 _Document Conformance_
  - here `olist` enumerates a list of criteria with no implied ordering, so `ulist` would also work
- §3.2.1 _Generic Processor Conformance_
  - here `olist` enumerates a list of criteria with no implied ordering, so `ulist` would also work
- §3.2.2 _Transformation Processor Conformance_
  - here `olist` enumerates a list of criteria with no implied ordering, so `ulist` would also work
- §3.2.3 _Presentation Processor Conformance_
  - here `olist` enumerates a list of criteria with no implied ordering, so `ulist` would also work
- §3.3 _Claims_
  - here `olist` enumerates a list of criteria with no implied ordering, so `ulist` would also work
- §4 _Document Types_
  - `olist` doesn't state "in order" but prose uses "after" and "then" to make order explicit
- §5.2.4.2 _Content Profile Semantics_
  - [construct effective content profile]
    - the  first list says "any (or all) of the following", so this is an explicit exception to "in order"
    - the second list is missing "in order"
- §5.2.4.3 _Processor Profile Semantics_
  - [abort of unsupported processor profile]
    - missing "in order"
  - [construct effective processor profile]
    - outer level missing "in order"
    - item 2, inner level missing "in order"
  - [construct inferred processor profile]
    - outer level missing "in order"
    - item 2, inner level missing "in order"
  - [construct default processor profile]
    - missing "in order"
- §5.3.1.2 _Validation Processing Semantics_
  - [construct effective validation profile]
    - missing "in order"
  - [construct effective validation schema]
    - missing "in order"
  - [validate document]
    - missing "in order"
    - note that step 3 and the bulleted list and sublist in the paragraph that follows step 3 intentionally use `ulist` to indicate the use of unordered rules/steps (so do not mistakenly use ordered lists exceptionally marked as ordered by prose);
- §6.1.1 `ttp:profile` in two instances
- §6.1.3 `ttp:feature` in one instance (2nd bullet of 3 bullet `ulist`)
- §6.1.5 `ttp:extension` in one instance (2nd bullet of 3 bullet `ulist`)
- §9.1.5 `image` two element `olist` should be changed to bulleted `ulist` as no order is implied
- §10.2.12.1 `tts:direction` explicitly states exceptionally "the following ordered rules, where the first applicable rule applies"
- §10.2.16 `tts:extent`
  - 1st `olist` should be `ulist` since no order is implied
  - 2nd `olist` is missing "in order"
  - 3rd `olist` is missing "in order"
  - 4th `olist` is missing "in order"
- §10.2.27 `tts:lineHeight`
  - `olist` missing "in order"
- §10.2.40 `tts:showBackground`
  - `olist` missing "in order"
- §10.2.56.1 `tta:speak`
  - `olist` missing "in order"
- §10.4.2.2 _Region Style Inheritance_
  - `olist` should be `ulist` since no order is implied
- §10.4.5.1 _Containment Calculations_
  - `olist` missing "in order"
- §10.4.5.2 _Cover Calculations_
  - `olist` missing "in order"
- §11.3.1.3 _Intermediate Synchronic Document Construction_
  - [resolve timing]
    - `olist` missing "in order"
  - [construct anonymous spans]
    - `olist` missing "in order"
  - [associate region]
    - explicitly states exceptionally "the following ordered rules, where the first rule satisfied is used and remaining rules are skipped"
- §11.3.1.4 _Synchronic Flow Processing_
  - [flow transformation]
    - `olist` missing "in order"
- §13.1.1 `animate`
  - `olist` should be `ulist` since no order is implied
- §13.1.3 `set`
  - `olist` should be `ulist` since no order is implied
- §13.2.1 `animate`
  - `olist` missing "in order"
- §H.1 _Aspect Ratios_
  - **storage aspect ratio (SAR)**
    - `olist` missing "in order"
  - §H.1.1 _No Aspect Ratio_
    - `olist` missing "in order"
  - §H.1.2 _One Aspect Ratio_
    - `olist` missing "in order"
- §I.1 _Clock Time Base_
  - `olist` missing "in order"
- §I.3 _SMPTE Time Base_
  - `olist` missing "in order"

nigelmegitt:
Thank you for the analysis @skynavga . Given https://github.com/w3c/ttml2/pull/1179#issuecomment-542314468 I currently believe the best solution to this would be:

1. Use `olist` throughout so that steps may be explicitly referenced, AND:
2. Explicitly state prior to each list of steps if they must be executed in the specified order or in any order.

Point 2 above could be simplified perhaps by defining terms for:
* a "list of steps that must be executed, in the specified order", e.g. _specified-order steps_ and
* a "list of steps that must be executed, in any order", e.g. _any-order steps_, 

and using the applicable term as a shortcut for describing the list of steps with fewer words. We'd end up with language like "Execute the following any-order steps:".

skynavga:
@nigelmegitt I don't follow you. In none of the cases in my analysis is it the case that numbered items can be evaluated in any order **other than** the numbered order. In particular, in none of the cases does the text admit evaluation in **any order**.

To be very clear, numbered lists are **always** performed in the numbered order whenever they represent rules/steps/etc (as opposed to criteria/cases). The only distinction is whether (1) evaluation stops when the first rule/step is successfully evaluated or (2) evaluation conditions to the last numbered item regardless of whether a rule/step is successfully evaluated. The latter being the default, and the former being the exceptional case and explicitly marked in the text.

nigelmegitt:
> In none of the cases in my analysis is it the case that numbered items can be evaluated in any order

I am proposing that the fact that items are numbered should not be taken to imply that the steps are ordered. Numbering does two things: 1. it suggests some position within an order that may or may not be important for processing, 2. it facilitates external reference of the step, so we can say "step 3 does that job" or "there's a problem with step 5".

In your analysis @skynavga you include, for example:

>* §13.1.3 `set`
>   * `olist` should be `ulist` since no order is implied

If we do this, then we will no longer have an easy way to refer to the steps. However this would then mean that we need to be explicit in the text that no order is implied.

Hope that helps explain my thinking?

skynavga:
> I am proposing that the fact that items are numbered should not be taken to imply that the steps are ordered.

And I am saying that in all cases where numbered lists state rules or steps (of a procedural nature), then they are intended to be ordered (whether so stated or not).

>In your analysis @skynavga you include, for example:
>>§13.1.3 set
>>olist should be ulist since no order is implied
>
>If we do this, then we will no longer have an easy way to refer to the steps. However this would then mean that we need to be explicit in the text that no order is implied.

And in all of these cases (where `olist` was used when `ulist` should have been used), the items are not rules or steps (of a procedure) but are (unordered) criteria or constraints.

The correct (and logical) response is to change these latter to `ulist`, to add a general convention for numbered lists of rules/steps, and to remove references to "in the specified/indicated order (etc)".

nigelmegitt:
>  the items are not rules or steps (of a procedure) but are (unordered) criteria or constraints.

Regardless of this, it is still useful to be able to reference those criteria or constraints.

skynavga:
@palemieux I believe the above commits should address your comments. Let me know if otherwise.

skynavga:
@palemieux see https://github.com/w3c/ttml2/pull/1179/commits/54355ab2c4480d34c9ac7778d46088367894b9c8

palemieux:
@skynavga In the case of `p`, why wouldn't a child text node not always result in an anonymous span?

skynavga:
@palemieux I give up. I removed the text under `p` as well.

PR #1193, opened 2020-01-16
Add 2ED CR artifact.

skynavga:
For pre-publishing in [dated TR directory URI](https://www.w3.org/TR/2020/CR-ttml2-20200128/).


nigelmegitt:
Thank you for this @skynavga. I will wait until publication certain before approving this.

Issue #1189, opened 2019-12-19
Mention fingerprinting vectors in privacy considerations.


jyasskin:
If TTML is implemented natively in a user agent, it could expose [fingerprinting vectors](https://w3c.github.io/fingerprinting-guidance/) that aren't otherwise exposed. The [spec](https://w3c.github.io/ttml2/index.html#security-and-privacy) should mention this risk so native implementations know to make intentional choices. The things I noticed to call out are:

* Anything that's "implementation dependent" might be a fingerprinting vector.
   * Many initial style values are defined by the specification, so they wouldn't reveal anything. However, `<tts:color>` and some others are described as implementation-dependent.
* A user's preference for how fast they consume media (e.g. 2x vs 2.5x).
* The request for timed text indicates the user's language (which is also exposed in other ways) and that the user wants captions or subtitles (which isn't).
* The [`tts:fontFamily` attribute](https://w3c.github.io/ttml2/index.html#style-attribute-fontFamily) could expose the system's fonts and should use the [same restrictions as CSS](https://github.com/w3c/csswg-drafts/issues/4497).
* The `<audio>` and `<image>` elements probably allow the server to detect the value of any `<condition>` expression. Many of the `condition-function`s seem to be already exposed by CSS media queries. The `supports-function`s probably don't expose any more than the UA string. So this may only be extra fingerprinting surface for UAs that aren't also general web browsers. However, if one of these functions exposes a user preference or device attribute, that would be extra fingerprinting surface.

* [`ttp:clockMode==local`](https://w3c.github.io/ttml2/index.html#parameter-attribute-clockMode) probably reveals the local time zone, if only by the timing of embedded resource requests. [`ttp:timeBase==clock`](https://w3c.github.io/ttml2/index.html#parameter-attribute-timeBase) reveals clock skew in the same way.
* If there's a way to pull out a display frame rate, that would also help fingerprinting.

skynavga:
Just noting that all of the above comments except for the second to last (related to `audio` and `image` elements and `@condition` attribute) pertain to TTML1, and that the remaining comment pertains to TTML2 in general (not just 2nd edition). Though not necessarily addressed individually, many of these comments are covered indirectly by the remarks found in TTML2 First Edition [Appendix P _Security and Privacy Considerations_](https://www.w3.org/TR/2018/REC-ttml2-20181108/#security-and-privacy).

nigelmegitt:
[...]

PR #1203, opened 2020-04-02
Add consideration for font fingerprinting

### LINKED ISSUE #1202: CSS font-matching algorithm may introduce fingerprinting issues
Review of TTML2 2nd Edition noted many potential fingerprinting vectors: https://github.com/w3c/ttml2/issues/1189
(Whether those issues present a privacy risk depends on a clearer understanding of what information is revealed by content processors to whom.)

Addition of external font loading and the CSS font-matching algorithm could introduce those fingerprinting issues to IMSC 1.2.

Mitigations for fingerprinting in CSS are under discussion now in CSSWG and PING.

More info in email: https://lists.w3.org/Archives/Public/public-privacy/2020JanMar/0055.html

palemieux:
Closes #1202

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -26343,6 +26343,11 @@ W3C Recommendation, 11 May 2010.
 W3C Recommendation 8 June 2017. 
 (See <xspecref href="https://www.w3.org/TR/xslt-30/">https://www.w3.org/TR/xslt-30/</xspecref>.)
 </bibl>
+<bibl id="finger" key="FINGER">Nick Doty, Ed., <titleref
+href="https://www.w3.org/TR/2019/NOTE-fingerprinting-guidance-20190328/">Mitigating Browser Fingerprinting in Web Specifications</titleref>,
+W3C Interest Group Note, 28 March 2019. (See
+<xspecref href="https://www.w3.org/TR/2019/NOTE-fingerprinting-guidance-20190328/">https://www.w3.org/TR/2019/NOTE-fingerprinting-guidance-20190328/</xspecref>.)
+</bibl>
 </blist>
 </inform-div1>
 <inform-div1 id="requirements">
@@ -29431,6 +29436,18 @@ This language preference constitutes
 information about the user. However, the offering of a <loc href="#terms-timed-text-document-instance">timed text document instance</loc>
 and the choice whether to retrieve and process it are characteristics of the application that makes the offer (e.g. a web
 application based on <bibref ref="html"/>), rather than of the Document Instance itself.</p>
+</div2>
+<div2>
+<head>Font Matching</head>
+<p>By conditionally dereferencing (downloading) <loc href="#terms-font-resource">font resources</loc> based on the existence of
+locally-installed <loc href="#terms-font-resource">font resources</loc>, a <loc href="#terms-content-processor">content processor</loc>
+introduces a potential fingerprinting vulnerability as defined in <bibref ref="finger"/>. Existence and mitigation of such vulnerability depends on the 
+<loc href="#terms-content-processor">content processor</loc> implementation and overall system architecture.</p>
+
+<note role="example">
+<p>As an example, a mitigation strategy can involve ignoring user-installed <loc href="#terms-font-resource">font resources</loc>
+when choosing whether to dereference (download) <loc href="#terms-font-resource">font resources</loc>.</p>
+</note>
 </div2>
 </inform-div1>
 <inform-div1 id="hdr-compositing">
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -29438,7 +29438,7 @@ and the choice whether to retrieve and process it are characteristics of the app
 application based on <bibref ref="html"/>), rather than of the Document Instance itself.</p>
 </div2>
 <div2>
-<head>Font Matching</head>
+<head>Font Detection</head>
 <p>By conditionally dereferencing (downloading) <loc href="#terms-font-resource">font resources</loc> based on the existence of
 locally-installed <loc href="#terms-font-resource">font resources</loc>, a <loc href="#terms-content-processor">content processor</loc>
 introduces a potential fingerprinting vulnerability as defined in <bibref ref="finger"/>. Existence and mitigation of such vulnerability depends on the 


spec/ttml2.xml
```diff
@@ -29432,6 +29437,18 @@ information about the user. However, the offering of a <loc href="#terms-timed-t
 and the choice whether to retrieve and process it are characteristics of the application that makes the offer (e.g. a web
 application based on <bibref ref="html"/>), rather than of the Document Instance itself.</p>
 </div2>
+<div2>
+<head>Font Matching</head>
+<p>By conditionally dereferencing (downloading) <loc href="#terms-font-resource">font resources</loc> based on the existence of
+locally-installed <loc href="#terms-font-resource">font resources</loc>, a <loc href="#terms-content-processor">content processor</loc>
```
nigelmegitt, line 29443: Since this is in TTML2 where the `condition` attribute potentially allows for any implementation-defined query I think we need to add something in P.7 to state more clearly that the application of `@condition` to external resources could be used to leak fingerprinting information to a remote server, and that this includes fonts.

Then, further to this, we should additionally mention (as here) the potential for implementations to download fonts conditional on them not being locally installed, but modify this pull request to clarify that this is considered a separate conditional mechanism to the `condition` attribute.

skynavga: I oppose expanding this PR to address general aspects of `@condition`. Are we going to explode the issue of privacy here? Please, let's deal with the specific issue on the table. If absolutely needed, please open new issues and PRs separately.

nigelmegitt: The issue is about font fingerprinting in general; there are two mechanisms I can think of in TTML2 that allow font fingerprinting: `@condition` and the semantics of font resource loading. 

This PR only addresses the font resource loading mechanism so far. It is in my view appropriate to expand the PR to address the `@condition` mechanism in the same PR so we have one PR that addresses the issue.

skynavga: If you are going to open up the discussion to `@condition`, then you would need to account for the fact that this mechanism can use media query as well as user language preferences to conditionalize image fetching.

skynavga: In other words, you are exploding what I thought was going to be a simple problem and solution into a much wider discussion.

nigelmegitt: In that case, apologies for not articulating what I thought was already evident. I didn't realise you thought that `@condition` was irrelevant to this issue; I assumed that we would bring it in as part of the transfer of the issue into TTML2. In IMSC 1.2 `condition` is prohibited, so it didn't form part of that discussion.

It's a good point that media queries as well as language preferences come into play. I think the list is completed by including supported feature queries. In other words, the three types of [`<condition-function>`](https://www.w3.org/TR/ttml2/#content-value-condition-function) each need to be taken into account.

palemieux: @nigelmegitt @skynavga +1 to opening separate issues for considerations strictly beyond font fetch fingerprinting. Mitigation might be different.

npdoty: #1189 was previously opened to note issues of @condition and image and font resource downloading. 

I'm not sure why it's useful to exclude known issues from this update to privacy concerns and mitigations (and I'm not sure what "exploding" is in this context), but updates can be managed in separate issues/PRs if that's the preferred working mode. If there are other conditionalities for resource fetching (media query and language preference are noted above) that aren't described in #1189 or #1202, that would be useful to document.

spec/ttml2.xml
```diff
@@ -29432,6 +29437,18 @@ information about the user. However, the offering of a <loc href="#terms-timed-t
 and the choice whether to retrieve and process it are characteristics of the application that makes the offer (e.g. a web
 application based on <bibref ref="html"/>), rather than of the Document Instance itself.</p>
 </div2>
+<div2>
+<head>Font Matching</head>
```
npdoty, line 29441: This might better be titled "Font Detection", as the privacy issue is whether processing the document reveals to the host of the TTML document whether the user has a font locally installed.

spec/ttml2.xml
```diff
@@ -29432,6 +29437,18 @@ information about the user. However, the offering of a <loc href="#terms-timed-t
...
+<p>By conditionally dereferencing (downloading) <loc href="#terms-font-resource">font resources</loc> based on the existence of
+locally-installed <loc href="#terms-font-resource">font resources</loc>, a <loc href="#terms-content-processor">content processor</loc>
+introduces a potential fingerprinting vulnerability as defined in <bibref ref="finger"/>. Existence and mitigation of such vulnerability depends on the 
+<loc href="#terms-content-processor">content processor</loc> implementation and overall system architecture.</p>
+
+<note role="example">
```
npdoty, line 29447: There was a suggestion in the past call that this could be a normative mitigation rather than just an example. E.g. 
> A content processor SHOULD NOT dereference external font resources conditionally on the presence of user-installed fonts, where that dereferencing could reveal information about the user's system or fingerprint the user.

Just a suggestion. Using SHOULD NOT might provide the flexibility for processor implementations that don't have end user privacy concerns, while still detailing an effective interoperable mitigation.

skynavga: We can't use SHOULD NOT in an informative appendix.

spec/ttml2.xml
nigelmegitt: The issue here is that font fetching might be a fingerprinting vector. The mechanisms by which that might occur include both ordinary processing of `font` elements and such processing in combination with `@condition`. To my mind this is one issue that should ideally have one pull request applied to fix it. There may also be a separate issue for generalising the fingerprinting opportunities of `@condition` beyond fonts, but I'm not proposing that we attempt to tackle that here, if indeed such an issue exists.

spec/ttml2.xml
npdoty: Right, the suggestion was to include a normative mitigation. If the group wanted to do that, then this appendix could be changed to be normative, or the requirement would be added elsewhere.

skynavga: Note that there are likely to be objections to introducing any normative language at this juncture of the process. Of course, such changes could be entertained at a different point in future processes.

samuelweiler: @skynavga  What is the substance of those potential objections?

skynavga: @samuelweiler since I will be the source of the objection, I can explain: there are two arguments, one technical, and another non-technical; the former is that any normative requirement or recommendation we could make in this regard would be untestable, since dereferencing occurs in the context of the application processing environment, which is wholly outside the scope of TTML; whether an implementer chooses to dereference an external font in the presence of a user installed font is a question for the implementor and not for TTML as we have specified it to date; TTML does not define the implementation with respect to resource fetch semantics or the privacy considerations the implementer ultimately decides to adopt; all we can do is take not of privacy issues, without going as far as making recommendations (SHOULD or SHOULD NOT); the second argument is based on process: namely, we (as a group) already decided not to introduce substantive changes after CR unless there was a clear technical breakage ***in a feature introduced in this CR***; the subject of this issue is not related to any change introduced in TTML2 2nd Edition; existing privacy considerations in TTML are cast as non-normative, and I expect it to stay that way; lastly I should note, however, that my objecting will not necessarily prevent the group from adopting a normative change if the group chooses to override my objection; my counsel will be that such a change is ill-advised, unnecessary, and untestable;

nigelmegitt:
I've added this to the agenda for this week's call at w3c/ttwg#118. I believe the main outstanding questions we need to resolve, from this pull request, are:

1. Change the title of the "Font Matching" section to "Font Detection" as per https://github.com/w3c/ttml2/pull/1203#discussion_r409704111
2. How best to articulate that other detections may be possible using the `condition` attribute, as per https://github.com/w3c/ttml2/pull/1203#discussion_r411271787

nigelmegitt:
For the discussion on Thursday, my hope is that we can agree to change "Font Matching" to "Font Detection" as per Nick's comment.

I also hope we can get to agreement on whether the use of the `condition` attribute provides any further ability to detect fonts. If we agree that it does not, then I will be happy to proceed with the current pull request. Then, consequently, I would like to know if the media queries available via the `condition` attribute are considered a fingerprinting risk by PING. If not, then no further action need be taken. However if they are, then I would like to learn more about this, since it seems to me that exactly the same risk must exist in CSS, and therefore there may be some off the shelf potential mitigations available. 

More information about the potential concern: the `condition` attribute supports queries including media queries. All the external resource elements, `audio`, `image`, `source` etc support this `condition` attribute. Therefore it would be possible, while processing a TTML2 document, to request a URL from a remote origin based on the result of a media query, thus providing information to that origin. This is effectively the same functionality as CSS Media Queries provides – see https://drafts.csswg.org/mediaqueries-4/#priv-sec and https://github.com/w3c/csswg-drafts/issues/3488 for example. The CSS3 Media Query Rec at https://www.w3.org/TR/css3-mediaqueries/ from 2012 does not include any security and privacy section.

spec/ttml2.xml
nigelmegitt: @npdoty we agreed to apply this change, and this has been actioned.

spec/ttml2.xml
nigelmegitt: In today's call TTWG agreed that `@condition` does not _add_ to the font fetching vector, and can be handled separately, and that no further change is needed prior to merging this PR.

spec/ttml2.xml
skynavga: @samuelweiler I'm not trying to persuade you of anything. You asked for the reason, and I gave you my answer. The current consensus of the WG is reflected in https://github.com/w3c/ttml2/pull/1203#issuecomment-638923838.

skynavga:
Per https://github.com/w3c/ttml2/pull/1203#issuecomment-638923838 (see last sentence), we have approvals and consensus for merging.

nigelmegitt:
> Per [#1203 (comment)](https://github.com/w3c/ttml2/pull/1203#issuecomment-638923838) (see last sentence), we have approvals and consensus for merging.

@skynavga That was a TTWG resolution, but in https://github.com/w3c/ttml2/pull/1203#issuecomment-638984684 I asked for the non-TTWG members who raised/commented on the issue for their view, with the intent of _not_ merging before we had confirmation. If there are further comments we may now need to open a new pull request and review cycle: I had been hoping to avoid that; it may still not be needed of course.

spec/ttml2.xml
nigelmegitt: @samuelweiler In my view testability is a Process requirement, or very close to it: in order to exit CR, implementability must be demonstrated, and that is in general done with tests. If we put untestable normative requirements into the spec then we can never demonstrate implementability. I will object to any change that prevents CR exit.

(background) In the past couple of years TTWG introduced a working practice of accompanying all substantive spec changes with tests at the time those spec changes are introduced, both so that the test suite does not lag behind the spec (making CR exit speedier) and so that the WG can better understand the intent of the spec change.

In this case the proposed normative change does affect semantics not defined by TTML, and writing a test for the change would be extremely challenging.

On the process point, I am less clear than @skynavga that the review scope is limited to changes introduced in this CR. I'm happy to be reminded, but I don't recall agreeing that, or making that part of the review request. I also think it is entirely reasonable for new issues to be raised against the spec, regardless of when they were introduced; however it is also reasonable for TTWG to defer issues raised against parts of the spec that were not changed by this CR.

samuelweiler:
> @samuelweiler @npdoty The TTWG has consensus to merge this PR. Are you in agreement that it resolves the issue to the best that we can get to at this stage? This does not prejudice us against making more normative statements in future editions, should we get agreement to do so.

No.  As in https://github.com/w3c/ttml2/pull/1203#discussion_r438418061, I think a normative mitigation is in order.

PR #1207, opened 2020-04-17
Issue 1204 positioning example

### LINKED ISSUE #1204: Include better positioning example in Introduction
From implementor feedback: the examples in the introduction at [1.2 Document Example](https://www.w3.org/TR/ttml2/#example) seem to suggest that specifying a `region` attribute on the `body` element is a good idea. 

However as we know from [[construct intermediate document]](https://www.w3.org/TR/ttml2/#procedure-construct-intermediate-document) step 3d, the effect of this is that if any descendant specifies a different region, then that descendant gets pruned. This is non-obvious to the casual reader, but the effect is that, except in simple cases, setting `@region` on `body` is probably a bad idea.

Another data point: setting `region` on `body` is prohibited in the EBU-TT-D profile.

I propose that we:

1. Change the TTML Body example fragment from:

```xml
<body region="subtitleArea">
  <div>
```

to:

```xml
<body>
  <div region="subtitleArea">
```

and

2. Add additional examples demonstrating positioning, perhaps with two regions defined, one at the top and one at the bottom, with content in both regions.

nigelmegitt:
Closes #1204 by adding a note and an example.

During w3c/ttwg#107 @skynavga objected to moving the region specification in the last example from the `body` element to the `div` element and instead suggested adding an explanatory note.

This pull request does that, and also adds a further example document showing content in two regions, overlapping temporally. The sample renderings are generated by sandflow/imscJS based on the source added at `spec/images/sources/example-complex-layout.ttml` - this necessarily includes both `tts:origin` and `tts:position` because that library currently does not support `tts:position` (since it is not part of IMSC). The example in the specification document omits the `tts:origin`.


nigelmegitt:
@skynavga please could you be more specific about the issue with the note? Is the problem only with the second part, for example? What did you have in mind as an alternative when you proposed this compromise in our call on Thursday?

As for the additional example, this is an informative example in an informative section, so it would be helpful to know more about why you think it is a bad idea to add it, given that it has no substantive impact and does, in my opinion, help to address the implementer feedback I received. That feedback included a statement that the examples are typically treated as examples of common practice, and are therefore misleading. My previous proposal was to amend the existing example, and you also objected to that.

skynavga:
@nigelmegitt upon additional consideration, I have concluded that (1) the current example does not warrant any comment since it only makes use of `@region` on `body`; had it also made use of `@region` on a descendant, then a comment would be in order; (2) the comment attempts to paraphrase a quite complicated piece of semantics in a introductory context where such details are intentionally excluded and doing so would only serve to confuse the reader; (3) the TTML2 specification is not a user guide, but a technical specification, and, as a consequence, authoring information is intentionally minimized;

nigelmegitt:
@skynavga I'm not sure, but it seems that you are dismissing the issue rather than accepting a proposal to resolve it?

Absent a better user guide, implementers do in fact read the introduction hoping to glean some kind of usage hints: that was the feedback I received that caused me to raise #1204. As things stand, the examples do not give a realistic sense of any typical usage that I'm aware of. That being the case, it seems reasonable to make some change to address the point.

If you feel that the note is going too far, I'd accept removing it if we could agree to add the new example, since the new example pulls the previous threads together in a "here's what it might look like in practice" sense, i.e. it provides what those readers are hoping to find, without going too far down the road of producing a user guide.

skynavga:
@nigelmegitt my larger issue is that this is not a typo fix, and this is not a wide review comment resolution; this is a new issue that is proposed to be resolved in CR at a point in the process after we have agreed to only entertain typo fixes and wide review comments; so if you want to move this to 3ED, then feel free to do so

nigelmegitt:
> this is not a wide review comment resolution

Ah, this is where we differ. We are not in the Process phase where we must demonstrate wide review, i.e. prior to entering CR, but comments are still welcome, as stated in the CR SOTD, and this is exactly the kind of implementer feedback that we need to accept and respond to positively.

Arguably this feedback was primarily provided against TTML1, but to my mind the best value in terms of future-facing fixes is to fix within TTML2. This also gives the opportunity to show how, for example, using `tts:position` is a good choice for region positioning.

So I am treating this issue as a wide review comment.

skynavga:
@nigelmegitt this issue is clearly not related to a change made in 2nd edition, or even in TTML2, it is clearly not related to a substantive bug, and it is clearly not a typo fix; therefore, it is an attempt to introduce a change that is outside the scope of what can be considered a response to the normal review process of a CR; as such, I find it to be NIGO (not in good order) as a wide review comment, and object to it for consideration in 2ED at this time

nigelmegitt:
> what can be considered a response to the normal review process of a CR

This is predicated on the idea that the only valid comments are those made to this specific version of the document. No such qualification is made in the SOTD where the invitation to comment is provided.

spec/images/sources/example-complex-layout.ttml
```diff
@@ -0,0 +1,45 @@
...
+                tts:fontSize="6.7rh"/>
+            <style xml:id="spanStyle"
+                tts:color="white"
+                tts:backgroundColor="black"
+                tts:fontFamily="proportionalSansSerif"/>
+        </styling>
+        <layout xmlns:tts="http://www.w3.org/ns/ttml#styling">
+            <region xml:id="topCenterArea"
+                tts:origin="10% 10%"
+                tts:position="center top 10% "
```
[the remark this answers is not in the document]
nigelmegitt: Fixed in 6154c1e.

spec/ttml2.xml
```diff
@@ -523,6 +537,108 @@ elements as they appear in the content hierarchy.</p>
 </tr>
 </tbody>
 </table>
+<p>The location of content spatially is independent of its timing.
```
palemieux, line 540: I suggest something less abstract, along the lines of: _TTML content consists of text that is flowed into regions over time._

nigelmegitt: On reflection, those concepts were already described in the text. Rather than trying to rewrite them with duplication I've reduced the amount of text and made it follow on a little better from the previous example. Let me know if that's an improvement @palemieux .

nigelmegitt:
@palemieux @cconcolato @skynavga please could you re-review given the changes from 12th May? 

@skynavga I could not derive any actions that would resolve your objection, other than not merging, but perhaps you could look again and consider reconsidering?

spec/ttml2.xml
```diff
@@ -427,6 +427,20 @@ the intended meaning. For instance, this example could be written to use
...
+The presentational effect would be the same if it were specified on the <el>div</el>
+element instead, or indeed on each individual <el>p</el> element. In typical usage
+the <att>region</att> would be specified at an appropriate part of the document
+hierarchy to allow for ease of expression while limiting verbosity.</p>
+<p>Note also that one effect of placing <att>region</att> on <el>body</el>
+is to prevent any descendant element from being presented in a <emph>different</emph>
+region. This is because any such descendant would be pruned from presentation,
+by the effect of the 
+<loc href="#procedure-construct-intermediate-document"><phrase role="strong">[construct intermediate document]</phrase></loc>
+procedure.</p>
```
skynavga, line 442: This explanation is not sufficiently specific to understand what part of the **[construct immediate document]** procedure is responsible for pruning.

spec/ttml2.xml
```diff
@@ -523,6 +537,112 @@ elements as they appear in the content hierarchy.</p>
 </tr>
 </tbody>
 </table>
+<p>The above examples demonstrate the temporal and lexical flow of content
+into a single region.
+This can be extended to multiple regions,
```
skynavga, line 542: This [what?] can be ...

spec/ttml2.xml
```diff
@@ -523,6 +537,112 @@ elements as they appear in the content hierarchy.</p>
...
+with different parts of the content flowed into each region
+at independent times.
+For example in <specref ref="ttml-example-complex-layout"/> there are two
```
skynavga, line 545: s/For example/For example,/

skynavga: s/in ... there/in ... there,/

spec/ttml2.xml
```diff
@@ -523,6 +537,112 @@ elements as they appear in the content hierarchy.</p>
...
+subtitle paragraphs, each positioned in a different region. The second one
+becomes visible before the first one is removed from view,
+and remains visible afterwards, so that there
+are progressively one, two and then one subtitles visible.</p>
```
skynavga, line 549: s/two/two,/

skynavga, line 549: s/subtitle/subtitle(s)/

spec/ttml2.xml
```diff
@@ -523,6 +537,112 @@ elements as they appear in the content hierarchy.</p>
...
+  &lt;/body&gt;
+&lt;/tt&gt;
+</eg>
+</td>
+</tr>
+</tbody>
+</table>
+<p>This example contains two subtitles that overlap in time.
+It produces three distinct intervals between 0.5s and 2s.
+In the document times are expressed in the hh:mm:ss format.
```
skynavga, line 606: s/document/document,/

spec/images/sources/example-complex-layout.ttml
```diff
@@ -0,0 +1,45 @@
...
+                <span style="spanStyle">I shall ask our Physiological Lecturer why<br/>
+                he never gave us that exquisite Theory!</span>
+            </p>
+            <p xml:id="subtitle2" begin="00:00:01" end="00:00:02" region="bottomLeftArea" style="pStyle">
+                <span style="spanStyle">I'd give something to be present<br/>
+                when the question is asked!</span>
+            </p>
+        </div>    
+    </body>
+</tt>
```
skynavga, line 45: Remove this as source file from repository. We don't store other examples as full TTML (XML) source files.

spec/ttml2.xml
```diff
@@ -523,6 +537,112 @@ elements as they appear in the content hierarchy.</p>
...
+Each region and paragraph is specified to have a larger size than its contents,
+and is shown with a contrasting background color
+so that its position and the position of its descendants are more clearly visible in this example.
```
skynavga, line 609: s/ in this example//

spec/ttml2.xml
```diff
@@ -523,6 +537,112 @@ elements as they appear in the content hierarchy.</p>
...
+<caption>Subtitle 1 &ndash; Time Interval [0.5, 1.0)</caption>
+<tbody>
+<tr>
+<td><graphic source="images/example-complex-layout_0.5.png" alt="Complex Layout Subtitle 1" css="width: 790px;"/></td>
+</tr>
+</tbody>
+</table>
+<p>In the first interval, 0.5 to 1.0 seconds, two lines are presented, aligned to the before edge of the top center region.
+The unoccupied part of the region is therefore visible as purple color <emph>beneath</emph> the text.
+Likewise the unoccupied part of the p is visible as cyan color around the black background showing the space occupied by the span.</p>
```
skynavga, line 623: s/Likewise/Likewise,/
s/of the p/of the p element/
s/the span/the span element/

spec/ttml2.xml
```diff
@@ -523,6 +537,112 @@ elements as they appear in the content hierarchy.</p>
...
+<table id="ttml-example-complex-layout-3" role="example-images">
+<caption>Subtitle 2 &ndash; Time Interval [1.0, 1.5)</caption>
+<tbody>
+<tr>
+<td><graphic source="images/example-complex-layout_1.5.png" alt="Complex Layout Subtitle 3" css="width: 790px;"/></td>
+</tr>
+</tbody>
+</table>
+<p>In the third interval, 1.5 to 2.0 seconds, the first subtitle is no longer
+presented.</p>
```
skynavga, line 645: Please introduce two new sub-divisions to organize the two examples, entitled *Single Region Example* and *Multiple Region Example*.

spec/images/sources/example-complex-layout.ttml
nigelmegitt: In fact we do store at least one other source for content generated for the spec: the `design` folder contains the source for the media timing image. I think we should as a rule try to store the sources. It's a real pain going back to old examples and not having the sources available to modify them, because unless you're the author of the example and happen to have the source available somewhere, the effect is that it needs to be rewritten.

spec/ttml2.xml
nigelmegitt: @skynavga I'm not convinced that it needs to be more specific, in this introduction example section. I'm trying to avoid falling foul of your https://github.com/w3c/ttml2/pull/1207#issuecomment-616642944 :
> attempts to paraphrase a quite complicated piece of semantics in a introductory context where such details are intentionally excluded and doing so would only serve to confuse the reader

However I will change "by the effect of the [construct intermediate document] procedure" to "by _an_ effect..." to avoid giving the impression that this is the only thing that procedure does.

nigelmegitt: Slight change of mind: I think the wording "as a consequence of" works better than "by an effect of"

spec/ttml2.xml
nigelmegitt: Aha, the "Oxford comma"! https://w3c.github.io/manual-of-style/#Grammar suggests having access to one of two grammar references that are both behind a login/paywall, which is unfortunate. From checking other W3C documents I'll work on the basis that we are indeed expected to use it.

spec/ttml2.xml
nigelmegitt: Addressed in 21e49933e99f178edfdfc9ddbebffcadbe9652c9

spec/ttml2.xml
nigelmegitt: Addressed in 21e49933e99f178edfdfc9ddbebffcadbe9652c9

spec/ttml2.xml
nigelmegitt: Addressed in 21e49933e99f178edfdfc9ddbebffcadbe9652c9

nigelmegitt: Addressed in 21e49933e99f178edfdfc9ddbebffcadbe9652c9

spec/ttml2.xml
nigelmegitt: Addressed in 21e49933e99f178edfdfc9ddbebffcadbe9652c9

spec/ttml2.xml
nigelmegitt: Addressed in 21e49933e99f178edfdfc9ddbebffcadbe9652c9

spec/ttml2.xml
nigelmegitt: Addressed in 21e49933e99f178edfdfc9ddbebffcadbe9652c9

spec/ttml2.xml
nigelmegitt: Addressed in 21e49933e99f178edfdfc9ddbebffcadbe9652c9

spec/images/sources/example-complex-layout.ttml
skynavga: ok

PR #1214, opened 2020-11-12
Remove applicability of tts:direction on <p>

### LINKED ISSUE #1211: Interaction between tts:writingMode and tts:direction on <p>
TTML2 specifies:

_When applied to a p element, the computed value of `tts:direction` explicitly establishes the paragraph level as specified by [UAX9], §4.3, Higher Level Protocol HL1._

It is unclear how this interacts with `tts:writingMode`. Specifically:

- are there any cases where `tts:direction` and `tts:writingMode` should conflict on `<p>`? In other words, does specifying `tts:direction="ltr"` ever make sense if `tts:writingDirection="rl`? CSS does not allow this inconsistency.

- what does `tts:direction` mean when the IPD is vertical? [UAX #9, clause 6.2](http://www.unicode.org/reports/tr9/tr9-37.html#Vertical_Text) implies that `tts:direction` should affect character rotation, but, again, this is ignored by CSS.

- the [Direction003.html test case](https://github.com/w3c/imsc-tests/blob/master/imsc1/ttml/direction/Direction003.ttml) implies that specifying `tts:direction="rtl"` on `<p>` changes the IPD.

- XSL specifies that `If the "direction" property is explicitly specified on the same formatting object the value of the "direction" property will override the inline-progression-direction set by the "writing-mode".`

I suggest considering the following clarification:

_when specified on `<p>`,  `tts:direction` sets the inline progression direction, overriding that set by `tts:writingDirection`. In the case where the block progression direction is horizontal, `tts:direction="ltr"` specifies a top-to-bottom inline progression direction whereas `tts:direction="rtl"` specifies a bottom-to-top inline progression direction._

### LINKED ISSUE #1212: Clarify application of unicodeBidi and direction on p
Arising from #1211 clarify the application of unicodeBidi and direction on the p element (only).

Suggest matching [CSS text direction](https://www.w3.org/TR/css-writing-modes-3/#text-direction).

* direction sets the paragraph embedding level
* unicodeBidi values `plaintext` and `bidi-override` have specific effects
* other values have no specific effect

### LINKED ISSUE #1213: make direction on p set the inline progression direction
Arising from #1211 change `tts:direction` as applied to `p` elements so that it additionally sets the inline progression direction, and overrides the inline progression direction set by `tts:writingMode`, if set.

This change makes TTML match the behaviour of XSL-FO and CSS.

palemieux:
Closes #1211 
Closes #1213 
Closes #1212

The _paragraph embedding level_ is determined by the `ipd`, instead of being determined by `tts:direction` as applied to `p`. This is compatible with CSS since, in CSS, `direction` sets the _paragraph embedding level_ and the combination of `direction` and `writing-mode` maps to `tts:writingMode`.

The effect of `tts:unicodeBidi` on `p` is clarified to match CSS.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -9787,8 +9787,8 @@ background color of the <loc href="#terms-root-container-region">root container 
 </div3>
 <div3 id="style-attribute-direction">
 <head>tts:direction</head>
-<p>The <att>tts:direction</att> attribute is used to specify a style property that, depending on the context of use,
-determines (1) the bidirectional paragraph level, or (2) the directionality of a bidirectional embedding or override,
+<p>The <att>tts:direction</att> attribute is used to specify a style property that 
+determines the directionality of a bidirectional embedding or override,
 about which see <bibref ref="uax9"/>.</p>
 <table id="style-property-details-direction" role="common">
 <col css="width: 25%;"/>
@@ -9810,7 +9810,6 @@ about which see <bibref ref="uax9"/>.</p>
 <tr>
 <td><emph>Applies to:</emph></td>
 <td>
-<loc href="#content-vocabulary-p"><el>p</el></loc>,
 <loc href="#content-vocabulary-span"><el>span</el></loc>
 </td>
 </tr>
@@ -9832,11 +9831,10 @@ about which see <bibref ref="uax9"/>.</p>
 </tr>
 </tbody>
 </table>
-<p>When applied to a <el>p</el> element, the computed value of this property explicitly establishes the
-<emph>paragraph level</emph> as specified by <bibref ref="uax9"/>, &sect;4.3, Higher Level Protocol <phrase role="strong">HL1</phrase>.</p>
-<p>When applied to a <el>span</el> element (or <loc href="#terms-anonymous-span">anonymous span</loc>), the computed value of this property, in combination with the computed value
-of the <att>tts:unicodeBidi</att> style property, determines the directionality of a bidirectional embedding or override as specified by
-<bibref ref="uax9"/>, &sect;4.3, Higher Level Protocol <phrase role="strong">HL3</phrase>.</p>
+<p>The computed value of this property, in combination with the computed value
+of the <att>tts:unicodeBidi</att> style property, determines the directionality of a bidirectional
+embedding or override as specified by <bibref ref="uax9"/>, &sect;4.3, Higher Level Protocol
+<phrase role="strong">HL3</phrase>.</p>
 <p>If a computed value of the property associated with this attribute is not supported,
 then a <loc href="#terms-presentation-processor">presentation processor</loc> must use the value <code>ltr</code>.</p>
 <p>The <att>tts:direction</att> style is illustrated by the following example.</p>
@@ -14176,6 +14174,9 @@ the Unicode bidirectional algorithm.</p>
 </table>
 <p>If a computed value of the property associated with this attribute is not supported,
 then a <loc href="#terms-presentation-processor">presentation processor</loc> must use the value <code>normal</code>.</p>
+<note role="elaboration">
+<p>The values <code>"normal"</code> and <code>"embed"</code> have no effect on <code>p</code>.</p>
+</note>
 <p>The <att>tts:unicodeBidi</att> style is illustrated by the following example.</p>
 <table id="style-attribute-unicodeBidi-example-1" role="example">
 <caption>Example Fragment &ndash; Unicode Bidirectionality</caption>
@@ -14465,6 +14466,9 @@ stacking block and inline areas within a region area.</p>
 </tr>
 </tbody>
 </table>
+<p>
+</p>
+<p>The <loc href="#terms-inline-progression-direction">inline progression direction</loc> determined by the computed value of this property explicitly establishes the <emph>paragraph level</emph> as specified by <bibref ref="uax9"/>, &sect;4.3, Higher Level Protocol <phrase role="strong">HL1</phrase> for all descendent <code>p</code> elements.</p>
 <p>If a computed value of the property associated with this attribute is not supported,
 then a <loc href="#terms-presentation-processor">presentation processor</loc> must use the value <code>lrtb</code>.</p>
 <note role="elaboration">
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -9835,6 +9835,9 @@ about which see <bibref ref="uax9"/>.</p>
 of the <att>tts:unicodeBidi</att> style property, determines the directionality of a bidirectional
 embedding or override as specified by <bibref ref="uax9"/>, &sect;4.3, Higher Level Protocol
 <phrase role="strong">HL3</phrase>.</p>
+<note role="elaboration">
+<p>In earlier editions of this specification, the <att>tts:direction</att> property applied to the <code>p</code> element. This use is now <phrase role="deprecated">deprecated</phrase> in absence of definitive semantics. Authors are encouraged to avoid specifiying <att>tts:direction</att> on a <code>p</code> element, which can result in different rendering results depending on the implementation.</p>
+</note>
 <p>If a computed value of the property associated with this attribute is not supported,
 then a <loc href="#terms-presentation-processor">presentation processor</loc> must use the value <code>ltr</code>.</p>
 <p>The <att>tts:direction</att> style is illustrated by the following example.</p>
@@ -14468,7 +14471,7 @@ stacking block and inline areas within a region area.</p>
 </table>
 <p>
 </p>
-<p>The <loc href="#terms-inline-progression-direction">inline progression direction</loc> determined by the computed value of this property explicitly establishes the <emph>paragraph level</emph> as specified by <bibref ref="uax9"/>, &sect;4.3, Higher Level Protocol <phrase role="strong">HL1</phrase> for all descendent <code>p</code> elements.</p>
+<p>The <loc href="#terms-inline-progression-direction">inline progression direction</loc> determined by the computed value of this property explicitly establishes the <emph>paragraph level</emph> as specified by <bibref ref="uax9"/>, &sect;4.3, Higher Level Protocol <phrase role="strong">HL1</phrase>.</p>
 <p>If a computed value of the property associated with this attribute is not supported,
 then a <loc href="#terms-presentation-processor">presentation processor</loc> must use the value <code>lrtb</code>.</p>
 <note role="elaboration">
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -14153,7 +14153,6 @@ the Unicode bidirectional algorithm.</p>
 <tr>
 <td><emph>Applies to:</emph></td>
 <td>
-<loc href="#content-vocabulary-p"><el>p</el></loc>,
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -14465,6 +14466,9 @@ stacking block and inline areas within a region area.</p>
 </tr>
 </tbody>
 </table>
+<p>
+</p>
+<p>The <loc href="#terms-inline-progression-direction">inline progression direction</loc> determined by the computed value of this property explicitly establishes the <emph>paragraph level</emph> as specified by <bibref ref="uax9"/>, &sect;4.3, Higher Level Protocol <phrase role="strong">HL1</phrase> for all descendent <code>p</code> elements.</p>
```
tairt, line 14471:
```suggestion
<p>The <loc href="#terms-inline-progression-direction">inline progression direction</loc> determined by the computed value of this property explicitly establishes the <emph>paragraph level</emph> as specified by <bibref ref="uax9"/>, &sect;4.3, Higher Level Protocol <phrase role="strong">HL1</phrase>.</p>
```
It may be better to take out the part where the application of the paragraph level to the TTML `<p>` elements is mentioned. This keeps it in line with the original text from chapter [10.2.12](https://www.w3.org/TR/2018/REC-ttml2-20181108/#style-attribute-direction). Also as I undestand it the paragraph level is not directly tied to TTML markup but is applied in a later stage during UAX9 processing.

spec/ttml2.xml
```diff
@@ -14176,6 +14174,9 @@ the Unicode bidirectional algorithm.</p>
 </table>
 <p>If a computed value of the property associated with this attribute is not supported,
 then a <loc href="#terms-presentation-processor">presentation processor</loc> must use the value <code>normal</code>.</p>
+<note role="elaboration">
+<p>The values <code>"normal"</code> and <code>"embed"</code> have no effect on <code>p</code>.</p>
+</note>
```
tairt, line 14179: I wonder why this note is necessary because the application of `tts:unicodeBidi` to `<p>` was already deleted in a change above.

spec/ttml2.xml
```diff
@@ -14176,6 +14174,9 @@ the Unicode bidirectional algorithm.</p>
 </table>
 <p>If a computed value of the property associated with this attribute is not supported,
 then a <loc href="#terms-presentation-processor">presentation processor</loc> must use the value <code>normal</code>.</p>
```
tairt, line 14176: It may be good to say:
* that in the previous version `tts:direction` was also applicable to the `<p>` element,
* specifiying `tts:direction` on `<p>` may lead to different rendering results depending on implementation,
* it is considered to re-introduce the application of `tts:direction`  to `<p>` with a clear, determinisic rendering expectation in a later version.

spec/ttml2.xml
palemieux: The intent was for tts:unicodeBidi to remain applicable to <p>

tairt: My conclusion from the previous discussions has been that `tts:unicodeBidi` is not applicable to `<p>` (even if `tts:direction` would still be applicable to `<p>`). 

In my previous comment, I have confused the removal of applicability of `tts:direction` to `<p>` with the removal of applicability of `tts:unicodeBidi` to `<p>`. As I see the latter has not been removed, but should.

nigelmegitt (commented):
Notes to consider (partly for myself, but sharing them in case others have views):
* Should we change the semantic derivation for `tts:direction` to be CSS 3 Writing Modes?
* #1213 doesn't say to remove the applicability of `tts:direction` to `p`, but that's what this PR does.
* #1213 says to make `tts:direction` on `p` set the inline progression direction in addition to the paragraph embedding level, but this PR does not do that.
* #1212 says to remove the applicability of `tts:unicodeBidi` from `p`, but this PR does not do that.

At the moment, it looks to me, unless I've misunderstood, that the PR does not match the requested changes from the issues it is intended to close. I look forward to discussing.

spec/ttml2.xml
palemieux: > My conclusion from the previous discussions has been that tts:unicodeBidi is not applicable to <p> (even if tts:direction would still be applicable to <p>).

Can you provide a link to that discussion?

palemieux:
> At the moment, it looks to me, unless I've misunderstood, that the PR does not match the requested changes from the issues it is intended to close. I look forward to discussing.

Yes, the PR attempts to find a middle ground to address the concerns with changing the semantics of `tts:direction` on `p` to influence the IPD.

cconcolato:
Following our discussion I found one example in our catalog that uses dir on p:
```
<p begin="10844167t" end="29195834t" region="bottom" tts:direction="rtl" tts:unicodeBidi="embed" xml:id="ID_57d7c57f-ef91-42fd-8f32-0d1f237b5131">
  <span style="span">- תוכן מקורי של NETFLIX -</span>
</p>
```
the region is defined as 
```
<region tts:backgroundColor="transparent" tts:displayAlign="after" tts:extent="80.000% 80.000%" tts:origin="10.000% 10.000%" tts:textAlign="center" xml:id="bottom"/>
```
writing-mode is not used at all in the document

palemieux:
> <p begin="10844167t" end="29195834t" region="bottom" tts:direction="rtl" tts:unicodeBidi="embed" xml:id="ID_57d7c57f-ef91-42fd-8f32-0d1f237b5131">

@cconcolato What is the expected rendering?

spec/ttml2.xml
palemieux: See https://github.com/w3c/ttml2/pull/1214/commits/18e8bf0ca7552d4ccc9774670372bb87aaac012a

spec/ttml2.xml
palemieux: See https://github.com/w3c/ttml2/pull/1214/commits/18e8bf0ca7552d4ccc9774670372bb87aaac012a

tairt: Looks good! Thanks!

spec/ttml2.xml
tairt: Exactly. Very good wording. Thanks!

spec/ttml2.xml
tairt: It follows the argumentation of @skynavga e.g. in the last part of his [comment](https://github.com/w3c/ttml2/issues/1211#issuecomment-705279511) from 2020-10-08 and at the beginning of the [TTWG discussion](https://github.com/w3c/ttml2/issues/1211#issuecomment-705696362)  on 2020-10-08. There was also an [issue](https://github.com/w3c/ttml2/issues/1212) opened where title indicates to remote application of unicodeBidi to p. There has been no final consensus on this question, but following the argumentation it make sense to remove the application to p. 

The question I have is also: What are the reasons to keep unicodeBidi when application of direction is removed? unicodeBidi depends on a direction counterpart...

A solution of the unicodeBidi question may be resolved in another PR. But I would then not mix it with this PR and therefore not add additional text to chapter 10.2.48.

palemieux: @tairt `tts:unicodeBidi` is no longer applicable to `p` per your comment above. I however think we should handle `tts:direction` and `tts:unicodeBidi` in the same PR, since, as you state above, they are intimately related.

tairt: @palemieux Yes, makes sense. Thanks for the edit. Looks good to me!

tairt (approved):

spec/ttml2.xml
```diff
@@ -14465,6 +14468,9 @@ stacking block and inline areas within a region area.</p>
 </tr>
 </tbody>
 </table>
+<p>
+</p>
+<p>The <loc href="#terms-inline-progression-direction">inline progression direction</loc> determined by the computed value of this property explicitly establishes the <emph>paragraph level</emph> as specified by <bibref ref="uax9"/>, &sect;4.3, Higher Level Protocol <phrase role="strong">HL1</phrase>.</p>
```
skynavga, line 14473: This language is semantically inconsistent with §10.2.12.1.

spec/ttml2.xml
```diff
@@ -9832,11 +9831,13 @@ about which see <bibref ref="uax9"/>.</p>
...
-<p>When applied to a <el>span</el> element (or <loc href="#terms-anonymous-span">anonymous span</loc>), the computed value of this property, in combination with the computed value
-of the <att>tts:unicodeBidi</att> style property, determines the directionality of a bidirectional embedding or override as specified by
-<bibref ref="uax9"/>, &sect;4.3, Higher Level Protocol <phrase role="strong">HL3</phrase>.</p>
+<p>The computed value of this property, in combination with the computed value
+of the <att>tts:unicodeBidi</att> style property, determines the directionality of a bidirectional
+embedding or override as specified by <bibref ref="uax9"/>, &sect;4.3, Higher Level Protocol
+<phrase role="strong">HL3</phrase>.</p>
+<note role="elaboration">
+<p>In earlier editions of this specification, the <att>tts:direction</att> property applied to the <code>p</code> element. This use is now <phrase role="deprecated">deprecated</phrase> in absence of definitive semantics. Authors are encouraged to avoid specifiying <att>tts:direction</att> on a <code>p</code> element, which can result in different rendering results depending on the implementation.</p>
```
skynavga, line 9839: It is inappropriate to deprecate functionality that has been deployed since TTML1 1st edition without providing a mechanism for an author to explicitly declare which behavior is desired (either the pre-existing behavior or the newly proposed, non-backwards compatible behavior). In other words, this change should not be treated as a candidate for deprecation.

spec/ttml2.xml
```diff
@@ -14176,6 +14176,9 @@ the Unicode bidirectional algorithm.</p>
 </table>
 <p>If a computed value of the property associated with this attribute is not supported,
 then a <loc href="#terms-presentation-processor">presentation processor</loc> must use the value <code>normal</code>.</p>
+<note role="elaboration">
+<p>In earlier editions of this specification, the <att>tts:unicodeBidi</att> property applied to the <code>p</code> element. This use is now <phrase role="deprecated">deprecated</phrase> in absence of definitive semantics. Authors are encouraged to avoid specifiying <att>tts:unicodeBidi</att> on a <code>p</code> element, which can result in different rendering results depending on the implementation.</p>
```
skynavga, line 14180: See above comment about deprecation.

skynavga (changes requested):

skynavga:
Re: https://github.com/w3c/ttml2/pull/1214#issue-519627186, please note that the abbreviation _ipr_ does not mean _inline progression direction_ in TTML, it means _inline progression dimension_; therefore, to avoid confusion, it is best to spell out _inline progression direction_.

skynavga:
@cconcolato Re: https://github.com/w3c/ttml2/pull/1214#issuecomment-726243664, I would expect the rendering to be such that the region is left-right, top-to-bottom, containing a horizontally center aligned paragraph that has its content effectively wrapped in a RLE/PDF Unicode Bidirectional Controls pair, which content consists of the text

```- תוכן מקורי של NETFLIX -```

The Bidi reordered text should appear as follows, and translates as "Original Content of NETFLIX".

```- NETFLIX תוכן מקורי של -```

spec/ttml2.xml
palemieux: What is the semantic inconsistence?

PR #1216, opened 2021-01-23
Clarify that certain tts:font* properties apply to tt:p (#1215).

### LINKED ISSUE #1215: Permit implementations to use fontSize and fontFamily when computing line height
In HTML/CSS implementations, even if the `line-height` property on a `p` element is set to a pixel based length, the line height that results depends on the used vales of `font-size` and `font-family` on that `p` element. The `line-height` is taken as a minimum (as per the CSS spec) and the font metrics are taken into account to increase the size if the implementation sees fit. 

See https://codepen.io/nigelmegitt/pen/qBaJZOv for an example of this. 

However in TTML, the equivalent style attributes, `tts:fontSize` and `tts:fontFamily` clearly do not _apply_ to `p` elements, so even if the author specifies the font size and family on the `p` element, implementations are permitted (required, arguably) to ignore them there (for rendering purposes) and use them only as the basis for inheritance down to the `span` level, where it does apply.

This can be surprising for content authors, who find that depending on context, the line height may not match what they specify in the TTML.

The specification for computing actual line height allows some freedom, so this is not "broken" per se. However we would allow authors a greater degree of control, and more closely match existing CSS-based implementations, if we were to allow presentation processors to _apply_ `tts:fontSize` and `tts:fontFamily` to `p` elements in TTML solely for the purpose of computing line height.

For example see also sandflow/imscJS#201 which proposes passing these attributes into the style attribute of the HTML `p` element when rendering an ISD, to achieve the expected line spacing.

Concretely, this issue proposes to add a special semantic to `tts:fontSize` and `tts:fontFamily` stating that their computed values MAY be used when computing line height on `p` elements, and to reference that from the section that defines `tts:lineHeight` when the specified value is not `"normal"`.

skynavga:
Closes #1215.


nigelmegitt (approved):
Thanks @skynavga this looks good to me.

skynavga:
Merge per request from @nigelmegitt in order to advance to new CR.

PR #1221, opened 2021-02-19
Fix link (#1200).

### LINKED ISSUE #1200: Link to fontSelectionStrategy from lineHeight algorithm points to fontStyle.
The §10.2.27 tts:lineHeight algorithm for when the value is `normal`  includes `tts:fontSelectionStrategy` but clicking on it takes the reader to `#style-attribute-fontStyle` instead of `#style-attribute-fontSelectionStrategy`.

skynavga:
Closes #1200.


nigelmegitt (approved):
Trivial editorial fix - can be merged immediately.

PR #1224, opened 2021-02-25
Fix pubrules

himorin:
This will fix issues from pubrules, except for ones we need to update status.xml


skynavga:
Merge branch 'fix-pubrules' into issue-1220-cr2-publishing-package. Please use latter branch for any further proposed fixes.

PR #1228, opened 2021-03-23
updated examples using https://

himorin:
Updated links in examples with using https://

* haven't touch any namespaces
* will file another PR for references


nigelmegitt (approved):
Looks good to me

skynavga:
I would note that these URLs are not actually links in the document, so it isn't necessary (by policy) to make these changes. Notwithstanding, I'm going ahead with accepting and merging.

nigelmegitt:
Thank you @skynavga - as I understand it the reason for changing the URLs in the examples is not to deal with the way pages are served, or HSTS say, but to demonstrate by example in the document that we are promoting the use of secure protocols in TTML documents.

PR #1233, opened 2021-05-25
Fix #1232 by clarifying the [resolve timing] procedure

### LINKED ISSUE #1232: Clarify if the first ISD must/may be constructed when empty
This issue arises because there's discussion in MPEG regarding 14496-30 about defining the correct semantics of processing sequences of documents each in a wrapper, where the draft text explains it in terms of sequences of ISDs. @cconcolato @mikedo 

The specification for [[resolve timing]](https://www.w3.org/TR/ttml2/#procedure-resolve-timing) step 2 is:
> divide the active time duration of the current document instance into an ordered sequence of time coordinates `{T0, T1, T2, ...}` where, at each time coordinate `Ti`, some element becomes temporally active or inactive.

The question I have is: in the case that the first element that becomes temporally active does so at a time `T1` where `T1 > 0`, does this imply or require that there is a first empty ISD that begins at time 0 and ends at `T1`?

Taking it literally, no element becomes temporally active or inactive at time 0, so arguably no ISD should be generated. However, the draft text I've had sight of implies an expectation that there is such an "empty document", and it is possible that implementers would like to be able to rely on this behaviour.

I'd like to discover what our expectation and understanding is here, and if there is implementation experience to guide us, so that we can state this more clearly in the specification, and feed this information to others who may depend on it.

nigelmegitt:
Fix #1232 by clarifying that the active time duration of the current document instance _is_ the root temporal extent, and adding an elaboration note about the permitted range of ISD time coordinates.

___

This change can be previewed at https://rawgit.com/w3c/ttml2/issue-1232-clarify-isd-construction-build/index.html#procedure-resolve-timing

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -25627,7 +25627,7 @@ times in the absence of other external information.</caption>
 <bibref ref="smil3"/>, the implicit duration of the <el>body</el>
 element is identical to its computed duration which,
 in the absence of any external constraints, 
-is also the <loc href="#terms-root-temporal-extent">Root Temporal
+is also the duration of the <loc href="#terms-root-temporal-extent">Root Temporal
 Extent</loc>.  However if
 the <loc href="#terms-document-processing-context">document processing
 context</loc> specifies a range of applicable media times, those may modify
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -25630,7 +25630,7 @@ in the absence of any external constraints,
 is also the <loc href="#terms-root-temporal-extent">Root Temporal
 Extent</loc>.  However if
 the <loc href="#terms-document-processing-context">document processing
-context</loc> specifies a range of applicable media times, those modify
+context</loc> specifies a range of applicable media times, those may modify
 the resolved begin and end times and therefore
 the <loc href="#terms-root-temporal-extent">Root Temporal
 Extent</loc>.</p>
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -18002,11 +18002,27 @@ otherwise, if not invoked at this time, then resolve the implicit duration for e
 implicit duration resolved had this procedure been invoked;</p>
 </item>
 <item>
-<p>divide the active time duration of the current <loc href="#terms-document-instance">document instance</loc> into an ordered sequence
+<p>divide the active time duration, i.e. the <loc href="#terms-root-temporal-extent">root temporal extent,</loc>
+of the current <loc href="#terms-document-instance">document instance</loc> into an ordered sequence
 of time coordinates <code>{T<sub>0</sub>, T<sub>1</sub>, T<sub>2</sub>, ...}</code> where, at each time coordinate <code>T<sub>i</sub></code>,
 some element becomes temporally active or inactive.</p>
 </item>
 </olist>
+<note role="elaboration">
+<p>The result of this procedure is a set of time coordinates: the first may be greater than zero
+and the last may be a finite value.</p>
+<p>However there may be a benefit to some applications if they require that
+the first time coordinate is zero,
+or that the last time coordinate is infinite,
+or both, in other words, if they extend the root temporal extent.</p>
+<p>Similarly, the <loc href="#terms-document-processing-context">document processing context</loc>
+may constrain the range of time coordinates to fit within a defined range,
+for example by clamping time coordinates outside that range to the value of the range,
+and removing any resulting duplicates.</p>
+<p>This specification requires that the last time coordinate is infinite only when
+there is no end to the document's active duration and
+there is no document processing context constraint that prevents it.</p>
+</note>
 </def>
 </gitem>
 <gitem id="procedure-construct-anonymous-spans">
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -1904,7 +1904,8 @@ content rectangles are coterminous.</p>
 <gitem id="terms-root-temporal-extent">
 <label>[root temporal extent]</label>
 <def>
-<p>The temporal extent (interval) defined by the temporal beginning and ending of a <loc href="#terms-document-instance">document instance</loc>
+<p>The temporal extent (interval) defined by the temporal beginning and ending of a <loc href="#terms-document-instance">document instance</loc>,
+i.e., the time interval over which a <loc href="#terms-document-instance">document instance</loc> is active,
 in relationship with some external application or presentation context.</p>
 </def>
 </gitem>
@@ -6125,7 +6126,7 @@ zero or one <el>body</el> element.</p>
 </tr>
 </tbody>
 </table>
-<p>The <loc href="#terms-root-temporal-extent">root temporal extent</loc>, i.e., the time interval over which a <loc href="#terms-document-instance">document instance</loc> is active, has an implicit duration that is equal to the
+<p>The <loc href="#terms-root-temporal-extent">root temporal extent</loc> has an implicit duration that is equal to the
 implicit duration of the <el>body</el> element of the document, if the <el>body</el> element is present, or zero, if the <el>body</el> element is absent.</p>
 <p>If the <loc href="#style-attribute-extent"><att>tts:extent</att></loc> attribute is specified on the <el>tt</el>
 element, then it must adhere to <specref ref="style-attribute-extent"/>, in which case it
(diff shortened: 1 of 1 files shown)


spec/ttml2.xml
```diff
@@ -18002,11 +18002,27 @@ otherwise, if not invoked at this time, then resolve the implicit duration for e
 implicit duration resolved had this procedure been invoked;</p>
 </item>
 <item>
-<p>divide the active time duration of the current <loc href="#terms-document-instance">document instance</loc> into an ordered sequence
+<p>divide the active time duration, i.e. the <loc href="#terms-root-temporal-extent">root temporal extent,</loc>
+of the current <loc href="#terms-document-instance">document instance</loc> into an ordered sequence
```
palemieux, line 18006: Adding `root temporal extent` here introduces more uncertainty than it resolves.

spec/ttml2.xml
```diff
@@ -18002,11 +18002,27 @@ otherwise, if not invoked at this time, then resolve the implicit duration for e
...
+or that the last time coordinate is infinite,
+or both, in other words, if they extend the root temporal extent.</p>
+<p>Similarly, the <loc href="#terms-document-processing-context">document processing context</loc>
+may constrain the range of time coordinates to fit within a defined range,
+for example by clamping time coordinates outside that range to the value of the range,
+and removing any resulting duplicates.</p>
+<p>This specification requires that the last time coordinate is infinite only when
+there is no end to the document's active duration and
+there is no document processing context constraint that prevents it.</p>
+</note>
```
palemieux, line 18025: What about simply stating _for any given time t is there always exactly one ISD where t falls within that ISD's interval_, as suggested at https://github.com/w3c/ttml2/issues/1232#issuecomment-841360458

palemieux (changes requested):

spec/ttml2.xml
nigelmegitt: I think it clarifies the existing uncertainty. Previously we had an undefined term, "active time duration", now we have a defined term. If you think the definition isn't clear or helpful, that's a separate issue. But tying them together seems like it makes sense to me.

spec/ttml2.xml
nigelmegitt: The conclusion I came to after the discussion was that this may be desirable in applications of TTML, but a) would be a normative change and b) is not necessarily desirable for every application.

spec/ttml2.xml
palemieux: Is t=0 always included in the _root temporal extent_?

spec/ttml2.xml
palemieux: Stating that an ISD exists for every time t in [0, inf) does not compel an implementation to generate such an ISD, so there is no change in requirements, right?

spec/ttml2.xml
nigelmegitt: No, not necessarily. For example see https://www.w3.org/TR/2021/CR-ttml2-20210309/#semantics-media-document-timing

spec/ttml2.xml
nigelmegitt: I believe it would compel in that way. The [[construct intermediate document]](https://www.w3.org/TR/2021/CR-ttml2-20210309/#procedure-construct-intermediate-document) procedure creates an ISD for every time interval obtained from [resolve timing]. 

If we were to make [resolve timing] generate a sequence that always begins with 0 and ends with infinity, then we would be requiring ISDs to be generated. That's a normative change, since right now TTML processors are _not_ required to do that.

spec/ttml2.xml
palemieux: The current text does not state that the time coordinates are limited to the _root temporal extent_. The proposed text introduces a new restriction, for which I do not see a clear motivation.

nigelmegitt: What restriction is introduced? It seems to me that "active time duration" is genuinely synonymous with "root temporal extent" and therefore this is editorial. Also, the ability to allow the document processing context to affect the used values of the root temporal extent means that this adjusted wording remains very open for implementations to behave as they need to.

spec/ttml2.xml
palemieux: > If we were to make [resolve timing] generate a sequence that always begins with 0 and ends with infinity, then we would be requiring ISDs to be generated. 

I propose not changing the algorithm, but merely stating that there is an ISD defined for every time t in [0, inf). Every application can choose what ISDs to generate, including none.

nigelmegitt: > Every application can choose what ISDs to generate, including none.

I think we're getting to the same endpoint but with different spec text changes. In effect, the edits I'm proposing already do state that.

spec/ttml2.xml
```diff
@@ -18002,11 +18002,27 @@ otherwise, if not invoked at this time, then resolve the implicit duration for e
 implicit duration resolved had this procedure been invoked;</p>
 </item>
 <item>
-<p>divide the active time duration of the current <loc href="#terms-document-instance">document instance</loc> into an ordered sequence
+<p>divide the active time duration, i.e. the <loc href="#terms-root-temporal-extent">root temporal extent,</loc>
```
skynavga, line 18005: Please move the comma out of the new <loc>...</loc> element, i.e., put it immediately after the end tag.

spec/ttml2.xml
skynavga, line 18025: I'm reluctant to accept this elaboration since it may be viewed as defining implicit semantic constraint(s) that we have not completely fleshed out (or verified). I prefer removing this new text completely, and only making the first change shown above.

skynavga (changes requested):

spec/ttml2.xml
nigelmegitt: I thought we'd agreed all of these points, worded slightly differently, in the discussion of the issue #1232. However, there is certainly scope for revisiting the terminology we use in this area, e.g. is root temporal extent a duration or an interval, and what is it, and who gets to define it, and does a document get to define an interval at all, etc etc.

@palemieux and I discussed this during today's call, notes in https://github.com/w3c/ttml2/pull/1233#issuecomment-849751289

spec/ttml2.xml
nigelmegitt: If there's consensus to make this change then I'll tweak that; since it seems there may not be, I'll leave it for the time being.

spec/ttml2.xml
skynavga: I think we had conceptual agreement on these points. However, we did not discuss actually writing those points into the spec, and if we had discussed it, I would have opposed it for the reasons that both Pierre and I have raised. Namely, adding a clarification of this nature is potentially wrong and very risky IMO.

spec/ttml2.xml
skynavga: Note that adding this parenthetical is semantically equivalent to the language found in §8.1.1:

>The root temporal extent, i.e., the time interval over which a document instance is active, ...

We SHOULD make this change (modulo fixing the comma placement).

skynavga (commented):
Suggest changing to "may modify" or ", depending upon document processing context, may modify".

spec/ttml2.xml
skynavga: Still need that comma moved.

nigelmegitt: Sorry, I don't see a comma that needs to be moved - can you point me at the one you mean @skynavga ?

skynavga: Change "extent,`</loc>`" to "extent`</loc>`,"

nigelmegitt: That isn't in the latest commit, I think?

nigelmegitt: Just confirmed, it was removed in 46e3bea.

skynavga (approved):
Thanks. LGTM.

spec/ttml2.xml
```diff
@@ -25623,11 +25625,12 @@ times in the absence of other external information.</caption>
 </table>
 <p>In this example, using the terminology of
 <bibref ref="smil3"/>, the implicit duration of the <el>body</el>
-element is identical to its computed duration which is also
-the <loc href="#terms-root-temporal-extent">Root Temporal
+element is identical to its computed duration which,
+in the absence of any external constraints, 
+is also the <loc href="#terms-root-temporal-extent">Root Temporal
```
palemieux, line 25630:
```suggestion
is also the duration of the <loc href="#terms-root-temporal-extent">Root Temporal
```

palemieux (changes requested):
The Root Temporal extent has a begin and end, in addition to a duration.

spec/ttml2.xml
nigelmegitt: I've reviewed the usage of Root Temporal Extent and agree that the way it is defined is as an interval that has both a duration and a begin time, so I'll update accordingly. There may be one or two inconsistencies elsewhere around this that need ironing out too.

nigelmegitt: Addressed in fc9c6ac - I looked for any other inconsistencies and did not find any other changes needed.

palemieux (approved):

skynavga (approved):

PR #1238, opened 2021-07-14
migrated TravisCI to GitHub Actions

himorin:
for https://github.com/w3c/ttwg/issues/198


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,37 @@
...
+on: [ push ]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK
+        uses: actions/setup-java@v2
+        with:
+          java-version: '11'
```
nigelmegitt, line 11: Have you got this to work? I have to switch to 8 to get everything to run fine, because of the Java components that are no longer bundled with 11.

nigelmegitt: Oh wait, I can see that it has worked from the "Show all checks". This is Good! Thank you.

nigelmegitt (approved):
A good follow-up would be to post a "preview" link to the originating pull request pointing to the built result, but this is a good start worth merging now, I think.

himorin:
how about this?

nigelmegitt:
Thanks @himorin that's a good improvement - these could be separate pull requests, but I wonder if it's possible also to:
* put the preview link on the first comment on the pull request, rather than adding an inline comment;
* add a link to a served version of the specification built HTML itself so that it can be directly viewed;
* diff the build HTML with the version on the pull request target branch, store that, and link to that too, like PR-preview does;

himorin:
For first point, it's (not a bit) somehow difficult to do with current way of building action script...

For second and third, how about this?

nigelmegitt:
That's great! I think it might overload the diff service though: if there's a way to store the result and point to that directly then the sys team might be a bit happier, I'm guessing.

himorin:
> That's great! I think it might overload the diff service though: if there's a way to store the result and point to that directly then the sys team might be a bit happier, I'm guessing.

unfortunately, (it seems) htmlpreview script conflicts with htmldiff script as above.

skynavga:
@nigelmegitt could you do a final review?

nigelmegitt (approved):
I suspect that the Diff will always be a Diff against the gh-pages version, even if the pull request target branch is _not_ that one, but this is a very unusual use case for us.

For example:

* Pull Request _n_: source branch `issue-1234-fix-whatever`, target branch `gh-pages` - this is fine.
* Pull Request _m_: source branch `proposed_incremental_change_for_1234`, target branch `issue-1234-fix-whatever` - I suspect this will lead to strange results in the Diff, and maybe on merge too.

Just something to note; since it is a rare case, happy to approve. Thanks for this @himorin , it'll be a great improvement.

himorin:
@nigelmegitt @skynavga thank you for review comments. I haven't noticed about the case with target as non-gh-branch... it could be possible to build diff at such situation. But somehow difficult considering possible situations such as no ready made built files for target working branch (should not be, since we are operating automatic build for PRs raised against here - but it means non PR one will not have such build), or getting additional information for building these.... So, let me give up on such cases.

And, let me raise PRs to relevant repositories using TravisCI for building spec doc (and hopefully) shortly.

himorin:
I've found a serious issue while pushing to other targets. I'll raise new PR shortly.

PR #1242, opened 2021-08-05
Address ambiguity in 'tts:lineHeight' normal algorithm (#1241).

### LINKED ISSUE #1241: Clarify "the first font obtained" in tts:lineHeight normal algorithm.
Clarify the following ambiguity regarding the `tts:lineHeight="normal"` algorithm in [§10.2.27](https://www.w3.org/TR/2021/CR-ttml2-20210309/#style-attribute-lineHeight):

- [ ] no contextual context is specified by which means `tts:fontSelectionStrategy` could function as a mapping constraint

If possible, the resolution of this ambiguity should favor the CSS Fonts Module definition of [first available font](https://www.w3.org/TR/css-fonts-3/#character-map), in which case the use of `tts:fontSelectionStrategy` as a mapping constrain becomes moot.

Note that this ambiguity is present regardless of whether `tts:lineHeight` is applied to `p` or `span`.

skynavga:
Closes #1241.

### DIFF
--- a/spec/ttml2.xml
+++ b/spec/ttml2.xml
@@ -11520,7 +11520,25 @@ where this set of available fonts is constrained as needed to satisfy the comput
 <loc href="#style-attribute-fontWeight"><att>tts:fontWeight</att></loc>, and
 <loc href="#style-attribute-fontVariant"><att>tts:fontVariant</att></loc>,
 style properties of <emph>P</emph>.
-</p></item>
+</p>
+<note role="clarification">
+<p>For compatibility with the <loc href="https://www.w3.org/TR/css-fonts-3/#character-map">first available font</loc>
+defined by <bibref ref="css3-fonts"/>, it is intended that the phrase <emph>the first font obtained</emph>, as used in the previous text, refers to the
+first font whose <loc href="#embedded-content-value-unicode-range">&lt;unicode-range&gt;</loc> includes the U+0020 (space) character,
+where (1) the <loc href="#embedded-content-value-unicode-range">&lt;unicode-range&gt;</loc> of a built-in
+<loc href="#terms-font-resource">font resource</loc>, i.e., a <loc href="#terms-font-resource">font resource</loc> provided by the
+<loc href="#terms-presentation-processor">presentation processor</loc>, is a
+<loc href="#embedded-content-value-unicode-range">&lt;unicode-range&gt;</loc> that contains each character of the font's
+<loc href="https://www.w3.org/TR/css-fonts-3/#character-map">character map</loc>; and (2) the
+<loc href="#embedded-content-value-unicode-range">&lt;unicode-range&gt;</loc> of an embedded <loc href="#terms-font-resource">font resource</loc>,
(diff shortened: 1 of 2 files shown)


spec/ttml2.xml
```diff
@@ -11520,7 +11520,25 @@ where this set of available fonts is constrained as needed to satisfy the comput
 <loc href="#style-attribute-fontWeight"><att>tts:fontWeight</att></loc>, and
 <loc href="#style-attribute-fontVariant"><att>tts:fontVariant</att></loc>,
 style properties of <emph>P</emph>.
-</p></item>
+</p>
+<note role="clarification">
+<p>For compatibility with the <loc href="https://www.w3.org/TR/css-fonts-3/#character-map">first available font</loc>
+defined by <bibref ref="css3-fonts"/>, it is intended that the phrase <emph>the first font obtained</emph>, as used in the previous text, refers to the
+first font whose <loc href="#embedded-content-value-unicode-range">&lt;unicode-range&gt;</loc> includes the U+0020 (space) character,
```
nigelmegitt, line 11527: I don't recognise this intent as anything that TTWG discussed. However, I think it's obvious that the CSS algorithm needs to be considered a valid implementation.

Rather than duplicating the definition of the CSS algorithm, can we reference it instead, simplifying this text? Something along the lines of "This requirement is intended to be compatible with the CSS3 font algorithm for selecting the first font obtained; however if that algorithm is used, then it effectively overrides any semantic defined by `tts:fontSelectionStrategy` when determining line height."

skynavga: that won't work for a number of reasons, which I would be happy to explain in a ttwg call

nigelmegitt: The next TTWG call will be in September, so it would be worth sketching out the reasons here in order to make progress sooner.

skynavga: Primarily:
- CSS Font Module Level 3 (which TTML2 references) has slightly different wording that has been updated since in CSS Font Module Level 4; namely, "the first available font that would match the U+0020 (space) character" (CSS Fonts 3) vs "the first font whose unicode-range includes the U+0020 (space) character" (CSS Fonts 4); see also https://github.com/w3c/csswg-drafts/issues/4796;
- AFAICT neither CSS Font Module Level 3 nor 4 specifies what it means to evaluate the unicode-range of a system (built-in) font;
- neither CSS Font Module Level 3 nor 4 has our `<font/>` element, but instead has a `@font-face` rule, so we need language that refers to the unicode-range, i.e., our `@range` attribute (either specified or defaulted);

Bottom line is we can't reference CSS Font Module Level 3 for the language we need, and we still need to add language about the meaning of unicode-range, all together ending up with about the same (if not more) text in the note.

skynavga: For this issue, i.e., for finding the first font having a mapping for 0x0020 (SPACE), it would be best to phrase the first bullet above as "the unicode range of a built-in font resource must include U+0020 for it to be a candidate for _the first font obtained_" or something to this effect.

nathaliesoucy689-code (dismissed):

PR #1253, opened 2022-09-26
echo a line of htmldiff js into diff.html

himorin:
As in #1251

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -44,7 +44,9 @@ jobs:
       - name: set target output
         if: ${{ github.event_name == 'pull_request' }}
         working-directory: ./spec/build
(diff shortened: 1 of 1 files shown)


nigelmegitt (approved):
I'm confused why this works, but I can see from https://github.com/himorin/ttml2/commit/91c934e0097526006a5e90534917e4092a610aba and https://github.com/himorin/ttml2/pull/4 that it does, so approving.

nigelmegitt:
@himorin please could you open this PR from the w3c/ttml2 repo to get it to build properly? It is failing at the git push stage because of this, which means we cannot see the diff output it should generate.

himorin:
> @himorin please could you open this PR from the w3c/ttml2 repo to get it to build properly? It is failing at the git push stage because of this, which means we cannot see the diff output it should generate.

Ah, yes. Sorry for pushing to wrong repository.

I've checked around, and found it seems impossible to override this restriction. See, e.g. https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ (will consider about alternative possibility a bit more...)

PR #1254, opened 2022-09-26
echo a line of htmldiff js into diff.html

himorin:
moved from #1253


nigelmegitt (approved):
Thanks!

himorin:
@nigelmegitt added hash for specifying which hash is source of build.

nigelmegitt (approved):
Nice addition of the hash

PR #1256, opened 2022-09-29
not push new branch for PR from forked repository

himorin:
as title.

(note, as test, PR is made from forked repository)

PR #1260, opened 2022-10-04
fix for PR #1256, no comment added for origin (non-forked) repository

himorin:
Fix for PR #1256, which results no comment added for PRs raised from non-forked repository.


himorin:
seems properly working, comment for preview and diff is added by GHAction.

nigelmegitt (approved):

PR #1270, opened 2024-01-24
Correctly position region in position example

### LINKED ISSUE #1269: Incorrect illustration of example for tts:position
In the published recommendation, section 10.2.34 [1], the example illustration [2] does not correctly illustrate the prose and alignment depicted above the example. The associated region's visible bottom edge has a distance of 48px from/to the bottom edge of the postioning rectangle having a vertical extent of 480px. However, the prose specifies "A percentage value component for a horizontal position offset is relative to the width of the positioning rectangle minus the width of the associated region. A percentage value component for a vertical position offset is relative to the height of the positioning rectangle minus the height of the associated region.", therefore the distance between the positioning rectangle's and associated region's reference edge(s) should be (480 - 92) * 0.1 = 38.8px, for which the coverage of the pixels in the bottom pixel row was/is already set correctly (as was/is the top pixel row).

[1] https://www.w3.org/TR/ttml2/#style-attribute-position
[2] https://www.w3.org/TR/ttml2/images/position.png

Presumably correct version:
ttml-position-9px-down

nigelmegitt:
Bottom edge should be 38.8px (rounded to 39px here) from bottom of rendering area, not 48px.

Fixes #1269


nigelmegitt:
Note that the diff link shows no differences because it cannot detect a difference that is only in the content of a PNG image.

skynavga (approved):

